PR #13, opened 2024-11-27
add optimization.operand and optimization.variable to pyproject.toml

xmhk:
* the `pyproject.toml` was missing build instructions for the nested modules in optimize: operand and variable.

the version on pypi works fine overall, however the `Tutorial_5c_Optimization_Case_Study.ipynb` example notebook throws an import error

`
ModuleNotFoundError: No module named 'optiland.optimization.variable'
`

* the same happend when i clone the repo and built + install the local the package.
* including explicit instructions in `pyproject.toml` works out 

great work! 

and 
best regards

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,6 +39,8 @@ packages = [
     "optiland.analysis",
     "optiland.geometries",
     "optiland.materials",
+    "optiland.optimization.operand",
+    "optiland.optimization.variable",
     "optiland.optimization",
     "optiland.rays",
     "optiland.samples",


xmhk:
However, I noticed: now the optimizer is doing something but it is not converging towards the solution shown in the example notebook. maybe starting parameters changed?

HarrisonKramer:
Thanks a lot for making this improvement.

I recreated the issue and see the addition in pyproject.toml fixes the error.

HarrisonKramer:
> However, I noticed: now the optimizer is doing something but it is not converging towards the solution shown in the example notebook. maybe starting parameters changed?

I see the same behavior, which is peculiar. I have made several changes to the optimization functionality over the last couple months, but hadn't caught this. I do see that changing the initial values of the lens radii of curvature brings the final optimization result to something closer to what's seen in the original notebook. I will investigate this separately, look for possible causes/bugs, and, if needed, add some tests to catch this if it happens again.

Thanks for pointing this out!

Regards,
Kramer

PR #17, opened 2024-12-02
Feat/rms vs field

HarrisonKramer:
- Add RMS spot size vs. field and RMS wavefront error vs. field
- Added into analysis in rms_vs_field.py


logiqal-jeff:
Added 04afa13d (bc-run-2 agent): Robin rim term is now assembled directly on the straight r=r_outer edge facets (analytic 1D Lagrange facet mass matrices) instead of via FacetBasis. Reason: on curved meshes whose sag-clamped rim folds (|R| < r_outer — e.g. iline-retouch coupled group s5–s10, R=−21.86 mm inside a 21.72 mm rim), the inverted rim cells make skfem's isoparametric Newton inversion diverge and every h_edge solve crashed ("Newton iteration didn't converge up to TOL=1e-12"). Dirichlet path unaffected (no FacetBasis).

Equivalence pinned by tests (element orders 1 and 2, flat + curved healthy meshes, ≤1e-10 relative); as-received 5 W/455/none robin-good reproduces the validation-record value to 1e-10 waves (0.073923904171 vs 0.073924). Folded-rim coupled geometry now solves and orders correctly between the h→0 and Dirichlet bounds (new test). Full thermal suite: 546 passed.

https://claude.ai/code/session_01E3aWhya4wcLBuhPpJWRyUd

PR #32, opened 2025-01-08
fix: x and y limits in optical system view

edavidfs:
Adding xlim and ylim to a 2D plot did not work because it seems that the plot boundaries were restored.


HarrisonKramer:
Thanks for fixing this! Will merge now.

HarrisonKramer:
Hi @edavidfs,

By the way, another collaborator set up a discord server for Optiland contributors (or users). The goal is to make it easier to discuss Optiland-related topics/planning. Feel free to join, if you wish:

https://discord.gg/VPrwKR9S

Regards,
Kramer

PR #58, opened 2025-02-01
feat: Zernike polynomials

drpaprika:
This PR adds the `zernike` surface type to Optiland. It is a distinct feature from the Zernike fitting module used to represent wavefront aberrations.

For reference:
```
Zernike polynomials are a set of orthogonal functions defined over the unit disk, 
widely used in freeform optical surface design. 
They efficiently describe wavefront aberrations and complex surface deformations 
by decomposing them into radial and azimuthal components. 
Their orthogonality ensures minimal cross-coupling between terms, 
making them ideal for optimizing optical systems. 
In freeform optics, they enable precise control of surface shape, 
improving performance beyond traditional spherical and aspheric designs.
```

I detailed the mathematical reasoning behind the equations of `zernike.py` in `zernike_description.markdown`, and its implementation follows those of the other surface types (polynomial, chebyshev...).

I followed your sign convention, which by the way seems to have a negative z-component of the normal (pointing “down”):
```python3
norm = np.sqrt(dzdx**2 + dzdy**2 + 1)
nx = dzdx / norm
ny = dzdy / norm
nz = -1 / norm
```
I would expect to have the opposite:
```python3
norm = np.sqrt(dzdx**2 + dzdy**2 + 1)
nx = - dzdx / norm
ny = - dzdy / norm
nz = 1 / norm
```

I updated the notebook `Tutorial_7d_Three_Mirror_Anastigmat.ipynb` with a TMA that has Zernike freeform mirrors, compared to the XY Polynomials before. The optical quality is not that great, but with the same number of coefficients the image height operands are better enforced:
image

This is probably due to the fact that Zernike Polynomials are orthogonal and the optimization converges better.

Best,
drpaprika


optiland/optic.py
```diff
@@ -404,13 +404,25 @@ def update_paraxial(self):
         yb = np.abs(np.ravel(yb))
         for k, surface in enumerate(self.surface_group.surfaces):
             surface.set_semi_aperture(r_max=ya[k]+yb[k])
+            self.update_normalization(surface)
 
-    def update(self):
+    def update_normalization(self, surface)->None:
```
HarrisonKramer, line 409: why is it required to update the normalization radius or x/y values in this way? Previously, I had static normalization constants for the Chebyshev surface. Adding the `update_normalization` call to the `update` method (through the `update_paraxial` method) could add some overhead, because `update` is called once per optimization iteration.

I'll admit it's not drastic - one call to `update_paraxial` runs in ≈2 ms on my machine, so won't be noticeable in most cases. But good to avoid if not necessary.

optiland/geometries/zernike_description.markdown
HarrisonKramer, line 1: This is great. Thanks for adding this as a reference. I'd probably add it somewhere under the `docs` directory instead, but I'm not opposed to keeping it here, if you wish.

This info can make its way into the readthedocs docs too, eventually. Would be a valuable addition.

tests/test_geometries.py
```diff
@@ -616,3 +616,15 @@ def test_str(self):
         geometry = geometries.OddAsphere(cs, radius=10.0, conic=0.5,
                                          coefficients=[1e-2, -1e-5])
         assert str(geometry) == 'Odd Asphere'
+
+
+class TestZernikeGeometry:
```
HarrisonKramer, line 621: we should add more tests eventually. Not needed for this PR though.

HarrisonKramer (commented):
Thanks for adding this! A really impressive addition.

The PR can probably already be merged, but please clarify the question about updating the normalization factors in `optic.py`. If you're convinced this is required, then we can merge as-is. Otherwise, would be good to remove/change that.

HarrisonKramer:
Regarding sign convention, I don't recall why I chose to use a negative z instead of x/y. If you're keen on changing it, then also fine.

At one point, I wasn't handling the surface normal signs robustly enough, so added the `_align_surface_normal` method to `rays.real_rays.RealRays`. If this can be changed/simplified, feel free to do so.

Thanks again!

optiland/optic.py
drpaprika: It is intended to update it at each optimization cycle, otherwise the aperture can grow while the normalization radius remains the same. This prevents a `ValueError` in `_validate_inputs()` and ensures that the polynomials are computed over an area that matches the surface aperture

optiland/geometries/zernike_description.markdown
drpaprika: Feel free to place it where you want

optiland/optic.py
HarrisonKramer: That's clear. I was looking at Zemax where it seems the normalization radius is a constant (correct me if I'm wrong). In that case, you'd have to choose a radius that's large enough to account for the size of the aperture changing.

I agree your implementation makes more sense though - I'd expect the normalization radius to match the surface semi-aperture. This behavior should be documented. I'll make a note to add this in the docs, but it's not blocking the PR.

HarrisonKramer:
Thanks for another impressive addition, @drpaprika! Merging now.

drpaprika:
Hi not reopening this PR, I just wanted to support my point and provide insight on the normalization radius relative to the surface aperture diameter.

> why is it required to update the normalization radius or x/y values in this way? Previously, I had static normalization constants for the Chebyshev surface. Adding the `update_normalization` call to the `update` method (through the `update_paraxial` method) could add some overhead, because `update` is called once per optimization iteration.
> 
> I'll admit it's not drastic - one call to `update_paraxial` runs in ≈2 ms on my machine, so won't be noticeable in most cases. But good to avoid if not necessary.

Here are a two publications that discuss this:

1. **On-the-fly surface manufacturability constraints for freeform optical design** by Kevin P. Thompson (2019) [DOI](https://doi.org/10.1364/OE.27.006129)

> For all designs, the normalization radius of each freeform surface was set to enclose the size of the effective aperture during optimization.


2. **Exit pupil quality analysis and optimization in freeform afocal telescope systems**  by Aaron Bauer (2023) [DOI](https://doi.org/10.1364/OE.496444) 

> One critical aspect of optimizing with orthogonal polynomials is handling the normalization radius. 
> Our preferred method is to let the normalization radius vary in optimization and constrain its value always to
> be 5% larger than the semi-diameter of the optical surface. 
> This technique allows the system to continuously change form without encountering scenarios 
> where the surface semi-diameters become larger than the normalization radius, which can cause raytrace errors.

That's why I included a call to `update_paraxial()` at each optimization cycle.
I hope that helps,
drapaprika

HarrisonKramer:
Nice! Thanks for sharing. That does help me understand this better. I am not a freeform expert, so didn't know what is standard.

I did make some minor optimizations to not call `update_paraxial` in an optimization iteration if it's not needed. I also plan to slightly rewrite the function (or the underlying tracing) to speed it up a bit. That will help too.

Kramer

PR #59, opened 2025-02-01
add SurfaceViewer to vizualize surfaces

drpaprika:
Following the implementation of Zernike surfaces I was curious about their shape. In freeform manufacturing, deviation to the best sphere is crutial.

Now, using `lens.draw_surface(surface_index=1)` the user can view the surface shape in 2D or 3D :
image
image

Note that this computed using `.sag()`: for aspheric surfaces it includes the contributions from the base conic and the polynomials.
I'm not sure that the deviations are actually correct, they seem to small but maybe I'm mistaken?

### DIFF
--- a/optiland/visualization/visualization.py
+++ b/optiland/visualization/visualization.py
@@ -20,6 +20,120 @@ from optiland.visualization.system import OpticalSystem
 plt.rcParams.update({'font.size': 12, 'font.family': 'cambria'})
 
 
+class SurfaceViewer:
+    """
+    A class used to visualize surfaces.
+
+    Args:
+        optic: The optical system to be visualized.
+        surface_index: Index of the surface to be visualized.
+    """
+
+    def __init__(self, optic):
+        self.optic = optic
+
+    def view(self,
+             surface_index: int,
+             projection='2d',
+             num_points=256,
+             figsize=(7, 5.5),
+             title: str = None):
+        """
+        Visualize the surface.
+
+        Args:
+            surface_index (int): Index of the surface to be visualized.
+            projection (str): The type of projection to use for visualization.
+                Can be '2d' or '3d'.
+            num_points (int): The number of points to sample along each axis
+                for the visualization.
+            figsize (tuple): The size of the figure in inches.
+                Defaults to (7, 5.5).
+            title (str): Title.
+
+        Raises:
+            ValueError: If the projection is not '2d' or '3d'.
+        """
+        surface = self.optic.surface_group.surfaces[surface_index]
+        x, y = np.meshgrid(np.linspace(-1, 1, num_points),
+                           np.linspace(-1, 1, num_points))
+        z = surface.geometry.sag(x, y)
+        z[np.sqrt(x**2+y**2) > 1] = np.nan
+
+        if projection == '2d':
+            self._plot_2d(z, figsize=figsize, title=title, 
+                          surface_type=surface.surface_type,
+                          surface_index=surface_index)
+        elif projection == '3d':
+            self._plot_3d(x, y, z, figsize=figsize, title=title,
+                          surface_type=surface.surface_type,
+                          surface_index=surface_index)
+        else:
+            raise ValueError('OPD projection must be "2d" or "3d".')
+
+    def _plot_2d(self, z, figsize=(7, 5.5), title: str = None, **kwargs):
+        """
+        Plot a 2D representation of the given data.
+
+        Args:
+            z (numpy.ndarray): The data to be plotted.
+            figsize (tuple, optional): The size of the figure
+                (default is (7, 5.5)).
+            title (str): Title.
+        """
+        _, ax = plt.subplots(figsize=figsize)
+        im = ax.imshow(np.flipud(z), extent=[-1, 1, -1, 1])
+
+        ax.set_xlabel('Normalized X [mm]')
+        ax.set_ylabel('Normalized Y [mm]')
+        if title is not None:
+            ax.set_title(title)
+        else:
+            ax.set_title(f'Surface {kwargs.get("surface_index", None)} deviation to plane\n'
+                         f'{kwargs.get("surface_type", None).capitalize()} surface')
+
+        cbar = plt.colorbar(im)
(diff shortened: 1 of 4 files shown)


optiland/visualization/visualization.py
```diff
@@ -20,6 +20,120 @@
...
+                for the visualization.
+            figsize (tuple): The size of the figure in inches.
+                Defaults to (7, 5.5).
+            title (str): Title.
+
+        Raises:
+            ValueError: If the projection is not '2d' or '3d'.
+        """
+        surface = self.optic.surface_group.surfaces[surface_index]
+        x, y = np.meshgrid(np.linspace(-1, 1, num_points),
```
HarrisonKramer: - You'll need to run `self.optic.update_paraxial`, then retrieve `surface.semi_aperture` to get the full extent to use for x & y.
- `sag` accepts (non-normalized) lens units, so your linear space should span (`-surface.semi_aperture`, `surface.semi_aperture`)

HarrisonKramer (changes requested):
Solid addition!

We can merge after you update the limits in the call to `sag` in the `SurfaceViewer` class.

However, you mention that deviation to the sphere is crucial, but you're plotting deviation to the plane. Would it make sense to add an argument to `SurfaceViewer.view` like `subtract_sphere`? This gives the user the option to subtract a sphere or a plane (i.e., subtract nothing). You'd have to add a condition for the title too (deviation from sphere vs. plane).

Here are some (untested) methods that you could incorporate into `SurfaceViewer` to compute the deviation from the best-fit sphere. Perhaps there are better approaches.

```python
import numpy as np
from scipy.optimize import minimize

class SurfaceViewer:

    # insert other methods here...

    @staticmethod
    def _sphere_sag(x, y, R):
        """Compute the sag of a sphere with radius R.
        
        Args:
            x, y: 2D arrays of coordinates.
            R: Sphere radius.

        Returns:
            2D array of sag values.
        """
        return R - np.sqrt(R**2 - x**2 - y**2)

    def _best_fit_sphere(x, y, z):
        """Find the best-fit sphere radius
        
        Args:
            x, y: 2D arrays of coordinates.
            z: 2D array of sags.

        Returns:
            Optimal sphere radius.
        """
        def error(R):
            z_s = self._sphere_sag(x, y, R)
            return np.sum((z - z_s) ** 2)  # RMS error

        R_guess = np.max(np.sqrt(x**2 + y**2))  # Initial guess
        res = minimize(error, R_guess)
        return res.x[0]

    def _compute_deviation(x, y, z):
        """Compute deviation from the best-fit sphere.
        
        Args:
            x, y: 2D arrays of coordinates.
            z: 2D array of sags.

        Returns:
            2D array of deviation values.
        """
        R = self._best_fit_sphere(x, y, z)
        z_s = self._sphere_sag(x, y, R)
        return z - z_s
```

drpaprika:
Following your comment the `view()` method uses `surface.semi_aperture` to compute the sag. The figure is now in non-normalized coordinates 
image

I'll take a look at plotting the deviation to the BFS later

HarrisonKramer:
Sounds good. I leave it up to you whether you want to add the deviation to the BFS in this PR or another. I'll quickly review the code now, then should be good to go.

optiland/visualization/visualization.py
```diff
@@ -20,6 +20,156 @@
...
+            title (str): Title.
+        """
+        _, ax = plt.subplots(figsize=figsize)
+
+        if 'semi_aperture' in kwargs:
+            semi_aperture = kwargs['semi_aperture']
+            extent = [-semi_aperture, semi_aperture, -semi_aperture, semi_aperture]
+            ax.set_xlabel('X [mm]')
+            ax.set_ylabel('Y [mm]')
+        else:
```
HarrisonKramer: It looks like `semi_aperture` is always passed as an argument, so the "else" condition will never occur. If that's the case, then you can remove the if statement entirely.

HarrisonKramer (approved):
Feel free to merge when you're ready to do so. Thanks again.

HarrisonKramer:
I quickly looked at this. First, I didn't have the sag equation quite right. It was missing a sign factor to account for negative radii of curvature. This seems to work as I expect:

```python
return R - np.sign(R) * np.sqrt(R**2 - x**2 - y**2)
```

Then, it seems to help when changing the algorithm to Nelder-Mead and improving the initial guess:

```python
initial_guess = np.mean(z + (x**2 + y**2) / (2 * z))  # first order approx
res = optimize.minimize(rms_error, initial_guess, method='Nelder-Mead')
```

I do not know if this will be generally robust, but it retrieves the correct radius of curvature for every surface of the Cooke triplet. By the way, I see it either fails or plots empty axes if you use the object or image surface. Maybe good to catch and prevent that, or update the code to handle it.

drpaprika:
Implemented your suggestions, the result looks better. However the radius of the bfs is -1500mm while the radius of the surface is -100mm

image
image

I'll take a closer look when I have a moment :)

HarrisonKramer:
I wonder if this has something to do with the Zernike geometry's sag definition. I see you used normalized radial coordinates (rho) for the base conic, so your BFS might be closer to the nominal radius * norm_radius.

Now that I think about this - the base class of the Zernike geometry, `NewtonRaphsonGeometry`, actually assumes the base conic is defined with a non-normalized radial component. It should work fine still, but maybe it converges more slowly. Do you think we should quickly change that? I don't know if there are other implications, or if it doesn't follow typical freeform standards. We can put that into a separate issue if needed.

drpaprika:
I tried to see if the results are improved by fitting a conic instead of a sphere : not really. I don't see an issue if we delete this to keep a best fit sphere, it is exploratory.

Regarding your point I think that it's better to have the same convention everywhere to avoid unecessary complexity. Please implement the changes you have in mind, maybe I'm missing something.

I'm still not convinced that the plotted deviations have any physical meaning at the moment, I suggest to not close this PR until we have a better idea of what's happening.

drpaprika:
Closing this PR as the feature has been implemented in  [PR#229](https://github.com/HarrisonKramer/optiland/pull/229).

PR #66, opened 2025-02-13
Airy disc tests

SeckinBerkay:
Implemented airy disc radius tests for the Cooke triplet in tests/test_analysis.py. 

Possible Changes:
- The tolerance 1e-9 could be loosened.
- For readability, I split the tuple into x and y axes separately. For compatibility, I could change the airy_disc_x_y function so that it is in the form of [[x_1, y_1], [x_2, y_2], [x_3, y_3]], instead of ([x_1, x_2, x_3], [y_1, y_2, y_3]).

p.s: Recognised there are two "tripet" typos in the file, don't know if they are intentional or not, so didn't touch them.


HarrisonKramer:
Thanks for submitting this PR!

It looks like you also cleaned up the spot diagram code a bit. Much appreciated :). I may later make minor changes following PEP-8, but not critical for now.

All tests look good, other than a failure to update code coverage with Python 3.9. I've never seen that before. I will investigate this, but it's not blocking. I expect we can merge soon.

Kramer

SeckinBerkay:
recomitted with PEP-8 standards, deleted double spaces, and space for the function input. Tests, passed. The other error in the first tests said it was due to some token error? Might be a bug from the Codecov side. Also I noticed that the coverage is lower? I may add some more tests to other functions as I mentioned earlier in the spot_diagram PR.

HarrisonKramer:
Awesome, thanks for updating! I appreciate it. Looks good. Merging now.

Thanks again for another solid contribution.

PR #69, opened 2025-02-16
Spot Diagram: Correctly transform global coordinates to local image surface coordinates for tilted surfaces

### LINKED ISSUE #68: [Bug] Spot Diagram with Rotated Image Plane
**Describe the bug**
When designing systems in which the image plane is tilted/rotated, the spot diagram sometimes represents the data we don't want.

**To Reproduce**
Steps to reproduce the behaviour, simply setup a system for which the image plane is rotate about one of its axes:

```python
import numpy as np
import optiland
from optiland.materials import Material, IdealMaterial, AbbeMaterial
from optiland import optic, analysis, optimization

# Initiate the optical system
lens = optic.Optic()

# Set the system settings
wvl = 0.355
lens.set_aperture(aperture_type='EPD', value=8.0)
lens.set_field_type(field_type='angle')
lens.add_field(y=0)
lens.add_wavelength(value=wvl, is_primary=True)

# Setup the LDE data
lens.add_surface(index=0, radius=np.inf, z=-np.inf)
lens.add_surface(index=1, z=-40)
lens.add_surface(index=2, z=-20, is_stop=True)
lens.add_surface(index=3, x=0, y=0, z=0, ry=np.pi/4, material="mirror", surface_type="standard") # M1
lens.add_surface(index=4, x=-15, y=0, z=0, ry=-np.pi/4, rx=np.pi/2, material="mirror", surface_type="standard") # M2
lens.add_surface(index=5, x=-15, y=20, z=0, rx=np.pi/2) # Image plane

lens.draw3D()

spot = analysis.SpotDiagram(lens)
spot.view()

```

**Expected behavior**
If correctly defined, the spot diagram should take into account the local (rotated, e.g) coordinates of the image plane, and not the global ones. 

**Screenshots**

This is what the system looks like:

Image

This is what the wrong spot diagram will look like:

Image

**Environment**  
- Optiland Version: v0.2.6

**Additional Comments**
Creating a PR soon with the bug fix.

manuelFragata:
(fixes #68)
### Description
When the image surface is tilted (e.g., with a rotation such as `rx=np.pi/2`), the spot diagram generated by the `SpotDiagram` analysis does not display the correct spot location. Instead, it shows an incorrect distribution (e.g., a horizontal line at Y=0).

### Cause
The issue arises because the ray-tracing routine stores the intersection points in the global coordinate system. The current implementation of `_generate_field_data` directly extracts these global coordinates from the final (image) surface without converting them into the local coordinate system of the tilted image plane.

### Proposed Fix
I modified `_generate_field_data `in `spot_diagram.py` to transform the global coordinates into the image surface's local coordinate system using the existing `transform()` utility function with `is_global=True`. The spot is now correctly visualized:

### Testing
With this change, when running the same optical system as described in the bug issue, the resulting spot diagram now shows the spot correctly positioned according to the image plane's local coordinate system, as in the image:

spot_better

I would also appreciate some input regarding the new axes naming, because here having "Y" and "X" loses a bit the sense I think. Any suggestions?

Let me know if any further changes are needed! :)

### DIFF
--- a/optiland/analysis/spot_diagram.py
+++ b/optiland/analysis/spot_diagram.py
@@ -450,10 +450,19 @@ class SpotDiagram:
             # Without the airy disk, just use the geometric spot radius with a buffer.
             adjusted_axis_lim = axis_lim * buffer
 
+        # Determining the labels for the x and y axes based on the image surface effective orientation.
+        cs = self.optic.image_surface.geometry.cs
+        effective_orientation = np.abs(cs.get_effective_rotation_euler())
+        # Define a small tolerance to apply the new label 
+        tol = 0.01 # adjust it, if necessary
+        if effective_orientation[0] > tol or effective_orientation[1] > tol:
+            x_label, y_label = 'U (mm)', 'V (mm)'
+        else:
+            x_label, y_label = 'X (mm)', 'Y (mm)'         
         
         ax.axis('square')
-        ax.set_xlabel('X (mm)')
-        ax.set_ylabel('Y (mm)')
+        ax.set_xlabel(x_label)
+        ax.set_ylabel(y_label)
         ax.set_xlim((-adjusted_axis_lim, adjusted_axis_lim))
         ax.set_ylim((-adjusted_axis_lim, adjusted_axis_lim))
         ax.set_title(f'Hx: {field[0]:.3f}, Hy: {field[1]:.3f}')
--- a/optiland/analysis/spot_diagram.py
+++ b/optiland/analysis/spot_diagram.py
@@ -371,13 +371,25 @@ class SpotDiagram:
                 rays. Defaults to 'hexapolar'.
 
         Returns:
-            list: A list containing the x, y, and intensity values of the
+            list: A list containing the local x-coordinates, local y-coordinates, and intensity values of the
                 generated spot data.
         """
         self.optic.trace(*field, wavelength, num_rays, distribution)
-        x = self.optic.surface_group.x[-1, :]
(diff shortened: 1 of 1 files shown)


HarrisonKramer (approved):
Hi Manuel,

Looks great! Thanks for finding this bug and making the fix.

I may later add some more unit tests here to ensure the behavior is as expected for non-symmetric systems. That's not blocking though.

Kramer

HarrisonKramer:
Regarding the axis naming, I'm inclined to use "U (mm)" and "V (mm)" for local coordinates. It would be nice to only use UV when XY is not appropriate. Do you want to think of a way to check this, then plot with either UV or XY depending on image surface orientation?

Maybe one option would be to compare the transformed coordinates to the original coordinates. You could also look into the coordinate system properties of the image surface. I can help here if needed. Just let me know :)

Kramer

manuelFragata:
> Regarding the axis naming, I'm inclined to use "U (mm)" and "V (mm)" for local coordinates. It would be nice to only use UV when XY is not appropriate. Do you want to think of a way to check this, then plot with either UV or XY depending on image surface orientation?
> 
> Maybe one option would be to compare the transformed coordinates to the original coordinates. You could also look into the coordinate system properties of the image surface. I can help here if needed. Just let me know :)
> 
> Kramer

Done, as you suggested. I applied a small tolerance for which we check the rotation of the surface and then decide which name we give to the axes. Let me know if it's okay :)

HarrisonKramer:
Awesome, thanks! I will add more unit tests later on.

I really appreciate the contribution. Merging now!

PR #77, opened 2025-02-19
Fix: Handle Missing Extinction Coefficient Data by Defaulting to Zero

manuelFragata:
### Description
This PR addresses the issue where the `MaterialFile.k()` function raises a `ValueErro`r when no extinction coefficient data is found in the material file (e.g., with Malitson.yml - this came up when I was using the "fused_silica" material, for example). 

### Proposed Fix
- **Updated the `k()` Method in `material_file.py`:**
  - **Data Check:**  
    The method now checks whether `self._k` or `self._k_wavelength` is `None`. If one of them is missing, the function will set a default extinction coefficient of zero.
  - **Warning Message:**  
    A print statement has been added to notify the user, because it may be important for some of their analyses:
    > "WARNING: Extinction coefficient not found. It is assumed to be 0"
  - **Return Values:**  
    - If the input `wavelength` is a scalar, the method returns `0.0`.
    - If the input `wavelength` is a NumPy array, it returns an array of zeros matching the shape of the input.
  - **Interpolation:**  
    If the extinction data is present, the method uses `np.interp` to interpolate the extinction coefficient for the given wavelengths, as the already implemented approach.

### Testing
Here is the warning being triggered, but it allows the script to continue running:

image

### Additional Modifications
I would also need some suggestions on how to show this warning message only once. Right now it will show up multiple times, as many as the method k() is called. 
Any ideas on how I can improve this?

Thanks :)
Manuel

### DIFF
--- a/optiland/materials/material_file.py
+++ b/optiland/materials/material_file.py
@@ -91,10 +91,10 @@ class MaterialFile(BaseMaterial):
         """
         # If the extinction coefficient is missing from the file, return 0
         if self._k is None or self._k_wavelength is None:
-            if not self.k_warning_printed:
+            if not self._k_warning_printed:
                 material_name = os.path.basename(self.filename)
                 print(f"WARNING: No extinction coefficient data found for {material_name}. Assuming it is 0.")
-                self.k_warning_printed = True # we set it to True to avoid printing the warning again            
+                self._k_warning_printed = True # we set it to True to avoid printing the warning again            
             
             if np.isscalar(wavelength):
                 return 0.0
--- a/optiland/materials/material_file.py
+++ b/optiland/materials/material_file.py
@@ -91,12 +91,14 @@ class MaterialFile(BaseMaterial):
         Raises:
             ValueError: If no extinction coefficient data is found.
         """
-        try:
-            return np.interp(wavelength, self._k_wavelength, self._k)
-        except ValueError:
-            file = os.path.basename(self.filename)
-            raise ValueError(f'No extinction coefficient data found for '
-                             f'{file}.')
+        # If the extinction coefficient is missing from the file, return 0
+        if self._k is None or self._k_wavelength is None:
+            print("WARNING: Extinction coefficient not found. It is assumed to be 0")
+            if np.isscalar(wavelength):
+                return 0.0
+            else:
+                return np.zeros_like(wavelength) # if there is an array of wvls
+        return np.interp(wavelength, self._k_wavelength, self._k)
 
     def _formula_1(self, w):
         """
--- a/optiland/materials/material_file.py
+++ b/optiland/materials/material_file.py
@@ -40,7 +40,7 @@ class MaterialFile(BaseMaterial):
     """
     def __init__(self, filename):
         self.filename = filename
-
+        self._k_warning_printed = False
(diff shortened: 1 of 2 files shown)


manuelFragata:
I think it failed these tests because it did not raise the exception error (that I removed). Suggestions?

HarrisonKramer:
Hi Manuel,

Thanks for opening this!

Yes, the failure is expected, so no problem. Those tests should either be removed or updated to catch the warning output. The latter is not trivial, though I do it elsewhere in the tests, I believe.

I used a similar method in `materials.Material` when no exact material match is found - it simply prints a warning using `print`. _However_, I wonder if we should make warnings more robust and actually use `warnings`. This would enable you to have this warning raised only once - otherwise I don't think it's possible (at least not easily, as far as I know). See [docs for warnings](https://docs.python.org/3/library/warnings.html#the-warnings-filter).

Simple example:
```python
import warnings
warnings.simplefilter("once", UserWarning)
```

One other nice thing about using `warnings` is that users can also ignore them, if they wish. I am not certain of the best approach, but I'm inclined to use `warnings`, especially in your case when the print statement appears multiple times (could be 100s for some operations). If we go this route, then I'd want to change the warning in the materials module too, though that doesn't need to be in this PR.

What do you think?

Regards,
Kramer

optiland/materials/material_file.py
```diff
@@ -80,23 +80,28 @@ def n(self, wavelength):
...
-            ValueError: If no extinction coefficient data is found.
         """
-        try:
-            return np.interp(wavelength, self._k_wavelength, self._k)
-        except ValueError:
-            file = os.path.basename(self.filename)
-            raise ValueError(f'No extinction coefficient data found for '
-                             f'{file}.')
+        # If the extinction coefficient is missing from the file, return 0
+        if self._k is None or self._k_wavelength is None:
```
HarrisonKramer, line 93: can you move the attribute definition into `__init__`?

For example:
```python
self._k_warning_printed = False
```
(I also make it protected with the `_`)

Then, you can simplify this logic:
```python
if not self._k_warning_printed:
    print('warning...')
    self._k_warning_printed = True
```

HarrisonKramer (changes requested):
Looks good. This approach works well for now.

Can you delete the failing tests and make the minor change I suggested?

Then we can merge 👍 

Kramer

optiland/materials/material_file.py
manuelFragata: Thanks! Yes, will change now. I also updated the tests.

manuelFragata:
Hi Kramer,

Thank you for the feedback!

I agree, we can use `warnings` in the future, to be more robust. I did manage to find a quick solution to printing the warning only once. I simply created a flag `self.k_warning_printed`  that if True, then the warning will be printed only once, which is already a lot better. I wouldn't mind working on this new approach for warnings in the material module, so I can open another PR. 

Thank you!

Manuel

HarrisonKramer (approved):

HarrisonKramer:
Thanks a lot for the contribution, @manuelFragata! and thanks for volunteering to work on warnings in the material module.

PR looks good. Merging now!

Regards,
Kramer

PR #79, opened 2025-02-22
Add option to pass a callback function to `optimize()`

drpaprika:
This simple PR allows to pass a callback function to scipy's `optimize()`.
This allows the user to add anything he likes to vizualize the optimization progress, for example (tested in a notebook):

```python3
from IPython.display import clear_output

optimizer = optimization.OptimizerGeneric(problem)

history = []
def callback(_):
    # Plot the lens at each merit function call
    clear_output(wait=True)
    lens.draw()
    
    # Plot merit function evolution
    f_val = optimizer.problem.sum_squared()
    history.append(f_val)
    plt.figure(figsize=(6, 4))
    plt.plot(history, color='black')
    plt.yscale('log')
    plt.title("Merit function value")
    plt.xlabel("Iteration")
    plt.ylabel("Merit function value")
    plt.grid(alpha=0.25)
    plt.show()

res = optimizer.optimize(tol=1e-9, callback=callback)
```

plot_during_optimization

Only `LeastSquares` doesn't seem to accept a callback function.

I also included a fix to update the lens at the end of the optimization with the best paramerters (`result.x`), as the last merit function call is not necessarily the lowest one.

Of course plotting during the optimization slows it significantly, but that's up to the user.
drpaprika

### DIFF
--- a/optiland/optimization/optimization.py
+++ b/optiland/optimization/optimization.py
@@ -330,13 +330,14 @@ class DualAnnealing(OptimizerGeneric):
     def __init__(self, problem: OptimizationProblem):
         super().__init__(problem)
 
-    def optimize(self, maxiter=1000, disp=True):
+    def optimize(self, maxiter=1000, disp=True, callback=None):
         """
         Runs the dual annealing algorithm to optimize the problem.
 
         Parameters:
             maxiter (int): Maximum number of iterations.
             disp (bool): Whether to display the optimization process.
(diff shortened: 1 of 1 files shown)


HarrisonKramer (approved):

HarrisonKramer:
Looks good. Thanks for the contribution!

And nice catch on updating to the best parameters after the optimization completes. I'll merge now.

Regards,
Kramer

PR #80, opened 2025-02-23
fix: asphere coeff `get_value()`

drpaprika:
Fix the `IndexError` raised when when no coeffs are defined on the surface.
It happens, for example, when a surface is defined as an even asphere with an empty coefs list and an operand is set:

```python3
lens.add_surface(index=1,  radius=100, thickness=10, surface_type='even_asphere', coefficients=[])
problem.add_variable(lens, 'asphere_coeff', surface_number=1, coeff_number=5, min_val=-1, max_val=1)
```

raises `IndexError: list index out of range`.
I simply reused what's already implemented in `zernike_coeff.py`.
drpaprika

### DIFF
--- a/optiland/optimization/variable/asphere_coeff.py
+++ b/optiland/optimization/variable/asphere_coeff.py
@@ -8,6 +8,7 @@ optical system.
 Kramer Harrison, 2024
 """
 
+import numpy as np
 from optiland.optimization.variable.base import VariableBehavior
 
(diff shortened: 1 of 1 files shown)


HarrisonKramer (approved):

HarrisonKramer:
Looks good. Merging now.

Thanks,
Kramer

Issue #67, opened 2025-02-14
Quick Focus


SeckinBerkay:
**Is your feature request related to a problem? Please describe.**
A quick focus for the image surface may be added. 

**Describe the solution you'd like**
_Option 1 for Function Placement:_
    Implement the functionality to add_surface function. In the add_surface definition, "quick_focus=True" is present and adjusts the surface length minimizing the RMS spot size for the primary wavelength. 
    Pro(s): Easier implementation for the user, and more apparent for readability.
    Con(s): Harder for GUI implementation?

_Option 2 for Function Placement:_
    Implement the functionality to optic class. After all the surfaces are defined, the user can call a lens.minimize_spot_size function, updating the last surface length.
  Pro(s): It will also be implemented easily by the user. 
  Con(s): Easier for GUI implementation. It will not be more apparent for readability (The optical system is already defined), however if GUI is used, then it is irrelevant.

_Options for RMS Spot Size Minimization:_
    1. Directly use Scipy minimize function, easier to implement, and seems to be secure. However, if Option 1 is preferred, the user can't control the iteration number. 
    2. By the angles of the rays, we can analytically calculate the arrival position, and then take the mean position. Might be tricky for some geometries, such as rotated image plane.

**Additional context**
The function can quickly get out of hand if used improperly. Strict caution should be added. For instance, if the user wants to use quick focus for a beam expander system. It simply can not. An error could be thrown, or an iteration number handles it anyway.

CAClaveau:
(Not sure, but it might help: [link](https://github.com/singer-yang/DeepLens/blob/24ea83c8d37ec2eab96930c25ac1a2925e47b514/deeplens/geolens.py#L1583))

HarrisonKramer:
Thanks for opening this issue, @SeckinBerkay. And thanks for sharing this link, @CAClaveau!

I like the idea of a quick focus adjustment, but I think it would fit better as a `Solve` rather than being directly integrated into `Optic`. This delegates the functionality and keeps the `Optic` class cleaner IMO.

Solves are already set up and can be added to an optic, so we would just need to implement a new Solve class for this. I’m not certain of all the details yet, but using a quick focus calculation like @CAClaveau referenced, could work well.

I think we'll need to:

- Implement a new class, e.g., `QuickFocusSolve`, which subclasses `BaseSolve`.
- Update the `SolveFactory` with the new class.
- Add tests and docs.

Feel free to open a PR if you want to work on this. I can also support wherever needed.

[...]

PR #82, opened 2025-02-23
add quick focus to solves.py, and tests

SeckinBerkay:
Added quick focus functionality and their corresponding tests.
However, there is a problem with the test_to_dict test. Need help with it. When I run the pytest, the asserted value is shown as 2.9520799999999987, where it should have been 60.2342033372244.

Possible change: Avoid direct assert, and use pytest.approx.

Example usage for the quick focus:

```python
from optiland.samples.objectives import CookeTriplet
from optiland.solves import QuickFocusSolve

lens = CookeTriplet()
quick_focus = QuickFocusSolve(lens)
quick_focus.apply()
```

### DIFF
--- a/optiland/solves.py
+++ b/optiland/solves.py
@@ -9,7 +9,7 @@ Kramer Harrison, 2024
 """
 
 from abc import ABC, abstractmethod
-from numpy import mean
+from numpy import mean, errstate, where, nan, nanmean
 
 class BaseSolve(ABC):
     """
@@ -123,14 +123,10 @@ class QuickFocusSolve(BaseSolve):
     def __init__(self, optic):
         self.optic = optic
         self.num_surfaces = self.optic.surface_group.num_surfaces
-        if self.num_surfaces > 2:
-            self.pos2 = self.optic.surface_group.positions[-1].astype(float)[0]
-            self.pos1 = self.optic.surface_group.positions[-2].astype(float)[0]
-            self.thickness = self.pos2 - self.pos1
-        elif self.num_surfaces <= 2:
+        if self.num_surfaces <= 2:
             raise ValueError('Can not optimize for an empty optical system')
         
-    def optimal_focus_distance(self, Hx=0, Hy=0, wavelength=0.55, num_rays=5, distribution='hexapolar', max_iter=50_000):
+    def optimal_focus_distance(self, Hx=0, Hy=0, wavelength=0.55, num_rays=5, distribution='hexapolar'):
         """
         Compute the optimal location of the image plane where the RMS spot size is minimized.
         This is based on solving the quadratic equation that describes the RMS spot size as a function
@@ -143,77 +139,24 @@ class QuickFocusSolve(BaseSolve):
             t_opt : The propagation distance from the image plane that minimizes the RMS spot size.
         """
         # Trace rays to the image plane
-        rays = self.optic.trace(Hx=Hx, Hy=Hy, wavelength=wavelength, num_rays=num_rays, distribution=distribution)
-        z_focus_old = mean(rays.z)
-        for iter in range(max_iter):
-            # Compute centroids at t = 0 (i.e., on image plane)
-            x0 = mean(rays.x)
-            y0 = mean(rays.y)
-            
-            # Compute average direction cosines
-            L_avg = mean(rays.L)
-            M_avg = mean(rays.M)
-            
-            # Differences from the means
-            delta_x = rays.x - x0
-            delta_y = rays.y - y0
-            delta_L = rays.L - L_avg
-            delta_M = rays.M - M_avg
-            
-            # Compute coefficients A and B from the quadratic:
-            # f(t) = A*t^2 + 2B*t + C --> we want to find the minimum of f(t)
-            A = mean(delta_L**2 + delta_M**2)
-            B = mean(delta_x * delta_L + delta_y * delta_M)
-            
-            # Check to avoid division by zero if A is zero (i.e., no dependence on t)
-            if A == 0:
-                return 0.0
-            
-            t_opt = -B / A
-            z_focus = mean(rays.z + t_opt * rays.N)
-            if abs(z_focus - z_focus_old) <= 1e-6:
-                break
-            rays.propagate(t_opt)
-            z_focus_old = z_focus
+        # Trace rays to the image plane
 
+        rays = self.optic.trace(Hx=Hx, Hy=Hy, wavelength=wavelength, num_rays=num_rays, distribution=distribution)
+
+        A = rays.L**2 + rays.M**2
+        B = rays.L * rays.x + rays.M * rays.y
+        with errstate(divide='ignore', invalid='ignore'):
+            t_opt = where(A != 0, -B/A, nan)
+        z_focus = nanmean(rays.z + t_opt * rays.N)
+        
         return z_focus
         
     def apply(self):
         """Applies QuickFocusSolve to the optic"""
         z_focus = self.optimal_focus_distance(wavelength=self.optic.wavelengths.primary_wavelength.value)
-
+        
         self.optic.surface_group.surfaces[-1].geometry.cs.z = z_focus
 
-        self.thickness = z_focus - self.pos1 
-
-    def to_dict(self):
-        """
-        Returns a dictionary representation of the solve.
-
-        Returns:
-            dict: A dictionary representation of the solve.
-        """
-        solve_dict = super().to_dict()
-        solve_dict.update({
-            'thickness': float(self.thickness)
-        })
-        return solve_dict
-
-    @classmethod
-    def from_dict(cls, optic, data):
-        """
-        Creates a QuickFocusSolve from a dictionary representation.
-
-        Args:
-            optic (Optic): The optic object.
-            data (dict): The dictionary representation of the solve.
-
-        Returns:
-            QuickFocusSolve: The solve.
-        """
-        instance = cls(optic)
-        instance.thickness = data['thickness']
-        return instance
 
 class SolveFactory:
     """
--- a/tests/test_solves.py
+++ b/tests/test_solves.py
@@ -90,49 +90,18 @@ class TestQuickfocusSolve:
         pos2 = optic.surface_group.positions[-1].astype(float)[0]
         pos1 = optic.surface_group.positions[-2].astype(float)[0]
 
-        assert solve.thickness == pytest.approx(thickness, rel=1e-3)
+        assert pos2 - pos1 == pytest.approx(thickness, rel=1e-3)
 
         # Implementing the extreme shift case.
         optic.surface_group.surfaces[-1].geometry.cs.z += 1000
         solve = solves.QuickFocusSolve(optic)
         solve.apply()
-        
-        assert solve.thickness == pytest.approx(thickness, rel=1e-1)
 
-    def test_to_dict(self):
-        optic = CookeTriplet()
-        optic.surface_group.surfaces[-1].geometry.cs.z -= 10
-        thickness = 32.20778     # The thickness before optimization (.apply is not called)
+        # Check that surface has been shifted
+        pos2 = optic.surface_group.positions[-1].astype(float)[0]
+        pos1 = optic.surface_group.positions[-2].astype(float)[0]
 
-        solve = solves.QuickFocusSolve(optic)
-        data = solve.to_dict()
-
-        assert data['type'] == 'QuickFocusSolve'
-        assert data['thickness'] == thickness
-
-    def test_from_dict(self):
-        optic = CookeTriplet()
-        optic.surface_group.surfaces[-1].geometry.cs.z -= 10
-        data = {
-            'type': 'QuickFocusSolve',
-            'thickness': 42.21812063592369
-        }
-
-        solve = solves.BaseSolve.from_dict(optic, data)
-
-        assert solve.thickness == data['thickness']
-
-    def test_from_dict_invalid_type(self):
-        optic = CookeTriplet()
-        data = {
-            'type': 'Invalid',
-            'thickness': 42.21812063592369
-        }
-
-        with pytest.raises(ValueError):
-            solves.BaseSolve.from_dict(optic, data)
-
-
+        assert pos2 - pos1 == pytest.approx(thickness, rel=1e-3)
 
 class TestSolveFactory:
     def test_create_solve(self):


SeckinBerkay:
#67

SeckinBerkay:
Okay I found the problem, fixing and committing it now.

tests/test_solves.py
```diff
@@ -72,6 +72,54 @@ def test_from_dict_invalid_type(self):
...
             solves.BaseSolve.from_dict(optic, data)
 
+class TestQuickfocusSolve:
+    def test_marginal_ray_height_solve_constructor(self):
+        optic = CookeTriplet()
+        solve = solves.QuickFocusSolve(optic)
+
+        assert solve.optic == optic
+
+    def test_marginal_ray_height_solve_apply(self):
```
SeckinBerkay: update the name to test_quick_focus_solve_apply, avoid copy-pasting issue.

tests/test_solves.py
```diff
@@ -72,6 +72,54 @@ def test_from_dict_invalid_type(self):
         with pytest.raises(ValueError):
             solves.BaseSolve.from_dict(optic, data)
 
+class TestQuickfocusSolve:
+    def test_marginal_ray_height_solve_constructor(self):
```
SeckinBerkay: update the name to test_quick_focus_solve_constructor, avoid copy-pasting issue.

optiland/solves.py
```diff
@@ -113,6 +113,66 @@ def from_dict(cls, optic, data):
...
+
+    Raises:
+            ValueError: If the optical system is not defined.
+    """
+    def __init__(self, optic):
+        self.optic = optic
+        self.num_surfaces = self.optic.surface_group.num_surfaces
+        self.last_surf_idx = self.optic.surface_group.num_surfaces - 2 # Index of the last surface before the image surface.
+        if self.num_surfaces > 2:
+            pos2 = self.optic.surface_group.positions[self.last_surf_idx]
```
SeckinBerkay: you can change the indexing self.optic.surface_group.positions[-1], and  self.optic.surface_group.positions[-2]

HarrisonKramer (changes requested):
Hi Seckin,

Thanks for working on this!

In general, the structure of the code is very good. However, I propose we do not use optimization here, because an analytical solution is more efficient. We should be able to use the majority of your code and just swap out the optimization call with a call to a function (or even better, a protected method in the `QuickFocusSolve` class).

I spent a bit of time working out this solution. You can check my math if you like. I quickly checked that it gives the correct answer when I manually defocus a lens.
```python
def optimal_focus_distance(lens, Hx=0, Hy=0, wavelength=0.55, num_rays=5, distribution='hexapolar'):
    """
    Compute the optimal propagation distance (t_opt) where the RMS spot size is minimized.
    This is based on solving the quadratic equation that describes the RMS spot size as a function
    of the propagation distance.
    
    Args:
        lens (Optic): the optic to analyze
        ...
        
    Returns:
        t_opt : The propagation distance from the image plane that minimizes the RMS spot size.
    """
    # Trace rays to the image plane
    rays = lens.trace(Hx=Hx, Hy=Hy, wavelength=wavelength, num_rays=num_rays, distribution=distribution)

    # Compute centroids at t = 0 (i.e., on image plane)
    x0 = np.mean(rays.x)
    y0 = np.mean(rays.y)
    
    # Compute average direction cosines
    L_avg = np.mean(rays.L)
    M_avg = np.mean(rays.M)
    
    # Differences from the means
    delta_x = rays.x - x0
    delta_y = rays.y - y0
    delta_L = rays.L - L_avg
    delta_M = rays.M - M_avg
    
    # Compute coefficients A and B from the quadratic:
    # f(t) = A*t^2 + 2B*t + C --> we want to find the minimum of f(t)
    A = np.mean(delta_L**2 + delta_M**2)
    B = np.mean(delta_x * delta_L + delta_y * delta_M)
    
    # Check to avoid division by zero if A is zero (i.e., no dependence on t)
    if A == 0:
        return 0.0
    
    t_opt = -B / A
    return t_opt
```

How I derive this:

- I am looking for the displacement `t` from the image plane that results in the minimal RMS spot size.
- For rays on the image surface, we have their nominal `x` and `y` positions, as well as their `L` and `M` direction cosines. We can determine the x position after propagating by `t` as `x'(t) = x + t*L`, where `x'` is the new x value and `x` is the value on the image surface.
- You can write out the equation for the RMS spot size as a function of `t`. I omit it here, because it's a bit long, but this is a quadratic function of `t`.
- We want to find the minimum of this function, so we take the derivative and set to zero.
- Solving for t, this results in equation above.

Do you think you can update your approach to use this function (or something like this)? Do you agree with the approach?

Kramer

HarrisonKramer:
And adding an example showing how the function above works:

```python
lens = CookeTriplet()

# move the image plane 10 mm towards last lens
lens.surface_group.surfaces[-1].geometry.cs.z -= 10

lens.draw()

t_opt = optimal_focus_distance(lens)
print(f"Optimal focus distance: {t_opt:.2f} mm")  # Result = 10.04 mm
```
image

HarrisonKramer:
Actually, I think I made a mistake in the derivation somewhere. I need to check this. It works fine for the on-axis field, but seems to give strange results for off-axis fields.

I will recalculate this (a bit slower this time) and see where my mistake was. Feel free to try this yourself too.

Kramer

HarrisonKramer:
A few things - first, I was missing the last conversion to a `z` position once I derived `t_opt` above. I fixed this in the function. Second, this function assumes linearity in spot centroid with propagation distance. This is not generally true.

In any case, the function above will converge if run a few times. This is probably still faster than using the full-fledged optimization framework. If we are willing to perform the focusing based on only the on-axis field, then this can also be simplified further.

I leave it up to you. If you prefer, we can already merge this, then investigate improvement options later on.

SeckinBerkay:
I actually wanted to hear your opinion/suggestion as well, while I was figuring out if the implementation was easy for the user to use. I can definitely implement your method, it is way more computationally greedy. It will be great. On the other hand, thanks for the code, is there an option to add you as a co-author to this PR?

tests/test_solves.py
```diff
@@ -72,6 +71,61 @@ def test_from_dict_invalid_type(self):
...
 
+class TestQuickfocusSolve:
+    def test_quick_focus_solve_constructor(self):
+        optic = CookeTriplet()
+        optic.surface_group.surfaces[-1].geometry.cs.z -= 10
+        solve = solves.QuickFocusSolve(optic)
+
+        assert solve.optic == optic
+
+    def test_quick_focus_solve_apply(self):
```
HarrisonKramer, line 82: can you add one more check with an extreme misalignment, e.g., 1000 mm? I think it's good to check large values too.

optiland/solves.py
```diff
@@ -113,6 +112,102 @@ def from_dict(cls, optic, data):
...
+        if A == 0:
+            return 0.0
+        
+        t_opt = -B / A
+        z_focus = mean(rays.z + t_opt * rays.N)
+        return z_focus
+        
+    def apply(self):
+        """Applies QuickFocusSolve to the optic"""
+        t_opt = self.optimal_focus_distance()
```
HarrisonKramer: We need to call this a few times to iteratively solve for the best focus position (see previous comment). Do you think it makes sense to monitor convergence and continue calling the function in a `while` loop until the delta is less than a small number (e.g., 0.001 mm)?

optiland/solves.py
```diff
@@ -113,6 +112,102 @@ def from_dict(cls, optic, data):
...
+        self.optic = optic
+        self.num_surfaces = self.optic.surface_group.num_surfaces
+        if self.num_surfaces > 2:
+            self.pos2 = self.optic.surface_group.positions[-1].astype(float)[0]
+            self.pos1 = self.optic.surface_group.positions[-2].astype(float)[0]
+            self.thickness = self.pos2 - self.pos1
+        elif self.num_surfaces <= 2:
+            raise ValueError('Can not optimize for an empty optical system')
+        
+    def optimal_focus_distance(self, Hx=0, Hy=0, wavelength=0.55, num_rays=5, distribution='hexapolar'):
```
HarrisonKramer, line 129: Optionally, we could trace rays only once in the `apply` method, then perform the rest of the calculations here. So we trace rays once, then enter a `while` loop (see below). We would need to propagate the rays to the correct plane in each iteration before passing the rays to this function though.

HarrisonKramer (commented):
Looks good! I had a few more thoughts on the implementation. What do you think?

HarrisonKramer:
No need to add me as a coauthor. I am just glad we can work out the logic here!

optiland/solves.py
SeckinBerkay: I was just writing it, exactly. We shall implement the logic inside a for loop, then stop if convergence or a maximum number of iterations is reached.

tests/test_solves.py
SeckinBerkay: All right, makes sense. To check if I understand it clearly, you mean a misalignment of the image plane, like how far it is from the optimal focus position?

HarrisonKramer: Exactly. You can move the image plane +1000 mm or so, then confirm that the algorithm converges. Larger offsets should result in more iterations. We just want to check that the algorithm is robust in this case.

SeckinBerkay:
> No need to add me as a coauthor. I am just glad we can work out the logic here!

I'm very glad too, and appreciate all the collaboration and guidence.

optiland/solves.py
HarrisonKramer: Yes, as you mention, a `for` loop makes more sense with some max number of iterations.

tests/test_solves.py
SeckinBerkay: It works with 1000 mm, I have been tweaking parameters but it doesn't converge to the same distance as if it was 10 mm misalignment, even if I do a while loop. I will set the approximation error more in the extreme misalignment test case if it is okay.

HarrisonKramer: That is surprising. I will investigate this a bit. I expect it should converge toward the same value.

optiland/solves.py
```diff
@@ -113,6 +112,108 @@ def from_dict(cls, optic, data):
...
+    Args:
+        optic (Optic): The optic object.
+
+    Raises:
+            ValueError: If the optical system is not defined.
+    """
+    def __init__(self, optic):
+        self.optic = optic
+        self.num_surfaces = self.optic.surface_group.num_surfaces
+        if self.num_surfaces > 2:
```
HarrisonKramer: Perhaps I miss the logic, but why are these attributes (`pos1`, `pos2` and `thickness`) needed in the class? You don't need them to apply the solve, it seems. If that's the case, then you can completely remove them and only check that num surfaces > 2.

optiland/solves.py
```diff
@@ -113,6 +112,108 @@ def from_dict(cls, optic, data):
...
+            self.pos2 = self.optic.surface_group.positions[-1].astype(float)[0]
+            self.pos1 = self.optic.surface_group.positions[-2].astype(float)[0]
+            self.thickness = self.pos2 - self.pos1
+        elif self.num_surfaces <= 2:
+            raise ValueError('Can not optimize for an empty optical system')
+        
+    def optimal_focus_distance(self, Hx=0, Hy=0, wavelength=0.55, num_rays=5, distribution='hexapolar', max_iter=50_000):
```
HarrisonKramer: 50_000 iterations seems like way too much. I was considering only using about 5. Is it really the case that 50_000 iterations are actually needed?

optiland/solves.py
```diff
@@ -113,6 +112,108 @@ def from_dict(cls, optic, data):
...
+
+        return z_focus
+        
+    def apply(self):
+        """Applies QuickFocusSolve to the optic"""
+        z_focus = self.optimal_focus_distance(wavelength=self.optic.wavelengths.primary_wavelength.value)
+
+        self.optic.surface_group.surfaces[-1].geometry.cs.z = z_focus
+
+        self.thickness = z_focus - self.pos1
```
HarrisonKramer: if the `thickness` attribute is not needed (seems to be the case), then delete this.

optiland/solves.py
```diff
@@ -113,6 +112,108 @@ def from_dict(cls, optic, data):
...
+        
+    def apply(self):
+        """Applies QuickFocusSolve to the optic"""
+        z_focus = self.optimal_focus_distance(wavelength=self.optic.wavelengths.primary_wavelength.value)
+
+        self.optic.surface_group.surfaces[-1].geometry.cs.z = z_focus
+
+        self.thickness = z_focus - self.pos1 
+
+    def to_dict(self):
```
HarrisonKramer: if the other attributes are removed, then I think you can completely remove the `to_dict` and `from_dict` methods. There's no need to override these if that behavior is the same as the base class. I expect this is the case, but you can confirm.

HarrisonKramer (changes requested):
Good progress. I still wonder a bit about the algorithm. The fact that it's not converging that quickly is troubling. I propose we do not merge until we understand why.

By the way, you're a collaborator now, so you can create a branch directly in the main repository, if you don't want to go through a fork for the next PR. Up to you.

HarrisonKramer:
To make our lives easier, let's just apply this for the on-axis field only (at the primary wavelength). I think this is what you're already doing. This will simplify the mathematics significantly and means we can avoid iterative solutions.

If you find the optimal propagation distance `t` to minimize the radial distance of the rays from the optical axis, then you get the following:
```python
lens = CookeTriplet()
lens.surface_group.surfaces[-1].geometry.cs.z += 1000

Hx = 0  # must remain 0
Hy = 0  # must remain 0
wavelength = lens.primary_wavelength
num_rays = 5
distribution = 'hexapolar'

# Trace rays to the image plane
rays = lens.trace(Hx=Hx, Hy=Hy, wavelength=wavelength, num_rays=num_rays, distribution=distribution)

A = rays.L**2 + rays.M**2
B = rays.L * rays.x + rays.M * rays.y
with np.errstate(divide='ignore', invalid='ignore'):
    t_opt = np.where(A != 0, -B/A, np.nan)
z_focus = np.nanmean(rays.z + t_opt * rays.N)

print(f'Optimal focus distance: {z_focus:.10} mm')

lens.surface_group.surfaces[-1].geometry.cs.z = z_focus

lens.draw()
```

Do you agree with this approach? If so, I would propose you implement this.

Thanks again!
Kramer

optiland/solves.py
SeckinBerkay: No it's not really required. While experimenting, I just set it 50_000 to mimick while condition, while preserving the computation time. It works fine with 5 iterations.

optiland/solves.py
SeckinBerkay: I wanted to make it compatible with the base class to preserve the architecture. To implement `to_dict` and `from_dict`, initializing the `thickness` is required. But if we don't need `to_dict` and `from_dict`, I can totally remove them.

optiland/solves.py
SeckinBerkay: I mentioned it above but I don't want to leave it empty here. Thickness is used for to_dict and from_dict, if we don't have to use them, it makes sense to remove thickness.

optiland/solves.py
SeckinBerkay: Yes exactly, this is the case. I can remove all the stuff related to to_dict, from_dict, and thickness.

SeckinBerkay:
> To make our lives easier, let's just apply this for the on-axis field only (at the primary wavelength). I think this is what you're already doing. This will simplify the mathematics significantly and means we can avoid iterative solutions.
> 
> If you find the optimal propagation distance `t` to minimize the radial distance of the rays from the optical axis, then you get the following:
> 
> ```python
> lens = CookeTriplet()
> lens.surface_group.surfaces[-1].geometry.cs.z += 1000
> 
> Hx = 0  # must remain 0
> Hy = 0  # must remain 0
> wavelength = lens.primary_wavelength
> num_rays = 5
> distribution = 'hexapolar'
> 
> # Trace rays to the image plane
> rays = lens.trace(Hx=Hx, Hy=Hy, wavelength=wavelength, num_rays=num_rays, distribution=distribution)
> 
> A = rays.L**2 + rays.M**2
> B = rays.L * rays.x + rays.M * rays.y
> with np.errstate(divide='ignore', invalid='ignore'):
>     t_opt = np.where(A != 0, -B/A, np.nan)
> z_focus = np.nanmean(rays.z + t_opt * rays.N)
> 
> print(f'Optimal focus distance: {z_focus:.10} mm')
> 
> lens.surface_group.surfaces[-1].geometry.cs.z = z_focus
> 
> lens.draw()
> ```
> 
> Do you agree with this approach? If so, I would propose you implement this.
> 
> Thanks again! Kramer

All right, giving it a shot right now.

optiland/solves.py
HarrisonKramer: Yes, I think totally removing them simplifies the structure, and as far as I can tell, their implementations do not need to differ from the `BaseSolve` class. Please let me know if I misunderstand this though.

SeckinBerkay: You're correct. Thanks for pointing it out.

SeckinBerkay:
Wow, now the optimization works nearly identically whether the misalignment is 10 or 1000. Committing the changed files.

HarrisonKramer (approved):

HarrisonKramer:
Thanks for this contribution, @SeckinBerkay! I appreciate the collaboration. This is a great addition.

Merging now.

Regards,
Kramer

PR #86, opened 2025-03-07
Add comments to surfaces

crnh:
OpticStudio allows to add comments to surfaces, which helps to add information about the function of each surface. When automating model creation, this helps with debugging because it is easier to check if the surfaces were added in the right order. This would be helpful in Optiland as well. I added a `comment` property to the `surface` class, and added this property to the `LensInfoViewer` as well, so the comments also show up when calling `Optic.info`.

## Example

```python

lens = Optic()

lens.set_aperture("EPD", 1)
lens.set_field_type("angle")
lens.add_field(0, 0)
lens.add_wavelength(0.55)

lens.add_surface(index=0, thickness=float("inf"), comment="Object")
lens.add_surface(index=1, radius=10, thickness=5, is_stop=True, material="BK7", comment="Lens front")
lens.add_surface(index=2, radius=-10, thickness=10, comment="Lens back")
lens.add_surface(index=3, thickness=float("inf"), comment="Image plane")

lens.info()
```

Output:
```
╒════╤═════════════════╤═════════════╤══════════╤═════════════╤════════════╤═════════╤═════════════════╕
│    │ Type            │ Comment     │   Radius │   Thickness │ Material   │   Conic │   Semi-aperture │
╞════╪═════════════════╪═════════════╪══════════╪═════════════╪════════════╪═════════╪═════════════════╡
│  0 │ Planar          │ Object      │      inf │         inf │ Air        │       0 │       0.5       │
│  1 │ Stop - Standard │ Lens front  │       10 │           5 │ BK7        │       0 │       0.5       │
│  2 │ Standard        │ Lens back   │      -10 │          10 │ Air        │       0 │       0.414634  │
│  3 │ Planar          │ Image plane │      inf │         nan │ Air        │       0 │       0.0596243 │
╘════╧═════════════════╧═════════════╧══════════╧═════════════╧════════════╧═════════╧═════════════════╛
```

## Considerations

Another possible name for this property would be 'label'.

## Additional remarks

The `LensInfoViewer` currently only prints the table. You could consider returning it instead, so the result can be included in e.g. logs of a program using OptiLand. It would also make `LensInfoViewer` easier to test (i.e. no need to capture output).

### DIFF
--- a/optiland/optic.py
+++ b/optiland/optic.py
@@ -97,7 +97,7 @@ class Optic:
             return self.polarization
 
     def add_surface(self, new_surface=None, surface_type='standard',
-                    index=None, is_stop=False, material='air', **kwargs):
+                    comment='', index=None, is_stop=False, material='air', **kwargs):
         """
         Adds a new surface to the optic.
 
@@ -119,8 +119,8 @@ class Optic:
             ValueError: If index is not provided when defining a new surface.
         """
         self.surface_group.add_surface(
-            new_surface=new_surface, surface_type=surface_type, index=index,
-            is_stop=is_stop, material=material, **kwargs
+            new_surface=new_surface, surface_type=surface_type, comment=comment,
+            index=index, is_stop=is_stop, material=material, **kwargs
             )
 
     def add_field(self, y, x=0.0, vx=0.0, vy=0.0):
--- a/optiland/surfaces/object_surface.py
+++ b/optiland/surfaces/object_surface.py
@@ -21,6 +21,8 @@ class ObjectSurface(Surface):
         geometry (Geometry): The geometry of the surface.
         material_post (Material): The material of the surface after
             interaction.
+        comment (str, optional): A comment for the surface. Defaults 
+            to ''.
 
     Attributes:
         is_infinite (bool): Indicates whether the surface is infinitely
--- a/optiland/surfaces/object_surface.py
+++ b/optiland/surfaces/object_surface.py
@@ -27,13 +27,14 @@ class ObjectSurface(Surface):
             far away.
     """
 
-    def __init__(self, geometry, material_post):
+    def __init__(self, geometry, material_post, comment=''):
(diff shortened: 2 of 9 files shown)


HarrisonKramer (approved):

HarrisonKramer:
Hi @crnh,

Looks great, thanks for submitting this PR! I also really appreciate that you added tests to cover the new functionality.

Regarding your last remark - this is a very good point. I agree it makes sense to return the table instead of just printing it. I will make a note to implement this. It's a quick change, but I'll need to update all the examples as well.

Merging this now. Thanks again.

Kramer

crnh:
> Merging this now. Thanks again.

That was quick, thanks!

> I agree it makes sense to return the table instead of just printing it. (...). It's a quick change, but I'll need to update all the examples as well.

Just to be clear, I think printing the table still makes sense, as I think the function will mostly be used interactively. In that case, modifying the examples shouldn't be necesary. Returning the dataframe would mainly be useful for other packages / libraries depending on Optiland. For example, I expect it would be useful in [visisipy](https://github.com/MREYE-LUMC/visisipy). For this use case, you could also add a parameter to disable printing.

HarrisonKramer:
Okay, that's clear. Maybe adding a `return_dataframe=False` (or similar) argument makes sense. I would also want to avoid, e.g., that running "lens.info()" in a Jupyter notebook cell first prints the table, then returns the table, possibly printing it a second time. In any case, I am open to any solution here, with a preference on changes that don't impact existing code.

HarrisonKramer:
Hi @crnh,

By the way, another collaborator set up a discord server for Optiland contributors (or users). I'm inviting all current contributors. The goal of the discord is to make it easier to discuss Optiland-related topics/planning, e.g., for topics that don't quite fit in github issues/discussions. Feel free to join, if you wish.

https://discord.gg/VPrwKR9S

Regards,
Kramer

PR #90, opened 2025-03-25
add draw3D inside jupyter functionality

SeckinBerkay:
Achieved drawing inside Jupyter, preventing the next code blocks from interruption. Used ipyvtklink, but I am not satisfied with its performance. Currently, the algorithm is controlled by a boolean input `is_jupyter` by the user, automatic detection of jupyter is straightforward but I kept it as a boolean for reproducibility for now. I don't know if we should move on from this sample, or just find a way to prevent 3D viewer interruption, or check if the default 3D viewing still interrupts in a QT environment.

Here are two examples to run, one static (I expected it to be interactive), and one interactive:

Starting with the imports and defining the lens system.
```python
from optiland.samples.objectives import CookeTriplet
from optiland.visualization.visualization import OpticViewer3D

lens = CookeTriplet()
```

This creates a static 3D visualization 
```python
drawer = lens.draw3D(is_jupyter=True)
drawer  # This will display the static visualization
```

and this creates an interactive but laggy 3D visualization

```python
viewer_3d = OpticViewer3D(lens)
widget = viewer_3d.view(is_jupyter=True)
widget  # This will display the interactive visualization
```

You can check if other codes are interrupted by adding and running another line:

```python
print("x")
```

### DIFF
--- a/optiland/optic.py
+++ b/optiland/optic.py
@@ -97,7 +97,7 @@ class Optic:
             return self.polarization
 
     def add_surface(self, new_surface=None, surface_type='standard',
-                    comment='', index=None, is_stop=False, material='air', **kwargs):
+                    index=None, is_stop=False, material='air', **kwargs):
         """
         Adds a new surface to the optic.
 
@@ -119,8 +119,8 @@ class Optic:
             ValueError: If index is not provided when defining a new surface.
         """
         self.surface_group.add_surface(
-            new_surface=new_surface, surface_type=surface_type, comment=comment,
-            index=index, is_stop=is_stop, material=material, **kwargs
+            new_surface=new_surface, surface_type=surface_type, index=index,
+            is_stop=is_stop, material=material, **kwargs
             )
 
     def add_field(self, y, x=0.0, vx=0.0, vy=0.0):
@@ -324,7 +324,7 @@ class Optic:
 
     def draw3D(self, fields='all', wavelengths='primary', num_rays=24,
                distribution='ring', figsize=(1200, 800), dark_mode=False,
-               reference=None):
+               reference=None, is_jupyter=False):
         """
         Draw a 3D representation of the optical system.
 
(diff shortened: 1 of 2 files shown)


HarrisonKramer:
Hi Seckin. I see that this is a draft PR, so I won't formally review it yet. Nonetheless, have a few thoughts here.

- I feel strongly that we should not add another dependency (i.e., `ipyvtklink`). Some users may not even be using Jupyter notebooks, so this will not be useful for them. I'd recommend that we have a try/catch block that attempts to load `ipyvtklink` and falls back to the default behavior if it's not installed. We can just point out in the docs that you can install `ipyvtklink` for interactive plotting in a Jupyter notebook.
- I see that `ipyvtklink` is archived, which is somewhat worrying. It could become incompatible with Optiland in the future if it's not maintained. I wonder if there are alternatives here.
- Similar to first point, but I prefer not to change the existing code in any way (e.g., new argument to `optiland.Optic.draw3D`). Ideally, the code handles visualization in the background and the API does not change for users.

Just some thoughts here. Curious to know your thoughts.

Regards,
Kramer

SeckinBerkay:
Thanks for the review, I didn't know `ipyvtklink` was archived. I'm not a fan of it anyway, it is laggy. Considering its lack of performance and it is archived, I think the same way, we should not include `ipyvtklink`. Just wanted to demonstrate what `ipyvtklink` may provide us here, seems not to be much. If it is all right, I will close the counterpart issue, and this PR if. Next, I will test if 3D functionality interrupts other functionalities in QT environment, so that we will know the current implementation is safe.

crnh:
Hi both,

Just stumbled upon this Pull Request. I am using Jupyter myself and would like this functionality, but not having it would not be a deal breaker and I agree with @HarrisonKramer's arguments. However, I'd like to add two comments that may help achieve this functionality in a different way that would address the points raised by @HarrisonKramer:

1. Like [PyVista](https://pyvista.org), you could use [Trame](https://kitware.github.io/trame) for interactive visualizations. This library is maintained and it supports Jupyter notebooks.
2. Instead of adding a default dependency on Trame (or whatever library you choose), make it an [optional dependency](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies). It can then, for example, be installed with `pip install optiland[jupyter]`.
3. Instead of adding a parameter to indicate if the notebook plotting backend should be used, it would be better to automatically detect if the code is run in a notebook.

For 1 and 3, much inspiration could be drawn from how PyVista handles this in [`pyvista.Plotter.show`](https://github.com/pyvista/pyvista/blob/5ea48c7d3cab0cd51372c770841b1d73ecba4873/pyvista/plotting/plotter.py#L6763).
Unfortunately, I currently do not have the time to assist with this, but if this helps to implement the proposed functionality, that would be great!

PR #99, opened 2025-04-01
Reciprocal radius

mirandadam:
This code adds the "reciprocal_radius" optimization variable so that the optimizers can use a continuous transition from a positive curvature surface to a negative curvature surface. No new attributes are created on the surfaces themselves.

A simple test case and an example are added.

This still does not work well on surfaces where the radius is set at exactly +infinity or -infinity, and I have no idea on proper "scaling values", so the code requires review. Also I do not have clarity whether this approach is the best one to make optimization continuous in near-flat surfaces - it works, but there may be a better way.

This pull request is proposed mostly as a suggestion, as this feature is useful to me.

### DIFF
--- a/docs/api/api_optimization.rst
+++ b/docs/api/api_optimization.rst
@@ -1,9 +1,9 @@
 Optimization
 ============
 
-This section covers the optimization functionality of the Optiland package. The optimization module 
-provides a set of tools for optimizing optical systems. It includes a variety of optimization algorithms, 
-such as gradient-based and evolutionary algorithms, as well as tools for defining optimization variables 
+This section covers the optimization functionality of the Optiland package. The optimization module
+provides a set of tools for optimizing optical systems. It includes a variety of optimization algorithms,
+such as gradient-based and evolutionary algorithms, as well as tools for defining optimization variables
 and objectives.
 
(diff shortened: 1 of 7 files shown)


mirandadam:
Ok, I got your CI feedback. I will fix the PR as soon as possible.

HarrisonKramer:
Hi @mirandadam,

Thanks for submitting this! A very nice addition. I don't think there are any serious issues. Hopefully a quick fix. I recently added the linting checks to the CI pipeline - you should be able to automatically fix these with Ruff.

For the scaling, this is meant to bring variables into similar ranges (a sort of normalization) to improve optimization. I should generalize this functionality eventually, but I've so far targeted a scaling that brings values somewhere in the ~(-1, 1) range. This does not need to be perfect, and many optimizers are robust without scaling regardless.

Kramer

HarrisonKramer:
Hi Daniel,

As of recently, I invite all contributors to a discord channel we have to discuss Optiland topics/planning. It's just a place that's a bit easier to chat than on github. So far it's been a bit quiet, but we'll see how it evolves. Feel free to join if you like!

https://discord.gg/NVYqvBQz

Regards,
Kramer

PR #117, opened 2025-04-19
Physical Apertures

SeckinBerkay:
Added set_test_backend to corresponding test functions, excluding some functions like `test_view` in `TestRadialAperture` class. 
Includes some changes such as conversion of `list` to `be.array`, and `be.testing.assert_array_equal` to `assert be.all``, since the former assertion resulted in errors.


HarrisonKramer:
Thanks a lot for writing this, @SeckinBerkay. Merging now!

PR #152, opened 2025-05-14
bugfix: new logic for infinite radius in _intersection_sphere method

manuelFragata:
**Fix**: Handle infinite radius in NewtonRaphsonGeometry._intersection_sphere

This commit addresses an issue in `NewtonRaphsonGeometry._intersection_sphere`
where an infinite `self.radius` could lead to `UnboundLocalError` or
propagation of `NaN` values.

When `self.radius` is infinite, the method now correctly calculates the
initial ray intersection with a tangent plane (z=0) instead of attempting
to use the standard spherical intersection formula. This prevents the
definition and use of variables (a, b, c, d, t1, t2) that are only
valid for finite radii, thus resolving the `UnboundLocalError`.

This change ensures more robust behavior for surfaces like toroidal lenses
when defined with an infinite primary radius of curvature.

Affected files:
- `optiland/geometries/newton_raphson.py`


.coverage.Manuel.21176.XCcVPHQx
HarrisonKramer, line 1: I guess this was included by accident. can you delete?

.coverage.Manuel.22244.XItjeWJx
HarrisonKramer, line 1: delete me

.coverage.Manuel.3836.XdMUHuPx
HarrisonKramer, line 1: delete me

optiland/geometries/newton_raphson.py
```diff
@@ -118,34 +118,61 @@ def _intersection_sphere(self, rays):
...
-
-        # take intersection closest to z = 0 (i.e., vertex of geometry)
-        t = be.where(be.abs(z1) <= be.abs(z2), t1, t2)
-
-        # handle case when a = 0
-        cond = a == 0
-        t[cond] = -c[cond] / b[cond]
+        inf_radius = False
+
+        if hasattr(self.radius, "item"):
```
HarrisonKramer: this group of conditions is very difficult to follow. Is there any way to simplify it? if not, then put it into a utility function at the start of the file and call it here instead. That keeps things cleaner.

optiland/geometries/newton_raphson.py
```diff
@@ -118,34 +118,61 @@ def _intersection_sphere(self, rays):
...
+            try:
+                inf_radius = bool(be.isinf(self.radius).item())
+            except TypeError:
+                inf_radius = bool(be.all(be.isinf(self.radius)))
+        elif hasattr(self.radius, "size") and self.radius.size > 1:
+            inf_radius = bool(be.isinf(self.radius[0]).item())
+        else:
+            inf_radius = bool(be.isinf(self.radius))
+
+        if inf_radius:
```
HarrisonKramer: the logic here no longer matches the description of the function, i.e., "intersection of rays with a sphere". I would propose you rework the class slightly:
1. move the plane calculation into `_intersection_plane`
2. move all the sphere intersection code into `_intersection_sphere`
3. Add a new function `_intersection` that checks if it's a plane and calls the appropriate intersection function

IMO, that makes the code cleaner and more maintainable.

optiland/geometries/toroidal.py
```diff
@@ -47,14 +47,18 @@ def __init__(
         tol: float = 1e-10,
         max_iter: int = 100,
     ):
-        # Pass radius_rotation as the base 'radius' for NewtonRaphsonGeometry.
+        # convert scalar parameters to backend arrays at initialization
+        _radius_rotation_be = be.array(radius_rotation)
```
HarrisonKramer: what's the benefit of creating a new variable and not just modifying the variable, e.g., `radius_rotation = be.array(radius_rotation)`? If it makes sense, I wouldn't create a new variable. Up to you. Not critical, of course

HarrisonKramer (changes requested):
Looks solid! Thank you for identifying and fixing this. I made a few requests for changes, but I'm open to discussion if you disagree. There's nothing too critical.

Thanks!
Kramer

optiland/geometries/toroidal.py
manuelFragata: 100% agree with you. Tbh it was just a temporary solution that later I forgot to refactor.

optiland/geometries/newton_raphson.py
manuelFragata: yeah, this is I wanted to wait for your review as well, cause I had the same thought. Will do so, as it makes more sense

optiland/geometries/newton_raphson.py
manuelFragata: will think about it, and keep you updated

manuelFragata:
Hi @HarrisonKramer ,

Thanks for the thorough review! Will work on the changes

HarrisonKramer (approved):

HarrisonKramer:
Looks great! Thanks for fixing this!

Merging now.

Kramer

PR #171, opened 2025-05-30
Improvement on LeastSquares Optimization

manuelFragata:
Refactor(optim): Optimize LeastSquares with dedicated residuals vector method

Refactored the `LeastSquares` optimizer to use its own internal
`_compute_residuals_vector` method. This method is responsible for:
1. Updating Optiland problem variables from SciPy's current iteration.
2. Updating the optical system (e.g., for pickups, solves).
3. Computing the vector of weighted residuals (`[op.fun() for op in problem.operands]`).
4. Handling potential NaNs or exceptions by returning a high-penalty error vector.
5. Ensuring the final output is a NumPy array, as required by `scipy.optimize.least_squares`.

`LeastSquares` manages its residuals calculation directly. This is done to address observed performance
degradations from alternative refactorings where the residuals vector
calculation was more deeply integrated into the `OptimizerGeneric`'s primary
scalar evaluation path (`_fun`).

The `OptimizerGeneric._fun` method remains optimized for scalar optimizers,
calculating the sum-of-squares directly without an intermediate vector step,
thus preserving their performance. `LeastSquares` now uses its own efficient
path for vector-based optimization.

### DIFF
--- a/optiland/optimization/optimization.py
+++ b/optiland/optimization/optimization.py
@@ -272,69 +272,127 @@ class OptimizerGeneric:
 
 
 class LeastSquares(OptimizerGeneric):
-    """LeastSquares optimizer class for solving optimization problems using the
-    least squares method.
-
-    Args:
-        problem (OptimizationProblem): The optimization problem to be solved.
-
-    Attributes:
-        problem (OptimizationProblem): The optimization problem to be solved.
-
-    Methods:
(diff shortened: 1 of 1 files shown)


HarrisonKramer:
Looks good, thanks!

You can merge when ready.

Kramer

HarrisonKramer (approved):

PR #173, opened 2025-05-30
Feature / add_wavelengths

kkrutkowski:
Here is the implementation of the "add_wavelengths" functionality which finally doesn't seem to violate any of the developer guidlines (or at least I hope so).

One issue I have with it is the overconstrained default method, although I'm not able to find anything remotely correct with less constraints than that.

As for now the default combination of arguments should select the wavelengths approximately optimizing maximum of the merit function if;

1. The merit function is approximately a linear function of dispersion
2. The [Cautchy's equation](https://en.wikipedia.org/wiki/Cauchy%27s_equation) truncated to two terms accurately 
3. The weights corresponding to all wavelengths are equal

Or at least that's the case if my reasoning isn't incorrect (and it definitely may be).

While I'm not sure if the proposed method is correct it seems to consistently outperform the default 'Gaussian quadrature' spacing in the Zemax OpticStudio.

Here are example values for 102 mm f/6 'Frauhoffer' doublet (S-FPL55 + N-KZFS2 glass), optimized for RMS across 400-800 nm range - for both results the designs were performing the worst at the edges of intervals.

> add_wavelengths with default settings for 5 wavelengths;
>     400 nm - 10.185 µm RMS, 28.049 µm geometric radious
>     800 nm - 9.375 µm RMS, 14.752 µm geometric radious

    

> Zemax's 'gaussian quadrature' (6 wavelengths, as the number has to be even);
>     400 nm - 17.012 µm RMS, 42.515 µm geometric radious
>     800 nm - 7.141 µm RMS, 10.748 µm GEO

    
    
Generally it seems to behave as expected with resulting RMS being slightly higher at the longer wavelength edge of the spectrum (as a result of higher-order terms), however the results seem to consistently fall pretty close to each other. The difference between geometric radii here is a result of spherochromatism and it doesn't concern me too much (if optimized for geometric spot size result is comparable GEO, but significant RMS discrepancy).

### DIFF
--- a/optiland/backend/torch_backend.py
+++ b/optiland/backend/torch_backend.py
@@ -160,6 +160,30 @@ def linspace(start, stop, num=50):
     )
 
 
+def arange(*args, step=1):
+    if len(args) == 1:
+        start = 0
+        end = args[0]
+    elif len(args) == 2:
+        start, end = args
+
+    if isinstance(start, torch.Tensor):
+        start = start.item()
+    if isinstance(end, torch.Tensor):
+        end = end.item()
+    if isinstance(step, torch.Tensor):
+        step = step.item()
+
+    return torch.arange(
+        start,
+        end,
+        step,
+        device=get_device(),
+        dtype=get_precision(),
+        requires_grad=grad_mode.requires_grad,
+    )
+
+
 def zeros_like(x):
     return torch.zeros_like(
         array(x),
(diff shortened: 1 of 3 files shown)


HarrisonKramer (approved):
Looks good! Thanks a lot for implementing this and for adding comprehensive tests.

PR #182, opened 2025-06-02
Add a live demo to the documentation

crnh:
Add a live demo to the documentation using [jupyterlite-sphinx](https://jupyterlite-sphinx.readthedocs.io/en/stable/).
Includes a custom Python kernel with Optiland and its dependencies. This requires the ReadTheDocs build to use conda instead of the normal Python configuration.
Please let me know if this breaks your local development workflow.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -3,11 +3,10 @@ version: 2
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.11"
+    python: "mambaforge-23.11"
 
-python:
-  install:
-    - requirements: docs/requirements.txt
+conda:
+  environment: docs/build-environment.yml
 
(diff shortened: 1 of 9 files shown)


crnh:
The demo works:

image

The contents are still quite basic and open for discussion. @HarrisonKramer do you have any ideas?

crnh:
- I added a note about this. Luckily, loading is / should be faster than the visisipy demo, because the visisipy demo runs in Pyodide and needs to install all packages, whereas this demo includes all packages in a pre-built kernel.
- Makes sense, NumPy does something similar (see https://numpy.org/).
- Yes, this could be used in a walkthrough, but you can also use it to make entire notebooks interactive. For now, I'd focus on a simple demo on the front page.

PR #193, opened 2025-06-10
Mimic OpticStudio sampling behavior in FFT PSF

crnh:
Following the discussion in #157, this is a proposal to make Optiland by default mimic the sampling behavior in OpticStudio.
This behavior is documented here: https://ansyshelp.ansys.com/public/account/secured?returnurl=/Views/Secured/Zemax/v251/en/OpticStudio_User_Guide/OpticStudio_Help/topics/FFT_PSF.html

This behavior can be overriden by specifying both `num_rays` and `grid_size`. If only `num_rays` is specified, the grid size will be based on `num_rays`. Furthermore, `num_rays` will be set to the 'effective pupil sampling' as specified in the OpticStudio documentation.

The current implementation changes the number of rays used for pupil sampling. I investigated a different approach where `num_rays` is interpreted as the effective pupil sampling and the pupil sampling and grid size are calculated from that value. However, due to rounding of the effective values, I found it difficult to obtain the pupil sampling from the effective pupil sampling, which made it impossible to replicate OpticStudio's behavior with this alternative approach.

### DIFF
--- a/optiland/psf/fft.py
+++ b/optiland/psf/fft.py
@@ -115,7 +115,7 @@ class FFTPSF(BasePSF):
         x, y = be.meshgrid(x, x)
         x = x.ravel()
         y = y.ravel()
-        R = be.sqrt(x**2 + y**2)
+        R2 = x**2 + y**2
 
         field = self.fields[0]  # PSF contains a single field.
         pupils = []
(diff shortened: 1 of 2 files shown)


crnh:
This PR depends partly on #192 and tests will only pass after #192 has been merged.

HarrisonKramer:
@crnh - I finally found the issue, which seemed to be due to floating point errors. In `UniformDistribution`, the bounds of the unit circle are determined based on `radius**2`, so points are considered in the pupil for `radius**2 <= 1.0`. In the `FFTPSF` class, the generate pupils function instead uses `radius <= 1.0`. I fixed this in `FFTPSF` and see that there is no longer an index mismatch and that the failing test now passes.

HarrisonKramer:
All looks good. Thanks for updating this! Merging now.

Kramer

Issue #195, opened 2025-06-11
Circular import in optiland.mtf


crnh:
### Checklist

- [x] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [x] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [x] I am using the **latest version** of Optiland (if not, please update and retry).
- [x] I have tried to reproduce or debug the issue myself before opening this.
- [x] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

*Thanks for taking the time to go through this — it really helps us help you!*

### Bug Report

**Describe the bug**
There is a circular import in `optiland.mtf` which makes it impossible to import members from that module.

**To Reproduce**
One-liner to check this with the latest development version:

```bash
uv run --with git+https://github.com/HarrisonKramer/optiland.git python -c "from optiland.mtf import FFTMTF"
``` 

Output:

```bash
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "D:\software\uv\cache\archive-v0\b3PhKhKP9SOZdfZdlVDne\Lib\site-packages\optiland\mtf.py", line 12, in <module>
    from optiland.analysis import SpotDiagram
  File "D:\software\uv\cache\archive-v0\b3PhKhKP9SOZdfZdlVDne\Lib\site-packages\optiland\analysis\__init__.py", line 13, in <module>
    from .through_focus_mtf import ThroughFocusMTF
  File "D:\software\uv\cache\archive-v0\b3PhKhKP9SOZdfZdlVDne\Lib\site-packages\optiland\analysis\through_focus_mtf.py", line 16, in <module>
    from optiland.mtf import SampledMTF
ImportError: cannot import name 'SampledMTF' from partially initialized module 'optiland.mtf' (most likely due to a circular import) (D:\software\uv\cache\archive-v0\b3PhKhKP9SOZdfZdlVDne\Lib\site-packages\optiland\mtf.py)
```

**Expected behavior**
I can import `FFTMTF` without problems.

**Environment**  
- Optiland Version: master branch
- Python Version: `3.12`
- OS: `[Windows]`
- Additional Info:

**Additional context**

crnh:
[...]

PR #196, opened 2025-06-11
Use FFT PSF sampling behavior in FFT MTF

crnh:
Change the default sampling behavior of the FFT MTF to be the same as the default FFT PSF behavior.

When making these changes, I also noticed a circular import, see #195. As a result, I couldn't test these changes (I could add them to the tests but not perform some simulations myself).


HarrisonKramer:
Hi @crnh,

I made a small change, which I think fixes the units, I did not test this extensively yet. Do you want to double check the implementation to confirm it gives expected results?

Kramer

crnh:
Apologies for the late response, it took a while before I had time to check this. I reran the example code from my previous comment with the current implementation and for all tested sampling settings, the problem with the MTF domain (and `dx` value) appeared to be fixed. Thanks!

PR #199, opened 2025-06-12
Fix/psf image space index

### LINKED ISSUE #186: PSF dimension calculation does not correct for image space refractive index
### Checklist

### Bug Report

**Describe the bug**

I  was comparing the FFT PSF extent between Optiland and OpticStudio and got different extents for systems with a refractive index $n \neq 1$ before the image surface. Optiland calculates the PSF dimensions based on (among others) the working F/#. OpticStudio reports the working F/# in air and also seems to use this value to calculate the PSF dimensions. Optiland uses the working F/# in the image space medium, which results in different PSF dimensions.

To be honest, I am not sure what is correct here. Definitions of the working F/# use the focal length, whereas OpticStudio seems to use the effective focal length. If you think OpticStudio is incorrect here, let me know.

**To Reproduce**

1. Create a system with a medium other than air in image space
2. Calculate the FFT PSF
3. Compare the extents with the FFT PSF of the same system in OpticStudio (see #157 for the settings needed to replicate OpticStudio's sampling behavior)

**Expected behavior**

The extents of both PSFs are (almost) the same

**Environment**  
- Optiland Version: `0.4.1`
- Python Version: `3.12`
- OS: `Windows`
- Additional Info:

**Additional context**

HarrisonKramer:
- Updates PSF calculation to use working F/# based on real marginal rays

Updated algorithm:
1. Retrieve the defined given wavelength and field coordinates.
2. Determine the image-space refractive index 'n' at the given wavelength.
3. Trace four marginal rays (top, bottom, left, right) at the pupil edges,
   as well as the chief ray.
4. Compute the angle between each marginal ray and the chief ray.
4. Calculate the average of the squared numerical apertures from all traced
   marginal rays.
5. Compute the working F-number as 1 / (2 * sqrt(average_NA_squared)).
6. Cap the calculated F/# at 10,000 if it exceeds this value.

Fixes #186

### DIFF
--- a/optiland/psf/base.py
+++ b/optiland/psf/base.py
@@ -364,9 +364,11 @@ class BasePSF(Wavefront):
         single defined field point and given wavelength.
 
         Algorithm:
-            1. Retrieve the defined given wavelength.
+            1. Retrieve the defined given wavelength and field coordinates.
             2. Determine the image-space refractive index 'n' at the given wavelength.
-            3. Trace four marginal rays (top, bottom, left, right) at the pupil edges.
+            3. Trace four marginal rays (top, bottom, left, right) at the pupil edges,
+               as well as the chief ray.
+            4. Compute the angle between each marginal ray and the chief ray.
             4. Calculate the average of the squared numerical apertures from all traced
                marginal rays.
             5. Compute the working F-number as 1 / (2 * sqrt(average_NA_squared)).
@@ -377,13 +379,12 @@ class BasePSF(Wavefront):
         """
         MAX_FNUM = 10000.0
 
-        Hx, Hy = 0, 0
+        Hx, Hy = self.fields[0]
         wavelength = self.wavelengths[0]
 
         n = self.optic.image_surface.material_post.n(wavelength)
         Px = be.array([0, 0, 0, 1, -1])
         Py = be.array([0, 1, -1, 0, 0])
-        numerical_apertures_squared = []
 
         rays = self.optic.trace_generic(
             Hx=Hx, Hy=Hy, Px=Px, Py=Py, wavelength=wavelength
--- a/optiland/psf/base.py
+++ b/optiland/psf/base.py
@@ -404,4 +404,9 @@ class BasePSF(Wavefront):
         if fno > MAX_FNUM:
             fno = MAX_FNUM
 
+        if be.isnan(fno):
+            raise ValueError(
+                "Working F/# could not be calculated due to raytrace errors."
+            )
+
         return fno
--- a/optiland/psf/base.py
+++ b/optiland/psf/base.py
@@ -359,18 +359,21 @@ class BasePSF(Wavefront):
         center_y = self.psf.shape[1] // 2
         return self.psf[center_x, center_y] / 100
 
-    def _get_effective_FNO(self):
-        """Calculates the effective F-number of the optical system.
+    def _get_working_FNO(self):
+        """Calculates the working F-number of the optical system.
+
+        Algorithm:
+            1. Trace real marginal ray for on-axis field point.
+            2. Determine the angle of the chief ray in angle space.
+            3. Retrieve the image-space refractive index at the primary wavelength.
+            3. Compute the working FNO as 1 / (2 * n * sin(theta))
 
         Returns:
-            float: The effective F-number.
+            float: The working F-number.
         """
-        FNO = self.optic.paraxial.FNO()
+        wavelength = self.optic.primary_wavelength
+        rays = self.optic.trace_generic(Hx=0, Hy=0, Px=0, Py=1, wavelength=wavelength)
+        sin_theta = be.abs(rays.M)
+        n = self.optic.image_surface.material_post.n(wavelength)
 
-        if not self.optic.object_surface.is_infinite:
-            D = self.optic.paraxial.XPD()
-            p = D / self.optic.paraxial.EPD()
-            m = self.optic.paraxial.magnification()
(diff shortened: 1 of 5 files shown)


HarrisonKramer:
I also see I broke some functionality related to rays. I will investigate this over the next day. Should be a simple fix.

HarrisonKramer:
Hi @crnh,

Thanks for making the comparison!

I have several remarks:
- Yes, you're right that the pre-material on the image surface should be set as well. That detail affects ray tracing and can influence the marginal ray angle, which in turn can affect the working F/# indirectly. The image-space refractive index, however, enters directly into the F/# calculation and had previously been omitted. This led to incorrect changes in the PSF extent when the image-space index varied. That’s now fixed.
- On the extent mismatch: I noticed that Optiland consistently gave slightly larger extents than ZOS. After checking the ZOS docs, I found they apply a scaling factor due to the pupil not being centered on the FFT grid. Specifically, for even n, they seem to apply a factor of `(n - 2) / n`. For `n = 128`, this gives `126/128 = 0.984375`, which aligns the on-axis extent "exactly" (172.49 µm). Applying this to off-axis points brings the values closer, but doesn't fully match, so there might be other factors at play.
- While the on-axis result for the Cooke triplet suggests the working F/# calculation is correct, I was less confident about the off-axis cases. In theory, the PSF extent should scale with the working F/#, so the ratio of extents across field positions in ZOS should match the ratio of working F/#s, regardless of whether they're from Optiland or ZOS. However, when I compare the F/# ratios from Optiland to the extent ratios in ZOS, they don’t line up. This suggests there may still be an issue in the F/# computation. I checked the system again in ZOS, which reports both tangential and sagittal working F/#s per wavelength, and it looks like the sagittal value is consistently used for extent calculations. Do you happen to know why that is? It’s still a bit unclear to me.

Here's a quick comparison using the Cooke triplet in air:
Field        | Optiland F/# | Optiland Extent, including (n-2)/n factor | Zemax Sag. F/# | Zemax Extent | Optiland Extent (Zemax F/# + (n-2)/n factor)
-------------|--------------|-----------------|----------------|---------------|---------------------------------------
Field 1 (on-axis) | 4.9781      | 172.4908        | 4.9782         | 172.49        | 172.4908
Field 2      | 5.2200      | 180.8736        | 5.1130         | 177.17        | 177.1655
Field 3      | 5.4799      | 189.8786        | 5.2837         | 183.08        | 183.0802

So using Zemax’s sagittal F/#s and applying the (n-2)/n correction gives the same extents you reported.

I'm still investigating the off-axis F/# issue and whether our calculation should match the sagittal one. Let me know if you have any ideas.

Next steps:
- Double-check F/# computation logic for off-axis fields
- Confirm whether sagittal F/# should always be used for extent scaling

Thanks,
Kramer

crnh:
> On the extent mismatch: I noticed that Optiland consistently gave slightly larger extents than ZOS. After checking the ZOS docs, I found they apply a scaling factor due to the pupil not being centered on the FFT grid. Specifically, for even n, they seem to apply a factor of (n - 2) / n. For n = 128, this gives 126/128 = 0.984375, which aligns the on-axis extent "exactly" (172.49 µm) for the Cooke triplet in air. Applying this to off-axis points brings the values closer, but doesn't fully match, so there might be other factors at play.

Great, good to see that incorporating the factor $\frac{n - 2}{2n}$ results in equal extents for on-axis fields.

> While the on-axis result for the Cooke triplet suggests the working F/# calculation is correct, I was less confident about the off-axis cases. In theory, the PSF extent should scale with the working F/#, so the ratio of extents across field positions in ZOS should match the ratio of working F/#s, regardless of whether they're from Optiland or ZOS. However, when I compare the F/# ratios from Optiland to the extent ratios in ZOS, they don’t line up. This suggests there may still be an issue in the F/# computation. I checked the system again in ZOS, which reports both tangential and sagittal working F/#s per wavelength, and it looks like the sagittal value is consistently used for extent calculations. Do you happen to know why that is? It’s still a bit unclear to me.

The values in the table indeed seem to justify the conclusion that Zemax consistently uses the sagittal F/# to calculate the extent. However, your later comment

> Could it be the case that ZOS uses the sagittal F/# for the x extent and the tangential F/# for the y extent?

could indeed be true. The term 'sagittal' also has a meaning in anatomy, and as a result it would make more sense to me if the sagittal plane would correspond to the y extent. However, if I understand the optical definitions correctly, it makes sense to call the x direction 'sagittal' if the y direction is the primary direction, which is the case in OpticStudio. As OpticStudio only reports the extent in a single direction, it is not possible to check this. The plots do not suggest that OpticStudio uses different extents and the fact that OpticStudio expresses the extent as "Data area is ... µm wide" may suggest that it uses a single extent and calculates it in the x direction. I doubt if it would make sense to follow OpticStudio here.

I noticed you now calculate Q using `num_rays - 1`, why is that? It seems likely that OpticStudio indeed traces $N - 1$ rays (in a single direction) through the pupil for a sampling of $N \times N$ where $N$ even, so it always samples the pupil center. However, as far as I know, Optiland will not 'correct' for not sampling the pupil center and trace $N$ rays through the pupil.

crnh:
> If we do want to enforce an odd number of rays, I’d suggest doing so only in the PSF classes (or anywhere else it’s actually critical). I don’t think it’s necessary to enforce it in the Wavefront class itself, though I’m open to it.

In my opinion it makes sense to do this only in the PSF classes. In the future we may encounter other cases where odd samplings are required, but for e.g. the Zernike decomposition, the current approach is clearly sufficient; those values match closely with OpticStudio's output.

Issue #206, opened 2025-06-27
Error / difference with OpticStudio in F1 calculation


crnh:
### Checklist

- [x] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [x] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [x] I am using the **latest version** of Optiland (if not, please update and retry).
- [x] I have tried to reproduce or debug the issue myself before opening this.
- [x] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

*Thanks for taking the time to go through this — it really helps us help you!*

### Bug Report

**Describe the bug**

I am using Optiland in [Visisipy](https://github.com/mreye-lumc/visisipy) to calculate the cardinal points of the human eye. Visisipy also supports performing this calculation in OpticStudio, and provides a comparison in its documentation: https://visisipy.readthedocs.io/en/latest/examples/Backend%20comparison/backend_comparison.html

I discovered this comparison shows a substantial difference in the object space cardinal points. The difference is the same for all points. Because all other object-space cardinal points depend on `F1`, I suspect there is an error in the calculation of `F1`.

I did not manage to reproduce this with simple systems (e.g. a thick lens (even not with the stop behind the lens) or a Cooke triplet), so I attached an Optiland file with an eye model to this issue.

**To Reproduce**

I used the following code to check the focal length (`f1`) and focal point (`F1`). I also included a thick lens as a reference, but the problem doesn't occur in that system:

```python
from optiland.fileio import load_optiland_file
from optiland.materials import IdealMaterial
from optiland.optic import Optic

eye = load_optiland_file("optiland_model.json")
eye.info()

print("Eye:")
print(" f1: ", eye.paraxial.f1())
print(" F1: ", eye.paraxial.F1())

thick_lens = Optic()
thick_lens.set_aperture("EPD", 1.0)
thick_lens.set_field_type("angle")
thick_lens.add_field(0)
thick_lens.add_wavelength(0.543)

thick_lens.add_surface(index=0, thickness=float("inf"))
thick_lens.add_surface(index=1, thickness=10, radius=10, material=IdealMaterial(1.5))
thick_lens.add_surface(index=2, thickness=0, radius=-10)
thick_lens.add_surface(index=3, thickness=10, is_stop=True)
[...]

PR #209, opened 2025-07-02
Skip image surface when calculating front focal point

crnh:
The front focal point calculation is based on a reverse ray trace that currently starts behind the image surface.
This results in incorrect values if the image surface is curved and has a refractive index $n \neq 1$, because the image surface then refracts the light (see #206).
Furthermore, the paraxial ray trace assumes $n = 1$ behind the image surface, which causes similar problems.
This PR addresses both issues:

- The image surface is skipped in the paraxial ray trace in `Paraxial.F1()`;
- The paraxial ray tracer now assumes the refractive index before the first surface is the same as the index after it (i.e. it prepends `n[0]` to `n` when calculating the power).

### DIFF
--- a/optiland/paraxial.py
+++ b/optiland/paraxial.py
@@ -83,7 +83,7 @@ class Paraxial:
         # start tracing 1 lens unit before first surface
         z_start = -1
         wavelength = self.optic.primary_wavelength
-        y, u = self._trace_generic(1.0, 0.0, z_start, wavelength, reverse=True)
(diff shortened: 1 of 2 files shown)


crnh:
I plan to add a sample system that is currently affected by this issue in a separate PR; this will include unit tests.

HarrisonKramer:
> I plan to add a sample system that is currently affected by this issue in a separate PR; this will include unit tests.

Sounds good.

PR looks good. I will go ahead and merge. Thanks!

PR #210, opened 2025-07-03
Add tabulate to demo dependencies

crnh:
`tabulate` is currently not present in the demo's kernel. As a results, `Optic.info()` fails.


HarrisonKramer (approved):

crnh:
Okay, installing `tabulate` with pip works. It doesn't look great because the lines are wrapped, but at least it doesn't raise an exception.

PR #211, opened 2025-07-04
Glass Expert

drpaprika:
Hi @Harrison, all,

It's been some time since my last contribution — it's nice to see that Optiland is continuously growing.

I've been working on an optimization feature called `GlassExpert`, which adds support for the optimization of categorical glass variables. 
It explores a user-defined list of glasses and iteratively tests alternatives to find better combinations. 
This is a powerful tool used by CodeV and Zemax as selling points, and I believe Optiland could benefit from this functionality as well.

The main challenge was to integrate discrete choices (glass materials) into the existing continuous optimization framework. 
To do this, I implemented a greedy search strategy that cycles through all glass surfaces marked as variables, 
evaluates options in (n_d, V_d) space, and keeps any substitution that reduces the objective function. 

Key components:
- `GlassExpert`: the main optimizer class, subclassing `OptimizerGeneric`
- `get_neighbour_glasses`: finds the closest glasses in material space (n_d, V_d)
- `downsample_glass_map`: uses K-Means to reduce the search pool while preserving diversity
- `plot_glass_map`: for debugging and visualization of the current selection.

Please run  `Tutorial_7e_GlasseExpert.ipynb` to get started.

The code is modular, and I've added comments and docstrings to explain each part. 
It should integrate well with existing workflows and allow users to experiment with glass substitutions in a controlled way.

Looking forward to your feedback, and happy to adjust anything you find unclear or unpolished.

Thanks again for the great work on this project.


drpaprika:
### Example

We start with a voluntary bad 6 lenses objective, all N-BK7 lenses:
image

For each lens, we define a list of glasses corresponding to the glasses in both Schott and Ohara catalogs that are transmissive in the wavelength band. For example `glasses = ['N-BK7', 'S-BSM22', 'LF5G19', ...]`. The list of glasses can be different for each lens.

We then run `GlassExpert`, which makes two optimization passes:

During the first pass, we perform for each glass variable a broad search over `num_neighbours` glasses in the entire glass catalogue. An example with 6 is glasses:
	
image

During the second pass, we retreive for each glass variable the `num_neighbours` nearest materials in (n_d, V_d) space, and perform a focused local around the current glass choices. An example of local search around `FK3`:
	
image

At each step, for every candidate glass we:
	- Substitute it into the design
	- Run a continuous local optimization to evaluate performance
	- If the objective improves, keep the substitution; otherwise, roll back.

Finally, one more local-only optimization is performed over all continuous variables to polish the solution.

HarrisonKramer:
Wow, this is an incredible contribution! Really well thought out, clearly implemented, and significantly expands the optimization capabilities. The approach to discrete material optimization is well-structured, and it fits neatly into the existing optimization framework. The two-pass strategy, wide then local, is a very practical way to handle glass selection, and I like that it leaves room for user control.

The code is modular, readable, and the example notebook helps clarify how to use it. Everything looks good on my end and tests pass, so I’ll go ahead and merge. Thanks!

I'd also like to expand the documentation a bit, e.g., adding this to the learning guide, example gallery, and maybe some references through the readthedocs docs, but that's not urgent. I can help with that as a follow-up.

Looking forward to more like this!

Best,
Kramer

HarrisonKramer:
Documentation for Glass Expert expanded in #213.

PR #212, opened 2025-07-05
`total_track` operand

drpaprika:
Adds the `total_track` operand to `ParaxialOperand`. 
It is called using:

```python3
input_data = {'optic': lens}
problem.add_operand(
    operand_type='total_track',
    target=100,  # [mm]
    weight=1,
    input_data=input_data,
)
```

It appears that currently the total track is computed in `surface_group.py` using `z = self.positions[1:-1]`. The last surface to image distance is not accounted for, is it intended ?
I assumed it was not and corrected it, as it makes sense to me to include the sensor distance in the computation.
Therefore some tests fail, I'll update them or revert the total track computation according to your inputs.

Best,
drpaprika

### DIFF
--- a/tests/test_optic.py
+++ b/tests/test_optic.py
@@ -365,7 +365,7 @@ class TestOptic:
 
     def test_total_track_property(self, set_test_backend):
         lens = HeliarLens()
-        assert lens.total_track == 3.6291
+        assert lens.total_track == 12.1357
 
     def test_total_track_error(self, set_test_backend):
         lens = HeliarLens()


HarrisonKramer:
Hi @drpaprika,

Thanks for opening this PR. You are right - the total track is not defined correctly. This was an oversight on my part. This can be updated to include the image surface, and tests can be updated. Happy to help here if needed.

Best,
Kramer

drpaprika:
Thanks for your feedback I fixed the tests, it should be ready to merge

PR #234, opened 2025-07-24
diffraction grating

mattemilio:
Hi @HarrisonKramer ,
as discussed this is the first implementation of a diffraction grating such that you can check and comment the overall idea.
I get a good agreement with commercial software but I need to optimise/clean the code.
Kind regards 
Matteo


dyson_spectrometer.py
HarrisonKramer, line 1: Can you delete this? If you want, you can instead add this (or a related example) into a Jupyter notebook and add it into the example gallery. I can do that later too.

gratingexample.py
HarrisonKramer, line 1: Same as comment above. Delete this, or move it into the docs in a Jupyter notebook.

optiland/geometries/plane_grating.py
HarrisonKramer, line 1: In my view, geometries are only meant to define the shape and position of a surface, not how rays interact with it once they intercept it. I propose that we move any logic that is needed here into the new grating surface type you created. What do you think?

optiland/geometries/standard_grating.py
HarrisonKramer, line 1: Same remark as above. Let's try to move this into the surface class.

optiland/rays/real_rays.py
HarrisonKramer, line 1: The logic makes sense here, but I wonder where we should put this code. For now, we can keep it, but I am likely to refactor this very soon into a dedicated "interactions" module that controls reflection, refraction, diffraction, etc.

optiland/rays/real_rays.py
```diff
@@ -163,6 +163,84 @@ def reflect(self, nx, ny, nz):
         self.M = self.M - 2 * dot * ny
         self.N = self.N - 2 * dot * nz
 
+    def gratingdiffract_p(self, nx, ny, nz, fx, fy, fz, m, d, n1, n2, is_reflective):
```
HarrisonKramer: why are there two diffraction methods? Can we not just have one that is general for all diffraction behavior?

optiland/rays/real_rays.py
```diff
@@ -163,6 +163,84 @@ def reflect(self, nx, ny, nz):
...
... self.M0*d*n1*nx*ny - self.N0*d*n1*nx*nz + fx*m*ny**2*self.w + fx*m*nz**2*self.w - fy*m*nx*ny*self.w - fz*m*nx*nz*self.w + nx*be.sqrt(-self.L0**2*d**2*n1**2*ny**2 - self.L0**2*d**2*n1**2*nz**2 + 2*self.L0*self.M0*d**2*n1**2*nx*ny + 2*self.L0*self.N0*d**2*n1**2*nx*nz - 2*self.L0*d*fx*m*n1*ny**2*self.w - 2*self.L0*d*fx*m*n1*nz**2*self.w + 2*self.L0*d*fy*m*n1*nx*ny*self.w + 2*self.L0*d*fz*m*n1*nx*nz*self.w - self.M0**2*d**2*n1**2*nx**2 - self.M0**2*d**2*n1**2*nz**2 + 2*self.M0*self.N0*d**2*n1**2*ny*nz + 2*self.M0*d*fx*m*n1*nx*ny*self.w - 2*self.M0*d*fy*m*n1*nx**2*self.w - 2*self.M0*d*fy*m*n1*nz**2*self.w + 2*self.M0*d*fz*m*n1*ny*nz*self.w - self.N0**2*d**2*n1**2*nx**2 - self.N0**2*d**2*n1**2*ny**2 + 2*self.N0*d*fx*m*n1*nx*nz*self.w + 2*self.N0*d*fy*m*n1*ny*nz*self.w - 2*self.N0*d*fz*m*n1*nx**2*self.w - 2*self.N0*d*fz*m*n1*ny**2*self.w + d**2*n2**2*nx**2 + d**2*n2**2*ny**2 + d**2*n2**2*nz**2 - fx**2*m**2*ny**2*self.w**2 - fx**2*m**2*nz**2*self.w**2 + 2*fx*fy*m**2*nx*ny*self.w**2 + 2*fx*fz*m**2*nx*nz*self.w**2 - fy**2*m**2*nx**2*self.w**2 - fy**2*m**2*nz**2*self.w**2 + 2*fy*fz*m**2*ny*nz*self.w**2 - fz**2*m**2*nx**2*self.w**2 - fz**2*m**2*ny**2*self.w**2))/(d*n2*(nx**2 + ny**2 + nz**2))
```
HarrisonKramer: These lines are extremely long and hard to understand. Can we break them up so they're more readable? Or, at least add a comment about the logic. It will be next to impossible to debug these if we ever find a bug here.

optiland/surfaces/factories/geometry_factory.py
HarrisonKramer, line 1: As mentioned above, we should not modify geometries and all logic should go into the Surface class. So, this factory shouldn't be changed in the end.

optiland/surfaces/grating_surface.py
```diff
@@ -0,0 +1,354 @@
...
+from optiland.geometries import BaseGeometry
+from optiland.materials import BaseMaterial
+from optiland.physical_apertures import BaseAperture
+from optiland.physical_apertures.radial import configure_aperture
+from optiland.rays import BaseRays, ParaxialRays, RealRays
+from optiland.scatter import BaseBSDF
+from optiland.surfaces.standard_surface import Surface
+
+
+class GratingSurface(Surface):
```
HarrisonKramer, line 20: When subclassing `Surface`, you only need to rewrite methods that are different from the `Surface` class. So, many of these methods can be removed, as they're identical. Please remove those.

provagrating3.py
HarrisonKramer, line 1: Same as above - please remove, or if useful, make an example out of this and move into docs.

HarrisonKramer (changes requested):
Hi @mattemilio,

This is a good start! I think we need to rework the structure a bit, but the underlying algorithms make sense and should be able to stay the same. If you have some thoughts on the comments I added, please let me know.

If you like, you can make a few of the smaller changes I proposed first (e.g., deleting files), then we can work on the final changes together. That might be easier in the end. Just let me know.

Thanks again,
Kramer

optiland/geometries/standard_grating.py
mattemilio: I agree with the approach. The only problem I see is that I need to get a 'grating vector' that is a unit vector that lies in the tangent plane of the surface in the intersection point with the ray and also perpendicular to the local groove direction. So it is at the end a geometrical feature. I don't know how to get this from the surface class

optiland/rays/real_rays.py
mattemilio: One method is based on a simple equation that I found in the Zemax forum (https://community.zemax.com/got-a-question-7/how-diffraction-ray-tracing-is-calculated-112)
But with this equation I don' t get a full match with Zemax simulations.
The second method should represent an exact solution of the vector grating equation (e.g. from handbook of optics). I get indeed a very good match with Zemax simulations but I would like to simplify the equations.

optiland/rays/real_rays.py
mattemilio: I fully agree. This is work in progress.

optiland/geometries/standard_grating.py
HarrisonKramer: Good point. I will dive deeper into this to see what options we have. This problem indicates that we might already need to refactor some of the logic. I'll come back when I have something here.

optiland/surfaces/grating_surface.py
mattemilio: Understood. Thanks.

HarrisonKramer:
> In general, moving the logic from your geometry classes into the `GratingSurface` class seems to work without issue.

This statement I made is unfortunately not true. I oversimplified the grating calculations and the geometry indeed must be included. We can proceed with the proposed approach. We can refine the code and equations a bit, but I think the logic so far makes sense.

Kramer

HarrisonKramer (approved):

HarrisonKramer:
I think we just need to add some tests, perhaps add an example or two in the docs, then we can merge this.

It would be great if you could add some tests and/or examples in the docs (probably in the gallery). I can also help there. Just let me know.

HarrisonKramer:
Hi @mattemilio,

I made minor changes to the test to get them to pass, then added your spectrometer examples into the docs in a new section "Diffractive Systems".

Thanks again for developing this! Really great addition and we can build on it to enable design of other diffractive/phase systems.

Merging now!

Kramer

Issue #194, opened 2025-06-10
FFT PSF calculation for off-axis fields


crnh:
### Checklist

- [x] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [x] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [x] I am using the **latest version** of Optiland (if not, please update and retry).
- [x] I have tried to reproduce or debug the issue myself before opening this.
- [x] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

*Thanks for taking the time to go through this — it really helps us help you!*

### Bug Report

**Describe the bug**
The PSF output of Optiland is qualitatively different from OpticStudio when using a field with two nonzero coordinates. Everything works fine if only one coordinate is nonzero.

**To Reproduce**

Here is a minimal working example:

```python
import matplotlib.pyplot as plt
import numpy as np
import zospy as zp
from zospy.zpcore import OpticStudioSystem

from optiland.materials import IdealMaterial
from optiland.optic import Optic
from optiland.psf import FFTPSF

FIELDS = [
    (0, 0),
    (0, 10),
    (10, 0),
    (10, 10),
    (0, 30),
    (30, 0),
    (30, 30),
    (10, 20),
    (20, 10),
    (20, 20),
]


def effective_pupil_sampling(pupil_sampling: int) -> int:
    return np.floor(32 * np.sqrt(2) ** (np.log2(pupil_sampling) - 5)).astype(int)


def build_optiland_system(
    fields: list[tuple[float, float]],
    wavelength: float,
) -> Optic:
    lens = Optic()

    lens.set_aperture("EPD", 1.0)
    lens.set_field_type("angle")

    for field in fields:
        lens.add_field(x=field[0], y=field[1])

    lens.add_wavelength(wavelength)

    lens.add_surface(
        index=0,
        comment="Object",
        thickness=np.inf,
    )
    lens.add_surface(
        index=1,
        comment="Lens front",
        thickness=10,
        radius=10,
        is_stop=True,
        material=IdealMaterial(1.5),
    )
    lens.add_surface(
        index=2,
        comment="Lens back",
        thickness=11,
        radius=-10,
    )
    lens.add_surface(
        index=3,
        comment="Image",
    )

    return lens


def build_opticstudio_system(
[...]

PR #261, opened 2025-08-13
Generic Wavefront Error Calculation Strategy

HarrisonKramer:
# Add centroid-based wavefront error calculation strategy

This PR introduces a new and generic method for computing the wavefront error (WFE) of an optical system, improving flexibility and separation of concerns in the codebase.

Previously, the `Wavefront` class contained a single, baked-in calculation method based on the chief ray and paraxial exit pupil. For _exotic_ optical systems, such as asymmetric, off-axis, or obscured systems, it's not uncommon that ray trace failures occur for paraxial or real rays. This can lead to a practical difficulty in identifying the on-axis paraxial exit pupil location, or even the chief ray intersection in image space, which are required for the original WFE calculation method. This PR introduces a generic WFE calculation method that does not rely on a single chief ray trace, nor any paraxial calculations.

This update extracts the WFE calculation logic into separate **strategy classes**, making it easier to add new approaches in the future.

Two strategies are now available:
- **ChiefRayStrategy** (existing approach): Uses the chief ray intersection with the image plane as the reference sphere center. The center of the exit pupil defines the sphere radius.
- **CentroidSphereStrategy** (new): Uses the centroid of ray intersections on the image surface as the fixed sphere center, then fits the sphere radius to the OPD data.

The new centroid-based method is robust for:
- Off-axis systems
- Asymmetric pupils
- Systems with obscurations
- Arbitrary optical configurations (no reliance on paraxial tracing)

This PR also adds an option in the wavefront to remove tilt, which helps in centering PSFs for off-axis fields.

These changes also enable all PSF and MTF calculations to accept a `strategy` argument for consistent and customizable wavefront error evaluation.

Example for a TMA:
```python
# Use strategy "chief_ray" (default) or "centroid_sphere" (new)
opd = OPD(lens, field=(0.24, -0.56), wavelength=0.55, strategy="centroid_sphere")
_ = opd.view()
```
image


HarrisonKramer:
Hi @crnh,

As mentioned in #194, this PR includes an update to the `Wavefront` class to optionally remove tilt. I propagated this argument to PSF and MTF classes. Currently, I do not remove tilt by default. However, if you agree, I will change this to remove tilt from all wavefronts by default prior to any PSF or MTF calculation. I think this is the right approach to match ZOS, but I am open to discussion.

I will already merge this, but let me know and I can make a quick fix in master later on.

Kramer

PR #262, opened 2025-08-14
Add type hints to `Optic`

crnh:
Large parts of Optiland's code base are currently untyped. Adding type hints improves the maintainability of the project, helps to spot errors early, and strongly improves the developer experience.

I added type annotations to the `Optic` class, because this is a core part of the public API. Types that are / can be used in more places are defined in `optiland._types`. Literals are used for string arguments, so intellisense now shows suggestions for parameters like `aperture_type`, `surface_type` etc.:

image

A `TypedDict` is used to specify the parameters for `**kwargs` in `Optic.add_surface`. The `TypedDict` currently merges all possible geometries, just like `GeometryConfig`. However, using [overloads](https://typing.python.org/en/latest/spec/overload.html) it is possible to separately specify these parameters for each geometry, which would let the type checker check if all specified parameters are compatible with the surface type. If we want to do this, it's probably best to do it in a separate PR.


crnh:
@HarrisonKramer I merged the PR, but it would be good to still discuss this question from the PR description:

> However, using [overloads](https://typing.python.org/en/latest/spec/overload.html) it is possible to separately specify these parameters for each geometry, which would let the type checker check if all specified parameters are compatible with the surface type. If we want to do this, it's probably best to do it in a separate PR.

Do you think what we have now is good enough, or do you prefer the solution described above? If you want, I can make a draft so you can see what it looks like.

crnh:
> Is it a lot of work to implement this?

It requires defining a separate signature for each geometry, like this:

```python
@overload
def add_surface(
    new_surface: Surface | None = None,
    surface_type: SurfaceType = "zernike",
    comment: str = "",
    index: int | None = None,
    is_stop: bool = False,
    material: str | BaseMaterial = "air",
    **kwargs: Unpack[ZernikeSurfaceParameters],
```

Where `ZernikeSurfaceParameters` is a separate `TypedDict` in `optiland._types` that specifies the parameters specific to the Zernike geometry. Using overload thus also requires splitting `SurfaceParameters` in a generic part and surface-specific `TypedDict`s that extend the generic type.

> In my view, the only argument to be made against using overloads would be the (slightly higher) complexity

Yes, it indeed adds an extra step to adding new geometries.

> or possibly the risk that the API changes. However, I have no expectation the geometry API will change anytime soon.

If many more geometries are added, the `GeometryConfig` class could become very large and be split in surface-specific classes; I wouldn't be surprised if that would happen somewhere in the future. However, a change like that may also require changes in the current typing implementation.

In any case, it is probably better to first add type annotations for the remaining parts of the code base and see if we can implement type checking. That will also take some time (and probably reveal some bugs).

PR #279, opened 2025-08-20
Material class fix and visualization

lordpositron:
This PR fixes an issue in the `CSV` catalog generation where some list elements were missing the required key. This caused an exception and prevented entries from being created for certain materials.
This is now resolved, and no entries are deleted.
Additionally:

- Restored the generation of the `Filename_no_ext` column, which was missing but is required for material lookup to work properly.
- Added a plotting utility function to visualize (n, k) vs wavelength for materials. Example :
```
from optiland.materials import Material, plot_nk
mat = Material("Cu")
plot_nk(mat, wavelength_range=(0.1, 2.5))
```
output


HarrisonKramer:
Hi @lordpositron,

Thanks for adding this! and for fixing the issues with `Material` instances. I really like the n-k plotting utility - it's a simple but very useful diagnostic tool that will be helpful for material analysis.

Merging now.

Kramer

Issue #254, opened 2025-08-12
Optimizer and general constraints


Jonas231:
### Checklist

- [ x] have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or feature request.

- [ x] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [ x] I am using the **latest version** of Optiland.
- [ x] I have included **all necessary context**.

*Thanks for taking the time to go through this — it really helps us help you!*

### Feature Request

**Is your feature request related to a problem? Please describe.**
I think what is missing is a Optiland optimizer contraints framework including general contraints like "MNEA" ... min. edge air or "MNEG" ... min. Edge glass, MXCG ... max. center glass and either constraint them to an exact value either with Lagrange multiplier or high weighting or a range of values greater or smaller than a value. Perhaps it is necessary to allow for user-defined contraints within an interface with sth. like macros and @a == (edg s2) evaluates the edge thickness and allows to use it within a constraint like "@a > 2". 

**Describe the solution you'd like**
Best would be operands similar to mnea etc...

**Describe alternatives you've considered**
Some routine which evaluates if lenses can be fabricated based on a list of crriteria. If they cannot be fabricated this is indicated. Then the user can reoptimize in a loop until e.g. the lens edge is thick enough. 

**Additional context**
This is essential for finding manufacturable designs for lenses with higher complexity.

manuelFragata:
Implemented in PR #290 - marking the issue as close for now. Please feel free to re-open if necessary.
Best,
Manuel

PR #290, opened 2025-08-24
feat: add new logic for edge_thickness operand + tests (#254)

manuelFragata:
- as per issue #254 , the logic for calculating the edge_thickness operand is completed. 
- added several tests to compare the results with expected output. Should give max coverage.


HarrisonKramer (approved):

HarrisonKramer:
Looks good, thanks for adding this! You can merge when you're ready.

Kramer

PR #308, opened 2025-09-05
Feature: MMDFT PSF, for accurate PSF calculation regardless of wavelength

scottpaine:
**THIS PULL REQUEST IS NOT COMPLETE YET. WARNING: WALL OF TEXT BELOW**

I'm presenting code that I wrote to compute PSFs using the matrix-multiply discrete Fourier transform method. This method requires no padding, and computes the FT at specific output points selected by the user. This means the user can supply the pixel pitch directly. 

In an FFT-based PSF, the output sample spacing is determined by the pad size of the input as well as the product of the F-number and the wavelength. If the user desires a specific output sample spacing, chances are there is not an integer pad size that corresponds to this spacing. This means simulation with a specific pixel pitch using an FFT requires interpolation. Similarly, the output sample spacing is dependent on wavelength, and so a polychromatic PSF requires different pad sizes and interpolation for the spacing between pixels to be consistent with wavelength.

The MMDFT directly computes the Fourier Kernel based on the desired sample spacing, and therefore has no limitations brought on by pad size. You could think of the MMDFT as allowing non-integer pad sizes. The MMDFT's complexity is (I think) O(n^2), which means it scales worse than the FFT which is O(n log(n)), but for large pad sizes (which comes about with a desire for small pixel pitch), the MMDFT is faster since there are fewer pixels to push through the FT.

Before the pull request is in, I do want some input from the team, based mostly around polychromatic PSF. Here are my big questions that need answered:

- Polychromatic PSFs are presently generated assuming all wavelengths have equal intensity. Do we want to leave it this way for now, or invent some way to add in wavelength weightings?
- Normalization now is done by generating the unaberrated PSF of the primary wavelength and considering its peak value. Right now, if you make a polychromatic PSF, you can have a Strehl ratio greater than 1, since you add all of the monochromatic PSFs together and then normalize just based on one PSF. The alternatives are:
    - Compute an unaberrated polychromatic PSF and use its peak value as the normalization factor (I would change this in FFTPSF too, for now HuygensPSF looks to be monochromatic)
    - Use energy conservation to say that PSF(0,0) = sum[wl_weight * sum(ideal_amplitude^2)]. This is taking advantage of the fact that when (x,y) = (0,0) in the image plane, the Fourier calculation is just sum(pupil_field) since the kernel is 1. This is much faster since it requires only summation, but falls apart if your calculated PSF does not include the point (0,0).
 - FFTPSF does not presently do padding correctly for polychromatic PSF evaluation. Right now it is just set to pad to `grid_size`, which means the spacing in the image plane is not consistent across all wavelengths (dx = wavelength * FNO / Q). The MMDFT does this appropriately with a different pad size for each wavelength. How would the team like to fix FFTPSF for this?

Things that I still have to do:
 - [ ] Lint the code with ruff
 - [ ] Documentation everywhere
 - [ ] Make everything more readable and self-documenting
 - [ ] Implement team-approved normalization and Strehl calculations
 - [ ] Implement any team-approved changes to other PSF files

### DIFF
--- a/optiland/psf/mmdft.py
+++ b/optiland/psf/mmdft.py
@@ -56,12 +56,13 @@ class MMDFTPSF(BasePSF):
         remove_tilt=False,
         **kwargs,
     ):
-        if image_size is None:
+        if image_size is None and pixel_pitch is None:
             if num_rays < 32:
                 raise ValueError(
-                    "num_rays must be at least 32 if grid_size is not specified."
+                    "num_rays must be at least 32 if image_size and pixel_pitch are "
+                    "not specified."
                 )
-            num_rays, image_size = calculate_grid_size(num_rays)
+            num_rays, grid_size = calculate_grid_size(num_rays)
 
         super().__init__(
             optic=optic,
@@ -75,10 +76,30 @@ class MMDFTPSF(BasePSF):
 
         clear_size = num_rays - 1
 
+        # General idea for below:
+        #  - If only pixel_pitch is None, calculate based off of image_size, assuming
+        #    image_size is full pad size
+        #  - If only image_size is None, calculate based off of pixel_pitch,
+        #    truncating any trailing decimals
+        #  - If both pixel_pitch and image_size are none, set image_size = grid_size
+        #    and solve for pixel_pitch
+        # I would implement this as a function, but since it needs access to the FNO,
+        # it has to happen after super() initialization, which requires knowing
+        # num_rays ahead of time
         if pixel_pitch is None:
-            # Set pad_size = image_size, solve for pixel_pitch
+            if image_size is None:
+                # Use grid_size above to set Q and therefore pixel pitch
+                image_size = grid_size
             pixel_pitch = (
                     wavelength * self._get_working_FNO() * clear_size / image_size
+            )
+
+        # Below triggers only if pixel_pitch was given but image_size was not
+        if image_size is None:
+            # Use pixel_pitch to calculate max pad size and set image size to be 1
+            # pixel less than that
+            image_size = int(
+                wavelength * self._get_working_FNO() * clear_size / pixel_pitch
             )
 
         self.image_size = image_size
@@ -212,8 +233,8 @@ class MMDFTPSF(BasePSF):
         """
         # Assume clear aperture is defined as (num_rays - 1) - done in FFTPSF
         clear_size = self.num_rays - 1
-        Q = self.wavelengths[0] * self._get_working_FNO() / self.pixel_pitch
-        pad_size = Q * clear_size
+        pad_size = (self.wavelengths[0] * self._get_working_FNO() * clear_size /
+                    self.pixel_pitch)
 
         # Check to make sure we aren't sampling outside the max extent of the image
         # domain (defined by pad_size)
--- a/optiland/psf/mmdft.py
+++ b/optiland/psf/mmdft.py
@@ -1,3 +1,14 @@
+"""MMDFT Point Spread Function (PSF) Module
+
+This module provides functionality for simulating and analyzing the Point
+Spread Function (PSF) of optical systems using the Matrix Multiply Discrete Fourier
+Transform. It includes capabilities for generating a monochromatic PSF at a desired
+image size and pixel pitch from given wavefront aberrations and calculating the
+Strehl ratio. Visualization is handled by the base class.
+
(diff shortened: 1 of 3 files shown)


optiland/psf/mmdft.py
```diff
@@ -0,0 +1,168 @@
...
+            remove_tilt=remove_tilt,
+            **kwargs,
+        )
+
+        if pixel_pitch is None:
+            pixel_pitch = be.min(self.wavelengths) * self._get_working_FNO() / 2
+
+        self.grid_size = grid_size
+        self.pixel_pitch = pixel_pitch
+        self.pupils = self._generate_pupils()
```
HarrisonKramer: Pupils and PSFs are already generated via the `super` call, so these are redundant and can be removed.

*This comment may be obsolete if you rework to use `BasePSF`

optiland/psf/mmdft.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # TODO: Strehl Ratio Calculations (ties in with normalization)
+
+    def _compute_base_kernels(self):
+        # TODO: Documentation
+        # clear_size is a reference to the number of pixels in the pupil grid
+        # covering the pupil. The assumption here is that num_rays spans the X and Y
+        # extend of the pupil with no pixels inside having an amplitude of zero.
+        clear_size = self.num_rays
+        Qs = self.wavelengths * self._get_working_FNO() / self.pixel_pitch
```
HarrisonKramer: `self.wavelengths` is a list, so this calculation fails. The fact that it is a list stems from how the wavelengths are passed to the `Wavefront` class, which is responsible for generating the wavefront error in the exit pupil. Currently, all PSF implementations only accept a scalar wavelength, so `self.wavelengths` will also be a list with one element. This means you should use `self.wavelengths[0]`.

Yes, this is messy and we should probably clean it up, but I propose we wait till we handle the polychromatic PSFs.

*This comment may be obsolete if you rework to use `BasePSF`

optiland/psf/mmdft.py
```diff
@@ -0,0 +1,168 @@
+from __future__ import annotations
+
+import numpy as np
+
+import optiland.backend as be
+from optiland.psf.fft import FFTPSF, calculate_grid_size
+
+class MMDFTPSF(FFTPSF):
```
HarrisonKramer: I would recommend we instead inherit from `BasePSF` here. To me, if the new class is not a variant on, or a "kind of" the super class, then it should not be used. Can we try to rework it with `BasePSF` instead?

I also think that will solve some other issues I see in `__init__` below.

optiland/psf/mmdft.py
HarrisonKramer: `self.pixel_pitch` will not have been defined at this point --> a reordering in the constructor would be needed

HarrisonKramer (changes requested):
The algorithm looks good in general, but I propose we revise the implementation. As mentioned in comments, I propose we subclass `BasePSF` here. That would simplify the constructor logic as well. I am open to suggestions here though if you prefer it a different approach.

my first thought on the constructor was that we could, in order:
1. handle grid size calculation via `calculate_grid_size`
2. call `BasePSF` constructor
3. calculate pixel pitch (what you already have)
4. generate pupil and PSF

(I did not test this).

I will answer your other questions in another comment. These are just my initial thoughts on the implementation. Let me know what you think. Happy to discuss or be shown a better way.

scottpaine:
@HarrisonKramer I think this is all very fair. For now I'll change the implementation in two ways:

 - Subclass `BasePSF` (I was really just stealing the pupil generation code from FFTPSF without thinking)
 - Treat the PSF as monochromatic
     - I'll let the user supply the `pixel_pitch` and `image_size` like in `HuygensPSF`

HarrisonKramer:
Thanks for continuing here, and for identifying existing issues / proposing improvements.

By the way, just invited you to be a collaborator in the repo, which gives you more rights here. You can also keep your branches here in the main repo, if you prefer.

We can create issues to track the new features / improvements. We can tie those to the upcoming PRs.

On testing - you can likely adapt the existing `test_fft_psf.py` into a new `test_mmdft_psf.py` file, making whichever changes needed to fully cover all use cases. LLMs or AI coding agents (copilot, gemini, jules, etc.) usually get you pretty far here if you want to avoid some of the "boring" stuff.

Kramer

PR #324, opened 2025-09-12
Add type hints to wavefront module

crnh:
Improve typing for the wavefront module.

I noticed that `OPDFan.view` calls `add_subplots` on a figure:

https://github.com/HarrisonKramer/optiland/blob/b3f89fa660be6ac3da2fea2d2b9d532a2a8e10c9/optiland/wavefront/opd_fan.py#L103

Shouldn't this be `add_subplot`? `add_subplots` doesn't seem to exist.

### DIFF
--- a/optiland/wavefront/opd_fan.py
+++ b/optiland/wavefront/opd_fan.py
@@ -100,7 +100,7 @@ class OPDFan(Wavefront):
         if is_gui_embedding:
             current_fig = cast("Figure", fig_to_plot_on)
             current_fig.clear()
-            axs = current_fig.add_subplots(
+            axs = current_fig.add_subplot(
                 nrows=len(self.fields),
                 ncols=2,
                 figsize=(figsize[0], num_rows * figsize[1]),


optiland/wavefront/strategy.py
```diff
@@ -11,12 +11,23 @@
...
 from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING, Literal
 
 import optiland.backend as be
 
 from ..fields.field_types import AngleField
 from .wavefront_data import WavefrontData
 
+if TYPE_CHECKING:
+    from numpy.typing import NDArray
```
HarrisonKramer, line 22: Should this not be backend-agnostic? We have `be.ndarray` defined in the backend, but perhaps we need a new type definition that can be either a numpy array or torch tensor.

optiland/wavefront/strategy.py
```diff
@@ -95,7 +116,13 @@ def _opd_image_to_xp(self, rays_at_image, xc, yc, zc, R, wavelength):
 
         return self.n_image * t
 
-    def _correct_tilt(self, field, opd, x=None, y=None):
+    def _correct_tilt(
+        self,
+        field: tuple[float, float],
+        opd: NDArray,
+        x: NDArray | float | None = None,
```
HarrisonKramer, line 123: see comment on backend-agnostic array-like types

optiland/wavefront/strategy.py
crnh: `be.ndarray` is a function, not a type. You cannot use it for type hints. Python won't raise an exception, but type checkers will show an error.

HarrisonKramer: I guess `be.ndarray` is just a variable, which will be a tuple containing `(numpy.ndarray, torch.Tensor)` if both numpy and torch are available.

In your opinion, does it make sense to define something like the following in our `_types.py` file?

```python
ArrayLike = Union[np.ndarray, torch.Tensor]
```

That could then be used throughout the package for type hinting. Is that a better option?

crnh: > I guess be.ndarray is just a variable

PyRight complains about `be.ndarray` being a function, so at least it's not valid as a type hint. It is also defined during runtime, so not suitable for static analysis.

Defining a custom union variable would indeed be an idea, but I'd rather not call it `ArrayLike`. `ArrayLike` already is a type in NumPy.

HarrisonKramer: Okay, clear. As you proposed, let's hold off on updating it for another PR. We can agree later on a suitable name that is logical for both backends.

crnh: I tried some ideas and this could work:

```python
from typing import TYPE_CHECKING, Union
from numpy.typing import NDArray

...

if TYPE_CHECKING:
    from torch import Tensor  # noqa: F401

...

BEArray = Union[NDArray, "Tensor"]
```

The `noqa`, and explicit `Union` instead of `|`, and quoted `"Tensor"` are necessary because `torch` may be unavailable; using `Tensor` without quotes would cause errors if `torch` is not installed.
I hope I can find some time to draft a proposal soon.

HarrisonKramer: Perfect. Looks good to me! Also, `BEArray` makes sense to me as well.

PR #332, opened 2025-09-23
Diffractive 

Hhsoj:
I added phase to a surface and included a grating and a radial phase function. I did some testing and it seems correct to me compared the results to known correct ray traces.  I still need to add Diffraction efficiency.  current it assumes Ideal, so no loss.


HarrisonKramer:
Hi @Hhsoj,

Thanks for submitting this! And sorry for the delay in responding.

I see there are some minor merge conflicts. I can work to resolve these. Once that's done, we'll also want to add some unit tests. That should be relatively straightforward once the core functionality is worked out.

Regards,
Kramer

PR #356, opened 2025-10-05
Allow setting `primary_index` in `WavelengthGroup`

rjmoerland:
This small PR adds a setter function for `WavelengthGroup.primary_index`, for convenience. I included Pytests, but unfortunately I could not test the whole suite because of some issues with the PyTorch backend on my machine (MacOS, if that matters - I've never used PyTorch before). Without the PyTorch backend, about half of the tests fail (makes sense), but with the backend installed, the Python interpreter crashes.

### DIFF
--- a/optiland/wavelength.py
+++ b/optiland/wavelength.py
@@ -154,6 +154,14 @@ class WavelengthGroup:
             if wavelength.is_primary:
                 return index
 
+    @primary_index.setter
+    def primary_index(self, index: int):
+        """set the wavelength indexed by `index` as primary"""
+        if not 0 <= index < len(self.wavelengths):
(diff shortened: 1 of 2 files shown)


HarrisonKramer:
Hi @rjmoerland,

Thanks for submitting this! Looks solid. I just approved the workflows, so tests are starting now. We can merge when all passing.

While I've never tried it, I think you can still install and use torch on CPU on MacOS. Perhaps manually setting the device to CPU fixes the issue? Not sure here, as I don't use MacOS.

Best,
Kramer

HarrisonKramer:
merged! and add you as an author in the docs: https://optiland.readthedocs.io/en/latest/authors.html

Kramer

Issue #368, opened 2025-10-12
Repeated use of `Optic.add_surface(index=1)` gives incorrect raytracing result


rjmoerland:
### Checklist

- [x] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [x] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [x] I am using the **latest version** of Optiland (if not, please update and retry).
- [x] I have tried to reproduce or debug the issue myself before opening this.
- [x] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

### Bug Report

**Describe the bug**
The insertion mechanism of `Optic.add_surface()` does not give correct results for traced (real) rays. This is best illustrated by one of the Pytests [here](https://github.com/HarrisonKramer/optiland/blob/d96bbb1dba4c0036520bd9ca8a6336f4d4290db4/tests/test_surface_group.py#L517). It defines a Cooke triplet by repeatedly inserting a new surface at index 1. The resulting `Optic` instance gives the wrong results. Interestingly, the pytest tests for consistency, but the number it tests against seems incorrect. The triplet has a focal length of nearly 50 mm, and the field is of the angular type at (max) 20 degrees. Therefore, the approximate mean intersection with the y-axis should be at tan(20°) * 50 mm ~ 18.2 mm. The number tested against is 3.47 mm, which is not what I expect.

**To Reproduce**
Run the below code:
```
import numpy as np
from optiland.optic import Optic
from optiland.samples import CookeTriplet

cooke = CookeTriplet()
cooke.draw()

lens = Optic()

lens.add_surface(index=0, radius=np.inf, thickness=np.inf)
lens.add_surface(index=1)

for surf in cooke.surface_group.surfaces[-2:0:-1]:
    lens.add_surface(index=1, radius=surf.geometry.radius, material=surf.material_post, is_stop=surf.is_stop, thickness=surf.thickness)

lens.set_aperture(aperture_type="EPD", value=10)
lens.set_field_type(field_type="angle")
lens.add_field(y=0)
lens.add_field(y=14)
lens.add_field(y=20)

lens.add_wavelength(value=0.48)
lens.add_wavelength(value=0.55, is_primary=True)
lens.add_wavelength(value=0.65)
lens.draw()
```

Output:

#### Cooke:

[...]

Issue #364, opened 2025-10-09
`Optic.set_material` does not update interaction model


rjmoerland:
### Checklist

- [x] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [x] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [x] I am using the **latest version** of Optiland (if not, please update and retry).
- [x] I have tried to reproduce or debug the issue myself before opening this.
- [x] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

### Bug Report

Using `Optic.set_material` to change the material of a surface does not update the interaction model, and therefore raytracing gives the wrong results.

**To Reproduce**
The following code sets up two slabs, the first with `n = 2.5`, the second with `n = 1.0001`. Refraction should take place at the front and back of the first slab.

```
import numpy as np
from optiland.optic import Optic
from optiland.materials import IdealMaterial

lens = Optic()
lens.add_surface(index=0, thickness=np.inf, material="Air")
lens.add_surface(index=1, surface_type="standard", material=IdealMaterial(n=2.5), thickness=10, radius=np.inf, is_stop=False)
lens.add_surface(index=2, surface_type="standard", material=IdealMaterial(n=1.0001), thickness=10, radius=np.inf, is_stop=True)
lens.add_surface(index=3, surface_type="standard", material="Air", thickness=20, radius=np.inf)
lens.add_surface(index=4, material="Air")
lens.set_field_type("angle")
lens.add_field(y=10)
lens.add_wavelength(500, unit="nm")
lens.set_aperture("float_by_stop_size", 25)
lens.draw()
```
(Correct) output:

<img width="641" height="371" alt="Image" src="https://github.com/user-attachments/assets/bb94d8d8-fde1-47df-be1c-9830a8303e04" />

Now set the material of the first slab to air, and draw again:
```
lens.set_material(IdealMaterial(1.0), 1)
lens.draw()
```
Incorrect result (refraction at air/n=1.0001 surfaces):

<img width="644" height="371" alt="Image" src="https://github.com/user-attachments/assets/96eb41ce-8137-46b4-ac9f-390bec9fe581" />

**Expected behavior**
No refraction at all (or virtually no refraction, given that the second slab has a 0.0001 larger refractive index):

[...]

PR #369, opened 2025-10-12
Avoid multiple copies of material objects

### LINKED ISSUE #362: Flipping with Optic.flip() results in thickness out of sync with z position
### Checklist

### Bug Report

When I use `Optic.flip()` to flip around a compound lens, I have noticed that the thickness associated with a surface and the position of the next surface do not match.

**To Reproduce**

Run the following code:
```
import numpy as np

lens = Optic()
lens.add_surface(index=0, thickness=np.inf, material="Air")
lens.add_surface(index=1, surface_type="standard", material="N-BK7", thickness=7, radius=45, is_stop=True)
lens.add_surface(index=2, surface_type="standard", material="SF5", thickness=2.5, radius=-34, is_stop=False)
lens.add_surface(index=3, surface_type="standard", material="Air", thickness=70, radius=-95.5, is_stop=False)
lens.add_surface(index=4, material="Air")
lens.set_field_type("angle")
lens.add_field(y=0)
lens.add_wavelength(500, unit="nm")
lens.set_aperture("float_by_stop_size", 25)

for idx, surf in enumerate(lens.surface_group.surfaces):
    print(f"{idx}: geom.z: {surf.geometry.cs.z}", f"Thickness: {surf.thickness}")
```

Output:
```
0: geom.z: -inf Thickness: inf
1: geom.z: 0.0 Thickness: 7
2: geom.z: 7.0 Thickness: 2.5
3: geom.z: 9.5 Thickness: 70
4: geom.z: 79.5 Thickness: 0.0
```

This is according to the definitions. Now:
```
lens.flip()
for idx, surf in enumerate(lens.surface_group.surfaces):
    print(f"{idx}: geom.z: {surf.geometry.cs.z}", f"Thickness: {surf.thickness}")
```
Output:
```
0: geom.z: -inf Thickness: inf
1: geom.z: 0.0 Thickness: 70
2: geom.z: 2.5 Thickness: 2.5
3: geom.z: 9.5 Thickness: 7
4: geom.z: 79.5 Thickness: 0.0
```

**Expected behavior**

I expected the following output (if starting from the last surface (== Image plane)):
```
0: geom.z: -inf Thickness: inf
1: geom.z: 0.0 Thickness: 70
2: geom.z: 70.0 Thickness: 2.5
3: geom.z: 72.5 Thickness: 7
4: geom.z: 79.5 Thickness: 0.0
```

Also acceptable, starting from the last "physical" surface:
```
0: geom.z: -inf Thickness: inf
1: geom.z: 0.0 Thickness: 2.5
2: geom.z: 2.5 Thickness: 7
3: geom.z: 9.5 Thickness: 70
4: geom.z: 79.5 Thickness: 0.0
```

**Environment**  
- Optiland Version: `master:7a2feca`
- Python Version: `3.13`
- OS: `macOS`
- Additional Info:

**Additional context**

It seems like the thicknesses aren't updated after the flip. What _seems_ to fix this is the following patch:
```
diff --git forkSrcPrefix/optiland/surfaces/surface_group.py forkDstPrefix/optiland/surfaces/surface_group.py
index d078881325bcdf0184ee99ca94f7cffd93142c75..4e58f70ff3f156fc12debb7da76272a77295a383 100644
--- forkSrcPrefix/optiland/surfaces/surface_group.py
+++ forkDstPrefix/optiland/surfaces/surface_group.py
@@ -486,5 +486,6 @@ class SurfaceGroup:
                 next_surf_in_new_order.geometry.cs.z = (
                     current_surf_in_new_order.geometry.cs.z + thickness
                 )
+                current_surf_in_new_order.thickness = thickness
 
         self.reset()

```

Note that `lens.info()` kind of swept this under the rug, since it doesn't actually use `Surface.thickness()` but calculates thicknesses based on the diff of the z positions. This makes me wonder if it isn't more robust to do this everywhere, instead of having two enitites that describe intimately related items?

### LINKED ISSUE #363: Removing a surface does not update `material_pre` for next surface
### Checklist

### Bug Report

Removing a surface leads to errors in ray tracing, as the `material_post` is not updated for the surface behind the removed surface. 

**To Reproduce**

See the model below, of two slabs with different refractive index. The first slab has `n = 2.5`, the second one has `n = 1.0001` (not 1.0 in order to have it show up). Refraction should take place at the front and back surface of the first slab, whereas virtually no refraction should take place at the back of the second slab:

```
import numpy as np

lens = Optic()
lens.add_surface(index=0, thickness=np.inf, material="Air")
lens.add_surface(index=1, surface_type="standard", material=IdealMaterial(n=2.5), thickness=10, radius=np.inf, is_stop=False)
lens.add_surface(index=2, surface_type="standard", material=IdealMaterial(n=1.0001), thickness=10, radius=np.inf, is_stop=True)
lens.add_surface(index=3, surface_type="standard", material="Air", thickness=20, radius=np.inf)
lens.add_surface(index=4, material="Air")
lens.set_field_type("angle")
lens.add_field(y=10)
lens.add_wavelength(500, unit="nm")
lens.set_aperture("float_by_stop_size", 25)
lens.draw()
```

(Correct) output:

Image

Remove the first physical surface (index 1). There should be virtually no refraction, given that the refractive indices are so close to each other (n = 1.0 vs n = 1.0001):
```
lens.surface_group.remove_surface(1)
lens.draw()
```

(Incorrect) output:

Image

Note that the output of `lens.info()` is correct, but not the rays that were traced:
```
╒════╤═══════════════╤═══════════╤══════════╤═════════════╤════════════╤═════════╤═════════════════╕
│    │ Type          │ Comment   │   Radius │   Thickness │ Material   │   Conic │   Semi-aperture │
╞════╪═══════════════╪═══════════╪══════════╪═════════════╪════════════╪═════════╪═════════════════╡
│  0 │ Planar        │           │      inf │         inf │ Air        │       0 │         12.5    │
│  1 │ Stop - Planar │           │      inf │          10 │ 1.0001     │       0 │         12.5    │
│  2 │ Planar        │           │      inf │          20 │ Air        │       0 │         14.2631 │
│  3 │ Planar        │           │      inf │         nan │ Air        │       0 │         17.7896 │
╘════╧═══════════════╧═══════════╧══════════╧═════════════╧════════════╧═════════╧═════════════════╛
```

**Expected behavior**

Image

**Environment**  
- Optiland Version: `X.X.X`
- Python Version: `X.X`
- OS: `[Windows/Linux/macOS]`
- Additional Info:

I think that `material_pre` needs to be fixed here: https://github.com/HarrisonKramer/optiland/blob/7a2feca6e061328e507b0692ce2e9a27f62a1165/optiland/surfaces/surface_group.py#L314

rjmoerland:
## Why

The current architecture requires diligent bookkeeping of materials (pre and post) at various locations scattered around various objects. Failure to update the material at one of those locations leads to erroneous results. The issues #363 , #364 and #368 are probably a direct result of missing a location to update.

This (draft) PR is for an initial implementation of an architectural change (proof of principle) that has the following benefits:
- avoids needing to set `material_post` in both a `Surface` and its `BaseInteraction` (or derivative) member variables
- avoids needing `material_pre` in `Surface`, and avoids having to set it in each `BaseInteraction` object
- avoids needing to set  `geometry` in `Surface` and the `BaseInteraction` member variables of `Surface`

## Implementation
In the implementation, I chose to change two main things:
1. Pass the preceding surface to the surface being initialized, and update the links when the order of the surfaces changes This way, `material_pre` in a surface simply becomes a link to `Surface.previous_surface.material_post`. 
2. Pass the parent surface to a child `BaseInteraction` object. This way, the `BaseInteraction` object (or derivative) has direct access to the materials and geometry as set in the `Surface` object.

To protect the links from being broken, the `SurfaceGroup.surfaces` member variable is now a read-only property, which returns a tuple containing all surfaces. Direct modification should be avoided (use `add_surface` and `remove_surface`), but is possible by modifying the underlying `_surfaces` list. 

 To keep the code as much in tact as possible (avoiding too much refactoring for a proof of principle), I added a property `Surface.material_pre` that simply returns the preceding surface's `material_post` value, if it exists. If the preceding surface is `None` (`ObjectSurface`), it returns `material_post` of the current surface, in line with the current behavior. 

The `BaseInteraction` class was updated with `material_pre`, `material_post` and `geometry` properties that now refer to the parent `Surface` instance and get their data from it. 

I finally updated the pytests, and they all pass on my Windows and Mac machines.

This implementation fixes #363 , #364 and #368 by avoiding the extra overhead that comes with the current implementation. It does not fix #362 . I know a fix for this has landed, but I think avoiding two things to maintain in an object that are derivatives of one another should be avoided. But first let's focus on this :)

As a consequence, the file format of an Optiland object will change. I have not looked at backward compatibility yet, but file IO should work as-is according to the passing pytests. I made an effort to make it possible to load older files that contain member variables that since have become obsolete. They are ignored.

### DIFF
--- a/optiland/surfaces/surface_group.py
+++ b/optiland/surfaces/surface_group.py
@@ -402,19 +402,12 @@ class SurfaceGroup:
         if not self._surfaces:
             return
 
-        effective_start_index = start_index
-        if start_index == 0:
-            effective_start_index = 1
-
-        if effective_start_index >= len(self._surfaces):
-            return
+        effective_start_index = max(start_index, 1)  # No update to object surface
 
         for i in range(effective_start_index, len(self._surfaces)):
             current_surface = self._surfaces[i]
 
-            if i == 0:  # no update to object surface
-                continue
-            elif i == 1:  # first surface lies at z=0.0 by definition
+            if i == 1:  # first surface lies at z=0.0 by definition
                 new_z = 0.0
             else:
                 prev_surface = self._surfaces[i - 1]
--- a/tests/test_surface_group.py
+++ b/tests/test_surface_group.py
@@ -80,7 +80,7 @@ class TestSurfaceGroupUpdatesRealObjects:
         sg._surfaces = initial_surfaces
         sg._update_surface_links()
 
-        if not use_absolute_cs and len(sg.surfaces) > 1:
+        if not use_absolute_cs:
             sg._update_coordinate_systems(start_index=0)
 
         return sg


rjmoerland:
Thank you! I have one thing, that has me a little worried: you mentioned in response to my questions that a negative image surface was by design. This PR disables that (point "`Optic.image_solve()` moves the image surface to negative z-positions for virtual images. I stopped it from doing that [...]") This might need to be reverted if it was by design?

HarrisonKramer:
On the negative image thickness, I hadn't thought too deeply about this to be honest. The negative thickness simply comes from the fact that the marginal ray has a positive slope in image space, so the intersection point with the optical axis (definitionally the image point) is "behind" the last surface.

I don't consider this a big problem, but if we want to keep things completely generic, then yes, we should revert it. However, it could be made clearer in the docs (or perhaps even in how the visualization is performed) that this is a virtual image.

Sorry that I missed that in my original review. Do you mind quickly reverting that? I can also do this.

Thanks!

PR #382, opened 2025-10-26
Paraxial focal length: fix wrong sign for negative lenses

rjmoerland:
# Why

The paraxial back focal length `f2` of a negative lens is returned as a positive number by Optiland:

```
lens = Optic()

lens.add_surface(index=0, radius=np.inf, thickness=10_000)
lens.add_surface(
    index=1, surface_type="paraxial", f=-50.0, thickness=50.0, is_stop=True
)
lens.add_surface(index=2)
lens.set_aperture(aperture_type="float_by_stop_size", value=5)
lens.add_wavelength(value=0.55, is_primary=True)
print(lens.paraxial.f2())
```
Output: `50.0`. The code contained a `be.abs([...])` statement that, when removed, would return the correct value of -50 for this negative lens. However, it would also return a negative value for compound lenses that, overall, have a positive focal length, but contain an internal focus, as the sign of the ray's y-location is not taken into account. 

# Implementation
I've updated the logic, where I hopefully make the correct choice depending on the ray location, slope, and whether or not there are internal foci:
- If the ray is exactly at the center of the lens surface, that counts as an internal (positive) focus
- If the ray has a positive starting point, but negative slope, it's a positive lens (system)
- Likewise, a negative starting point but positive slope also means it's a positive lens
- If the ray and slope are both positive or negative:
    - There is no internal focus: the lens system is negative
    - There is an internal focus: the lens system is positive
 
I've checked that a lens system is considered positive, even if the focal plane is inside the system, with Winlens3D (Basic) and OpTaliX. All (py) tests I could think of now check out, but it's worth having a critical review. FWIW, all new pytests that I thought of fail on `master`, which I think is a good sign.

I've learned more about paraxial ray tracing than I ever wanted to (also, not as trivial as it sounds 😅 )

### DIFF
--- a/optiland/interactions/thin_lens_interaction_model.py
+++ b/optiland/interactions/thin_lens_interaction_model.py
@@ -107,7 +107,7 @@ class ThinLensInteractionModel(BaseInteractionModel):
         # paraxial approximation -> direction is not necessarily unit vector
         rays.L = L
         rays.M = M
-        rays.N = be.ones_like(L)
+        rays.N = be.copysign(be.ones_like(rays.N), rays.N)
         rays.is_normalized = False
 
         return rays
--- a/tests/test_paraxial.py
+++ b/tests/test_paraxial.py
@@ -1,6 +1,7 @@
 import pytest
 
 import optiland.backend as be
+from optiland.materials import IdealMaterial
 from optiland.optic import Optic
 from optiland.paraxial import Paraxial
 from optiland.samples.eyepieces import EyepieceErfle


rjmoerland:
No worries, all good. I realized in the mean time that this does not fix the sign for mirror optics. That'll need another fix. 

I'm not sure you can get away from considering internal foci, but if you can that's great, and for sure I'd appreciate it if you critically look at my reasoning. 

I was also thinking about a more general approach, since the code (seems to) assume that all optical elements are on a straight line, even though the real rays can be traced in 3D space. It would be great if the paraxial tracing would equally "not care" about 45° mirrors and such. I am thinking along the lines of tracings a (paraxial) ray along the optical axis with normal (x, y, z) = (0,0,1), and tracing a ray at (say) +1.0 Y with normal (0,0,1). Where these two rays end up on the last surface defines the orientation of the projection of the coordinate system. The +y axis' origin on the last surface is wherever the paraxial ray along the optical axis ended, and the positive direction is wherever the ray starting at y=+1 ended up. 

I think that the same information that currently is used in the code (plus or minus y, plus or minus u) is present in those vectors, and my expectation is that the sign of the cross product of the "local" y and the ray leaving the last surface should tell us something about the sign of the focus. Would need to write it out to be sure though.

rjmoerland:
Hey @HarrisonKramer , did you manage to find some time to look into a potential simplification?

HarrisonKramer:
Hi @rjmoerland and @ZvejysMegejas,

Sorry for the long delay! I missed this one.

I revisited this and think we can solve it by making two changes:
1. Remove the `abs`, which was an attempted band-aid (I think) for reflective systems.
2. Use the correct definition for effective focal length: $f = \frac{-y_{in}}{n_{final}' * u_{final}'}$. Here we need to simply add the image-space refractive index, and note that after a mirror, the refractive index change sign.

I am not certain, but I think the `abs` was added because the image-space index was omitted. I don't recall the origin, but the current implementation is indeed wrong. I prefer not to overcomplicate this (if not needed) by checking the number of crossing of the optical axis. That could fail for "exotic" systems.

But please, check my logic. Do you agree with the simple approach above? I quickly checked this against a commercial tool and see it gives the correct signs.

Thanks,
Kramer

HarrisonKramer:
Hi @rjmoerland,

yes, that would be great if you can implement this, and thanks for double checking the calculation against other tools.

You can confirm the logic, but you should be able to retrieve the _unsigned_ image-space index via the standard `Optic.n()` method. Then, you can likely just count the number of reflective surfaces in the system (via the `is_reflective` attribute of the surface interaction models) - an odd number of reflective surfaces indicates the index should be negative. There may be an easier/cleaner way if you want to explore that.

Thanks again,
Kramer

rjmoerland:
Let me know how you would like to proceed. We can either roll back the merge and I continue on this branch, or I create a new branch based on the current version in `master`

PR #388, opened 2025-10-29
Return array for Abbe material extinction coefficient

crnh:
Ray tracing fails when using Abbe materials with the Torch backend, because `propagate` uses `be.any` to check if it should account for extinction:

https://github.com/HarrisonKramer/optiland/blob/399e9d1bbb9c9a719f5b2700a75c8f64fb72c95e/optiland/propagation/homogeneous.py#L48

Torch's `any` requires a tensor and fails when passed a single boolean.

### DIFF
--- a/optiland/backend/__init__.pyi
+++ b/optiland/backend/__init__.pyi
@@ -16,6 +16,7 @@ __all__ = [
 ]
 
 ndarray: TypeAlias = NDArray | Tensor  # noqa: PYI042
+inf: float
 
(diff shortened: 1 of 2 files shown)


HarrisonKramer:
Thanks for fixing this. PR looks good. Feel free to merge it when ready.

Regarding your comment on `IdealMaterial` - If you like, you can change this such that a backend array is always returned. That can be a part of this PR or another. I can also help there. Just let me know.

Regards,
Kramer

HarrisonKramer:
I will go ahead and merge this one.

PR #421, opened 2025-12-22
Fix Forbes surface tutorial notebook

Xiong-Ziyi:
- Use implemented freeform_coeffs key format for forbes_q2d ("a"/"b" tuple keys).
- Set field type before drawing to avoid field_definition=None.


HarrisonKramer:
Hi @Xiong-Ziyi,

Thanks for making this change! Merging now.

Kramer

PR #470, opened 2026-02-11
Feat/extended sources

manuelFragata:
feat: Extended source modelling with SMF source, irradiance compatible analysis, and visualization of source properties.

Adds extended source modeling to Optiland, enabling physically based ray generation from fiber sources and irradiance analysis on detector surfaces.

- [BaseSource] / [SMFSource] — abstract source interface + singleMode fiber source with Gaussian spatial/angular distributions via Sobol quasi-random sampling
- [ExtendedSourceOptic] - wrapper that pairs a source with an [Optic], providing source based [`trace()`] and [`draw()`] while delegating all standard optic properties.-
- [IncoherentIrradiance] - updated to accept a `source=` parameter for detector plane irradiance computation
- [SourceViewer] - 6 panel diagnostic visualization for source ray distributions
- Torch backend fix - `Lens2D` XY projection now converts tensors to floats before passing to matplotlib
- Documentation - API references, functionalities page, learning guide section, and gallery notebook
- suite of tests covering all new components with both numpy and torch backends

Overall, I think that the current [`ExtendedSourceOptic`] wrapper pattern is the right start for us to later on separate for example how rays interact with surfaces (e.g. sequential vs. mixed/non-sequential)...

### DIFF
--- a/docs/LEARNING_GUIDE.md
+++ b/docs/LEARNING_GUIDE.md
@@ -99,14 +99,17 @@ Welcome to the Optiland Learning Guide! This guide walks you through key concept
         - Adding new coating types
     - [Tutorial 10c - Custom Optimization Algorithms](https://github.com/HarrisonKramer/optiland/blob/master/docs/examples/Tutorial_10c_Custom_Optimization_Algorithm.ipynb)
         - Creating a "random walk optimizer" to optimize an aspheric singlet
-11. **Machine Learning in Optical Design** - note that these notebooks are hosted in the [LensAI repository](https://github.com/HarrisonKramer/LensAI)
-    - [Tutorial 11a - Random Forest Regressor to Predict Optimal Lens Properties](https://github.com/HarrisonKramer/LensAI/blob/main/notebooks/Example_1/Singlet_RF_Model_RMS_Spot_Size.ipynb)
+11. **Extended Sources**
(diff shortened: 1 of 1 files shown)


HarrisonKramer:
Thanks for submitting this, @manuelFragata! It is quite an extensive feature and looks very cleanly implemented. I've been quite busy, but I am hoping to formally review later this week. I will follow up here :)

I don't know why the docs build failed, but I'll check.

Thanks,
Kramer

optiland/sources/visualization.py
```diff
@@ -0,0 +1,341 @@
...
+        z = be.to_numpy(rays.z)
+        L = be.to_numpy(rays.L)
+        M = be.to_numpy(rays.M)
+        N = be.to_numpy(rays.N)
+
+        # --- Calculate Theoretical Radiance for Visualization ---
+        # This value is for color-coding only and is separate from rays.i (power)
+        from optiland.sources.smf import SMFSource
+
+        if isinstance(self.source, SMFSource):
```
HarrisonKramer, line 105: this is fine for now, but if we add signifcantly more source types, then we might want to restructure the ray coloring. Let's just keep that in mind. No action for now needed

HarrisonKramer:
A great addition, @manuelFragata! I can definitely see us extending this part of the package further. Maybe an interesting direction in the near term.

Merging now.

Kramer

PR #514, opened 2026-03-12
Paraxial surfaces: fixes for tracing and rendering

rjmoerland:
**Background**
While working on #382 I ran into a few issues regarding the handling of paraxial mirrors, and the backward tracing of real rays through paraxial surfaces. Those issues lead to erroneous results for the locations and magnitudes of the front and back focal plane when paraxial mirrors are involved. In addition, the RealRay interaction with paraxial surfaces was not handling backward tracing well, leading to confusing ray diagrams that did not reflect the expected outcome. 

This PR hopes to address these shortcomings, and consists of two parts: 
1. Fixing the tracing of paraxial rays through the system that includes rays bouncing back from paraxial mirrors.
2. Fixing the interaction of real rays with paraxial surfaces, in order to fix the display of the ray diagram. 

**Approach** 
The tracing of rays through real surfaces seems to be accurate regardless the direction, and also the paraxial calculations (`f1`, `f2`, etc.) seem accurate. Therefore, I've used the outcome of "real" models as the "gold standard" to compare the results with when using paraxial surfaces.

Sign convention:
For mirrors and a collimated beam coming from the left, the rays will be focused to the left of the surface when the focal length is negative. For a positive focal length, the rays will seem like coming from the right of the surface. The sign is flipped for rays coming from the right, i.e., after having reflected off an mirror earlier in the path. This sign convention seems to be in step with real mirrors: for a collimated beam coming from the left, a negative radius yields a real focus on the left, and a positive radius gives a virtual focus on the right. For rays from the right, it's vice versa.

**Examples**
Below I show a few examples of problematic lens systems. I tabulated the outcome of the paraxial calculations as obtained from `master` (at the time of writing based on [this](https://github.com/rjmoerland/optiland/commit/ebd2de23cab280bd10773ae27d5624055a49f611) commit ), together with the outcomes obtained with the code in this PR (labeled as "This PR") and the outcome of paraxial calculations for an equivalent "real" lens system. Sometimes the results from `master` and this PR are identical, but then the ray diagram is off, sometimes both are off.

Below each table I plot the ray diagrams as obtained with the `master` branch, this branch, and the equivalent real systems. Also here, the paraxial ray diagrams obtained with this branch and the equivalent real systems are identical.

## Example 1

A single paraxial mirror, with focal length -50 (real focus to the left):

|         | f1 | F1 | P1 | f2 | F2 | P2 |
|---------|----|----|----|----|----|----|
| Master  | 50.000   | 50.000   |  0.000  | -50.000   |  -50.000  | 0.000  |
| This PR  | -50.000   | -50.000   |  0.000  | -50.000   |  -50.000  | 0.000  |
| Real mirror*   | -50.000   | -50.000   |  0.000  | -50.000   |  -50.000  | 0.000  |

*R = -100

**master**
image

**This PR**
image

**Reference equivalent system**
image

## Example 2

A paraxial lens with f=50, followed by a flat mirror at 10 units distance, reflecting the light back through the lens:

|         | f1 | F1 | P1 | f2 | F2 | P2 |
|---------|----|----|----|----|----|----|
| Master  | -31.250   | -18.750   |  12.500  | -31.250   |  -18.750  | 12.500  |
| This PR  | -31.250   | -18.750   |  12.500  | -31.250   |  -18.750  | 12.500  |
| Real lens*  | -31.250   | -18.750   |  12.500  | -31.250   |  -18.750  | 12.500  |

*Lens consisting of two surfaces with R = 100 and R = -100, respectively, with n = 2 and thickness 0.

**master**
image

**This PR**
image

**Reference equivalent system**
image

## Example 3 

Two paraxial mirrors, with f = -50 and f = 50 respectively, and a gap of 20 in between:

|         | f1 | F1 | P1 | f2 | F2 | P2 |
|---------|----|----|----|----|----|----|
| Master  | 20.833   | 29.167   |  8.333  | 31.250   |  -1.250  | -32.500  |
| This PR  | -31.250   | -18.750   |  12.500  | 31.250   |  -1.250  | -32.500  |
| Real mirrors*  | -31.250   | -18.750   |  12.500  | 31.250   |  -1.250  | -32.500  |

*Real mirrors with radius -100 and +100, respectively.

**master:**
image

**This PR:**
image

**Reference equivalent system**
image

**Final remark**
_As far as my understanding goes_, the paraxial ray tracer seems like it was written largely with linear optical lens systems in mind, and perhaps at a time when computing power was more limited. Optiland's real ray tracing has no issues with tracing rays through 3D space, and though I don't know what the best approach would be, I would like to see a fully vectorial approach for paraxial rays as well. Ideally, surfaces would be placed in the optical system with well-defined surface normals, and those normals would be used for paraxial ray tracing. In that case, there could be no confusion about the direction from which light is hitting the element. But, perhaps this is already considered in the non-sequential mode?

As a thought, perhaps attaching a `ThinLensInteractionModel` to every real surface, to be used during paraxial ray tracing only, could be a way to leverage the current RealRay tracing engine to obtain paraxial results relatively easily?

### DIFF
--- a/optiland/interactions/thin_lens_interaction_model.py
+++ b/optiland/interactions/thin_lens_interaction_model.py
@@ -71,17 +71,25 @@ class ThinLensInteractionModel(BaseInteractionModel):
         rays.opd = rays.opd - (rays.x**2 + rays.y**2) / (2 * self.f)
 
         n1 = self.material_pre.n(rays.w)
+        n2 = n1 if self.is_reflective else self.material_post.n(rays.w)
+        L, M, N = [component / be.abs(rays.N) for component in (rays.L, rays.M, rays.N)]
+        if not be.isinf(self.f):
+            if self.is_reflective:
+                f1 = f2 = -self.f * be.copysign(be.ones_like(rays.N), rays.N)
+            else:
+                f = self.f * be.copysign(be.ones_like(rays.N), rays.N)
+                f1 = f * n1
+                f2 = f * n2
+            L = L * f1 - rays.x
+            M = M * f1 - rays.y
+            N = be.where(rays.N > 0, f2, -f2)
+            if self.f < 0:
+                L = -L
+                M = -M
+                N = -N
 
-        n2 = -n1 if self.is_reflective else self.material_post.n(rays.w)
-
-        ux1 = rays.L / rays.N
-        uy1 = rays.M / rays.N
-
-        ux2 = 1 / n2 * (n1 * ux1 - rays.x / self.f)
-        uy2 = 1 / n2 * (n1 * uy1 - rays.y / self.f)
-
-        L = ux2
-        M = uy2
+        else:
+            N *= n2 / n1
 
         # only normalize if required
         if self.bsdf or self.coating or isinstance(rays, PolarizedRays):
@@ -104,11 +112,13 @@ class ThinLensInteractionModel(BaseInteractionModel):
             # update polarization matrices, if PolarizedRays
             rays.update()
 
-        # paraxial approximation -> direction is not necessarily unit vector
+        if self.is_reflective:
+            N = -N
         rays.L = L
         rays.M = M
-        rays.N = be.copysign(be.ones_like(rays.N), rays.N)
-        rays.is_normalized = False
+        rays.N = N
+
+        rays.normalize()
 
         return rays
 
--- a/optiland/paraxial.py
+++ b/optiland/paraxial.py
@@ -67,7 +67,7 @@ class Paraxial:
         """
         z_start = -1
         wavelength = self.optic.primary_wavelength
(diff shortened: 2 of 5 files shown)


HarrisonKramer (approved):

rjmoerland:
@HarrisonKramer kind reminder that this one is good to go from my end :)

HarrisonKramer:
Hi @rjmoerland,

Sorry for the delay. It's been a busy week.

I will go ahead and merge this now. I need to think more about refactoring the raytracers. I do understand your rationale and I think you might be right here, but I'm just not sure how best to do this, nor if there would be unintended consequences. That would be a large refactor that touches most of the codebase, either directly or indirectly.

And on the new tutorial, feel free to submit something if you like. I'm also happy to have more examples/docs.

Thanks again,
Kramer

PR #522, opened 2026-03-15
Replace tabulated Gaussian quadrature with rule

rjmoerland:
## Rationale
Gaussian quadrature is an efficient scheme for numerical integration. The current GaussianQuadrature distribution in Optiland, for integration over the unit disk, uses tabulated numbers for orders 1 to 6. Higher order integration is therefore limited to other distribution types, that may be suboptimal, such as uniform or random sampling. 

This PR implements a new GaussianQuadrature class, a subclass of BaseDistribution, where the number of rings and spokes can be adjusted at will. The correct locations are calculated on the fly, together with the integration weights.

## Implementation

The implementation is based on the paper by William H. Peirce, "Numerical Integration Over the Planar Annulus,",  Journal of the Society for Industrial and Applied Mathematics, Vol. 5, No. 2 (Jun., 1957), pp. 66-73. 

I believe it is a drop-in replacement for many cases, but I still have open questions about the weights: 

1.  The typical normalization of the weights is such that the sum is equal to 1.0 or π. However, the current Optiland implementation uses a value of 1.0 if `is_symmetric` is `False` and 3.0 if `is_symmetric` is `True`. Why the difference?
1. The description of `is_symmetric` and the implementation seem to differ - according to the text, `is_symmetric` is to indicate symmetry in y, but activating it seems to indicate that the distribution is circularly symmetric. Not activating it seems to imply mirror symmetry in x. Is this correct, and should it be maintained?
1. It is not clear to me what exactly is the purpose of the symmetric/asymmetric distribution toggle in [`ray.py`](https://github.com/HarrisonKramer/optiland/blob/5a3cf127c623f261c83af560002cab6754c2a386/optiland/optimization/operand/ray.py#L371), when `Hx == Hy == 0`. I tried to maintain the functionality of `OPD_difference`, but I will need a little help because I don't understand the weights 😶 

As a consequence of my lack of understanding of these details, there are a few pytests failing.


HarrisonKramer (approved):

HarrisonKramer:
Thanks for finalizing this, @rjmoerland. If you're happy with it, feel free to merge. I believe you have write privileges.

-Kramer

rjmoerland:
Sorry, I merged but I forgot to ask what your preferred merging strategy is 🤦

HarrisonKramer:
No worries, thanks for merging. I generally prefer to squash and merge to keep the history on master cleaner. I should probably update that detail in the contributing docs.

rjmoerland:
Yeah, squash merging makes total sense, sorry again!

PR #553, opened 2026-04-03
Live optimization plotter

drpaprika:
Hi all,

This PR addresses issue [#518](https://github.com/optiland/optiland/issues/518) by introducing `LiveOptimizationPlotter`. It enables real-time visualization of the lens and merit function during optimization outside of the GUI.

If this approach is validated, I'll add the plot option to all SciPy and Torch-based optimizers.
I'm also open to suggestions on how to add tests for the class.

drpaprika


HarrisonKramer:
Hi @drpaprika,

Sorry for the long delay. Thanks for adding this functionality. I tested it and it works well. I also merged the latest master into this branch and pushed the changes.

On testing, I did not look into it deeply, but I expect you will want to patch the plotting functions, so that running the tests don't actually generate matplotlib plots. We use a similar strategy in testing for the 2D/3D visualization. You can possibly mock other inputs for speed, e.g., creating a simple optimizer mock so that no optimization needs to be performed during the test.

Do you want to go ahead and add the new parameter to each of the optimization classes? I can also help, if needed. Just let me know.

Regards,
Kramer

drpaprika:
Hi @HarrisonKramer,
No hurry no worries ! Thanks for reviewing the PR, I added the `plot` parameter to the other optimization classes and added a few tests for `LiveOptimizationPlotter`.

Note that there is one edge case: a test in `TestLeastSquares` fails only on Python 3.10. This happens because I added a `callback` argument to align this API with the other optimizer classes.

`scipy.optimize.least_squares()` only gained callback support in SciPy 1.16, and there appears to be a compatibility issue with Python 3.10 in this specific case.

I can remove `least_squares()`'s callback argument so that this test passes, but anyways  since Python 3.10 reaches end of life in October 2026, I suggest we consider dropping support for it soon.

HarrisonKramer:
Hi @drpaprika,

Based on your suggestion, I decided to drop 3.10 support and add 3.14 support. I updated the CI and merged master into this branch. I will merge it as soon as tests pass. Thanks again!

Kramer

PR #555, opened 2026-04-04
Particle Swarm Optimization (PSO)

drpaprika:
Hi, 
This PR introduces Particle Swarm Optimization (PSO), a population-based derivative-free optimizer.
It is called using:

```python3
optimizer = optimization.ParticleSwarm(problem)
res = optimizer.optimize()
```

image

I added extensive comments to guide the reader through the thought process behind PSO, which is suprisingly elegant and conceptually simple when you come to think of it!

Best,
drpaprika


manuelFragata:
Hi @drpaprika ,

nice! curious about the performance, did you do any benchmark?

drpaprika:
Hi @manuelFragata I did not do one formally but after trying on a few lenses my experience is that gradient-based optimizers work very well on most optical systems : if the solutions are comparable PSO seems to, in average, converge slower than L-BFGS-B. I say average because PSO is non deterministic so I believe a benchmark would need a Monte Carlo approach as well as meta-parameters tuning

HarrisonKramer:
Hi @drpaprika,

Thanks a lot for adding this! This is indeed an elegant optimization approach, and I appreciate the extensive comments you added.

I took the liberty of adding some tests and an example in the docs. I will go ahead and merge now, as all looks good!

Regards,
Kramer

PR #563, opened 2026-04-17
Fix/phase height profile and loading

### LINKED ISSUE #562: HeightProfile uses direct material instead of surface media, and phase_profile deserialization is broken
### Checklist

- [ x] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [ x] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [ x] I am using the **latest version** of Optiland (if not, please update and retry).
- [ x] I have tried to reproduce or debug the issue myself before opening this.
- [ x] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

### Bug Report

**Describe the bug**
There are two related bugs affecting phase-based surfaces and model loading:

1. `HeightProfile` keeps its own material reference and uses it directly to compute phase and gradients. This is incorrect because the phase behavior should come from the surface media, not from a material stored directly in `HeightProfile`. In practice, this makes the phase delay behave like it is using a fixed `(n - 1)` style assumption instead of the actual refractive index contrast across the surface. The fix removes this direct material dependency from `HeightProfile` and makes it derive the refractive indices from `parent_surface.material_pre` and `parent_surface.material_post`.

2. Interaction models do not properly deserialize `phase_profile` objects when loading from dictionaries. As a result, serialized optical systems using phase interaction models may fail to reconstruct the phase profile correctly, or may load incomplete interaction state.

This branch fixes both problems by:
- removing the direct material usage from `HeightProfile`
- making `HeightProfile` use `parent_surface.material_pre` and `parent_surface.material_post`
- restoring proper `phase_profile` deserialization in interaction models

**To Reproduce**

1. Create a phase surface using `HeightProfile`
2. Assign different materials before and after the surface
3. Evaluate the phase or gradient produced by the profile
4. Observe that the result is computed from the material stored directly in `HeightProfile`, instead of from the actual surface media

And/or:

1. Serialize an optical system or interaction model that contains a `phase_profile`
2. Load it back using the corresponding `from_dict` path
3. Observe that the `phase_profile` is not reconstructed correctly in the interaction model

**Expected behavior**
- `HeightProfile` should not carry or use a direct material reference for phase computation
- `HeightProfile` should derive refractive behavior from `parent_surface.material_pre` and `parent_surface.material_post`
- Phase delay and gradients should use the refractive index difference across the surface
- Interaction models should deserialize `phase_profile` objects correctly so serialized phase systems can be loaded without losing behavior

**Screenshots**

**Environment**  
- Optiland Version: `master @ 75859934`
- Python Version: `3.11`
- OS: `Linux`
- Additional Info: `pytest tests/test_height_profile_phase.py` passes locally after the fix

**Additional context**
This report corresponds to the following two commits in the branch:

- `d0a1348` - `Make HeightProfile use surface pre/post materials`
- `9ee5e7b` - `Fix phase profile deserialization in interaction models`

The first change removes the direct material dependency from `HeightProfile` and makes the phase profile computation follow the materials defined on the parent surface. The second change fixes the deserialization path for `phase_profile` in interaction models so saved systems can be loaded correctly.

gustavu92:
## Summary

This PR fixes the two bugs reported in #562: `HeightProfile` was using a direct material reference instead of the parent surface media, and `phase_profile` objects were not being deserialized correctly in interaction models. :contentReference[oaicite:0]{index=0}

## Changes

- remove the direct material dependency from `HeightProfile`
- make `HeightProfile` derive refractive behavior from `parent_surface.material_pre` and `parent_surface.material_post`
- fix `phase_profile` deserialization in interaction models
- keep phase-based systems loadable after serialization/deserialization :contentReference[oaicite:1]{index=1}

## Why

`HeightProfile` phase and gradient calculations should be based on the refractive index contrast across the surface, not on a material stored directly in the profile. The issue report notes that the previous behavior effectively acted like a fixed `(n - 1)` assumption instead of using the actual surface media. :contentReference[oaicite:2]{index=2}

The issue also reports that interaction models could fail to reconstruct `phase_profile` correctly when loading from dictionaries, which could break serialized phase-based optical systems or leave them partially reconstructed.

## Validation

- `pytest tests/test_height_profile_phase.py` passes locally after the fix :contentReference[oaicite:4]{index=4}

## Environment

- Optiland Version: `master @ 75859934`
- Python Version: `3.11`
- OS: `Linux`

## Related

Closes #562

### DIFF
--- a/optiland/interactions/base.py
+++ b/optiland/interactions/base.py
@@ -81,27 +81,31 @@ class BaseInteractionModel(ABC):
         }
 
     @classmethod
-    def from_dict(cls, data, parent_surface):
-        """Creates an interaction model from a dictionary representation."""
+    def _deserialize_init_data(cls, data):
         from optiland.coatings import BaseCoating
         from optiland.scatter import BaseBSDF
 
+        init_data = data.copy()
+        init_data.pop("type", None)
+        init_data.pop("material_pre", None)
+
+        if init_data.get("coating") is not None:
+            init_data["coating"] = BaseCoating.from_dict(init_data["coating"])
+
+        if init_data.get("bsdf") is not None:
+            init_data["bsdf"] = BaseBSDF.from_dict(init_data["bsdf"])
+
+        return init_data
+
+    @classmethod
+    def from_dict(cls, data, parent_surface):
+        """Creates an interaction model from a dictionary representation."""
         interaction_type = data["type"]
         subclass = cls._registry.get(interaction_type)
         if subclass is None:
             raise ValueError(f"Unknown interaction model type: {interaction_type}")
 
-        # Remove 'type' from data to avoid passing it to the constructor
-        init_data = data.copy()
-        init_data.pop("type")
-        # Ignore 'material_pre' that might be present in older files but is obsolete:
-        if "material_pre" in init_data:
-            init_data.pop("material_pre")
-
-        if "coating" in init_data and init_data["coating"] is not None:
-            init_data["coating"] = BaseCoating.from_dict(init_data["coating"])
-        if "bsdf" in init_data and init_data["bsdf"] is not None:
-            init_data["bsdf"] = BaseBSDF.from_dict(init_data["bsdf"])
+        init_data = subclass._deserialize_init_data(data)
 
         return subclass(
             parent_surface=parent_surface,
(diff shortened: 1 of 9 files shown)


HarrisonKramer:
Thanks for the fix, @gustavu92. Merging now.

PR #564, opened 2026-04-18
fix: force XCB on Linux

### LINKED ISSUE #556: No Wayland support?
### Checklist

### Bug Report

With `QT_QPA_PLATFORM=wayland` in the environment, or with it unset, I'm getting the following errors when launching `optiland` gui - both when running the `optiland` command, and when running `python -m optiland_gui.run_gui`:

```
X Error of failed request:  BadWindow (invalid Window parameter)
  Major opcode of failed request:  12 (X_ConfigureWindow)
  Resource id in failed request:  0xc3dc98b0
  Serial number of failed request:  7
  Current serial number in output stream:  8
QThreadStorage: entry 1 destroyed before end of thread 0x557ec2eaa2f0
QThreadStorage: entry 0 destroyed before end of thread 0x557ec2eaa2f0
```

I managed to workaround the issue with `env QT_QPA_PLATFORM=xcb optiland`. I build the application with Nix, in order to distribute it to Nixpkgs, so I'm not sure how to help you reproduce it. Using `env QT_QPA_PLATFORM=wayland` might reproduce the issue.

**Expected behavior**

No error, and Qt wayland backend is being used.

**Environment**  
- Optiland Version: `0.6.0`
- Python Version: `3.13`
- OS: [NixOS](https://nixos.org/) (GNU/Linux)
- Additional Info: I'm trying to distribute this package to [Nixpkgs](https://github.com/NixOS/nixpkgs) - Nix' package collection.

**Additional context**

I encountered another Xorg only Qt application called GoldenDict-Ng, and there they have managed to hard code the application to always use Xorg Qt backend, no matter what is `QT_QPA_PLATFORM` (or any other environment variables). I'm not sure how they implemented it, but here's the relevant issue in their tracker:

- https://github.com/xiaoyifang/goldendict-ng/issues/349

A similar approach might work for you if you are too constraint to use only XCB Qt backend.

HarrisonKramer:
## Description
This PR resolves a critical GUI crash affecting Linux users on Wayland-based display servers (e.g., NixOS, Ubuntu 24.04).

The crash was identified as a protocol mismatch between the Qt Wayland backend and the VTK/OpenGL rendering pipeline, resulting in a `BadWindow (invalid Window parameter)` error during `X_ConfigureWindow` requests. 

This change programmatically forces the `xcb` (X11) platform abstraction layer when running on Linux. This ensures the GUI utilizes XWayland, which provides a stable environment for the current rendering stack while maintaining compatibility across modern Linux distributions.

## Changes
- **`optiland_gui/run_gui.py`**: Added logic to detect Linux environments and explicitly set `QT_QPA_PLATFORM=xcb` before the `QApplication` instance is initialized.

**Closes: #556**

### DIFF
--- a/optiland_gui/run_gui.py
+++ b/optiland_gui/run_gui.py
@@ -11,6 +11,7 @@ Authors:
 from __future__ import annotations
 
 import ctypes
+import os
 import sys
 
 from PySide6.QtCore import QLocale, QSize, Qt
@@ -27,6 +28,12 @@ def main() -> None:
     if sys.platform == "win32":
         myappid = f"{ORGANIZATION_NAME}.{APPLICATION_NAME}.1.0"
         ctypes.windll.shell32.SetCurrentProcessExplicitAppUserModelID(myappid)
+
+    # Force XCB (X11) on Linux to avoid BadWindow / X_ConfigureWindow crashes
+    # caused by incompatibility between the VTK/OpenGL render pipeline and the
+    # Qt Wayland backend.  This must be set before QApplication is created.
+    if sys.platform.startswith("linux"):
+        os.environ.setdefault("QT_QPA_PLATFORM", "xcb")
 
     app = QApplication(sys.argv)
     app.setWindowIcon(QIcon(OPTILAND_ICON_PATH))


optiland_gui/run_gui.py
```diff
@@ -28,6 +29,12 @@ def main() -> None:
         myappid = f"{ORGANIZATION_NAME}.{APPLICATION_NAME}.1.0"
         ctypes.windll.shell32.SetCurrentProcessExplicitAppUserModelID(myappid)
 
+    # Force XCB (X11) on Linux to avoid BadWindow / X_ConfigureWindow crashes
+    # caused by incompatibility between the VTK/OpenGL render pipeline and the
+    # Qt Wayland backend.  This must be set before QApplication is created.
+    if sys.platform.startswith("linux"):
+        os.environ.setdefault("QT_QPA_PLATFORM", "xcb")
```
doronbehar, line 36: @HarrisonKramer This is not good, because on my system this environment variable is set explicitly to `wayland`, and therefor the `setdefault` doesn't set anything. You should have wrote:

```suggestion
        os.environ.set("QT_QPA_PLATFORM", "xcb")
```

PR #579, opened 2026-05-11
Fix Torch autograd for multiple thickness variables (#569)

### LINKED ISSUE #569: Gradient broken for multiple thickness variables – only the last thickness gets gradient
### Checklist

- [√ ] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [√ ] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [√] I am using the **latest version** of Optiland (if not, please update and retry).
- [√] I have tried to reproduce or debug the issue myself before opening this.
- [√ ] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

### Bug Report

**Describe the bug**
When switching Optiland to the PyTorch backend using be.set_backend("torch") and enabling gradient tracking (be.grad_mode.enable()), optimizing multiple thickness variables simultaneously via OpticalSystemModule results in **only the last thickness variable receiving a non‑zero gradient**. All preceding thickness variables get None or zero gradient. A single thickness variable works fine, and radius, conic, or aspheric coefficient variables are unaffected. This breaks joint optimization of several thicknesses.

**To Reproduce**
The following case is exactly what I described in discussion #566.
import torch
import optiland.backend as be
from optiland import optic, optimization
from optiland.ml import OpticalSystemModule

be.set_backend("torch")
be.grad_mode.enable()

lens = optic.Optic()
lens.surfaces.add(index=0, thickness=be.inf)
lens.surfaces.add(index=1, thickness=7, radius=1000, material="N-SF11", is_stop=True)
lens.surfaces.add(index=2, thickness=30, radius=-1000)
lens.surfaces.add(index=3)
lens.set_aperture(aperture_type="EPD", value=15)
lens.fields.set_type(field_type="angle")
lens.fields.add(y=0)
lens.wavelengths.add(value=0.55, is_primary=True)

problem = optimization.OptimizationProblem()
input_data = {
    "optic": lens,
    "surface_number": -1,
    "Hx": 0, "Hy": 0,
    "num_rays": 5,
    "wavelength": 0.55,
    "distribution": "hexapolar",
}
problem.add_operand("rms_spot_size", target=0, weight=1, input_data=input_data)
problem.add_variable(lens, "radius", surface_number=1)
problem.add_variable(lens, "thickness", surface_number=1)
problem.add_variable(lens, "radius", surface_number=2)
problem.add_variable(lens, "thickness", surface_number=2)

model = OpticalSystemModule(lens, problem)

params = list(model.parameters())
param_thick1 = params[1]
param_thick2 = params[3]

optimizer = torch.optim.Adam(model.parameters(), lr=0.1)

losses = []
grad_t1 = []
grad_t2 = []

for step in range(250):
    optimizer.zero_grad()
    loss = model()
    loss.backward()

    g1 = param_thick1.grad
    g2 = param_thick2.grad
    grad_t1.append(g1.item() if g1 is not None else 0.0)
    grad_t2.append(g2.item() if g2 is not None else 0.0)

    optimizer.step()
    model.apply_bounds()
    losses.append(loss.item())
    problem.info()

#…Drawing section

**Screenshots**
Image

Image

Image

**Environment**  
- Optiland Version: `0.6.0`
- OS: `[Windows]`

**Possible cause speculationt**
The problem lies in set_thickness() in optic_updater.py. When multiple thickness variables are updated sequentially:
Update thickness of surface 1 → set_thickness(new_t1, 1) is called. It reads positions = self.optic.surfaces.positions (the current z-coordinates of all surfaces). Then it executes:

if hasattr(positions, "detach"):
    positions = positions.detach()

At this point positions is not yet connected to any current parameter (this is the first modification), so detach does no harm. After computing the new offsets, the z-coordinates of surfaces 2 and 3 become dependent on thickness1.

Update thickness of surface 2 → set_thickness(new_t2, 2) is called. Now the positions array already carries gradients with respect to thickness1. The second positions.detach() severs all gradient links to thickness1. Only the new computation for thickness2 remains in the graph. Consequently, only thickness2 retains a gradient.

**Possible solutions**
Add a new method set_thicknesses(self, updates: dict) to OpticUpdater that accepts all thickness changes at once, performs a single detach on the baseline positions, computes cumulative offsets, and writes the updated coordinates in one pass. Simultaneously modify OpticalSystemModule.forward() to collect all thickness variables and call the batch method.

The following patches have been proven effective in the testing environment and require further invasive modifications:

from optiland.optic.optic_updater import OpticUpdater
from optiland.ml import OpticalSystemModule

def set_thicknesses(self, updates: dict):
    positions = self.optic.surfaces.positions
    if hasattr(positions, 'detach'):
        positions = positions.detach()
    if positions.ndim > 1:
        positions = positions.squeeze(-1)
    n = len(self.optic.surfaces)
    offsets = torch.zeros(n, dtype=positions.dtype, device=positions.device)

    for surf_idx, new_val in updates.items():
        if surf_idx == 0:
            self.optic.surfaces[0].thickness = new_val
            self.optic.surfaces[0].geometry.cs.z = -new_val
            continue
        old_thick = positions[surf_idx + 1] - positions[surf_idx]
        delta = new_val - old_thick
        offsets[surf_idx + 1:] += delta
    new_positions = positions + offsets
    new_positions = new_positions - new_positions[1]  
    for k, surf in enumerate(self.optic.surfaces):
        surf.geometry.cs.z = new_positions[k].reshape(1)
    for surf_idx, new_val in updates.items():
        if surf_idx < n - 1:
            self.optic.surfaces[surf_idx].thickness = new_val

OpticUpdater.set_thicknesses = set_thicknesses

def patched_forward(self):
    vars_list = self._original_variables
    thickness_updates = {}

    for param, var in zip(self.params, vars_list):
        if var.variable.__class__.__name__ == 'ThicknessVariable':
            phys_val = var.variable.inverse_scale(param) 
            surf = var.variable.surface_number
            thickness_updates[surf] = phys_val
        else:
            var.variable.update_value(var.variable.inverse_scale(param))
    if thickness_updates:
        self.optic.updater.set_thicknesses(thickness_updates)
    self.problem.update_optics()
    return self.objective_fn()

OpticalSystemModule.forward = patched_forward

Image

**Additional information**
This is my first time submitting an issue. Please excuse any non-standard aspects. As a non-native English speaker, I used AI assistance to translate my thoughts. I would like to express my sincere gratitude to the author for their excellent work!

RayPragma:
## Summary

Fixes #569.

When optimizing more than one thickness variable simultaneously under the torch backend, only the **last** thickness in the update sequence received a non-zero gradient — earlier ones got `None`/zero. Radius, conic, and aspheric variables were unaffected.

## Root cause

`OpticUpdater.set_thickness` read `surfaces.positions` (current `cs.z`), called `positions.detach()` to drop stale graphs from prior iterations, then wrote new `cs.z` values incrementally.

The detach is necessary across iteration boundaries — but within a single forward pass, after the *first* thickness variable was updated, `cs.z` already encoded a dependency on that thickness tensor. The detach on the *next* update severed it, leaving only the last thickness tensor in the autograd graph.

The original report (#569) contains an excellent walk-through; this PR implements an alternative, less invasive fix than the one suggested there.

## Fix

Stop reading `cs.z` back and applying deltas. Treat each surface's `.thickness` attribute as the source of truth and rebuild every downstream `cs.z` from those thickness values:

```
cs.z[1] = 0  # anchor
cs.z[k] = cs.z[k-1] + surfaces[k-1].thickness   for k >= 2
```

This keeps every current-iteration thickness tensor connected to the output graph, and the explicit overwrite of `cs.z` from each rebuild naturally drops references to prior-iteration tensors — equivalent to the old `detach()` for cross-iteration cleanup, but without severing in-iteration grad paths. No changes to `OpticalSystemModule` or any callers needed.

### Minimal repro from the issue, before / after

```
   name |          value |             grad (before)  →   (after)
-------------------------------------------------------------------
radius1 |              9 |         0.169545             0.169545
 thick1 |           -0.3 |             None       →    -0.273316   ← fixed
radius2 |            -11 |        -0.150307            -0.150307
 thick2 |              2 |         -1.00153            -1.00153
```

### Test plan

- [x] New regression test `test_multiple_thickness_variables_all_get_gradients` in `tests/test_ml.py`, derived from the original report — asserts both thickness gradients are non-`None` and non-zero after one forward/backward
- [x] Full local test suite passes: **6050 passed, 9 skipped** (2 pre-existing `test_formula_4` failures unrelated — Windows GBK encoding when reading a yaml material file)
- [x] `ruff check optiland/optic/optic_updater.py` clean

### DIFF
--- a/optiland/optic/optic_updater.py
+++ b/optiland/optic/optic_updater.py
@@ -82,19 +82,27 @@ class OpticUpdater:
             # No need to shift other surfaces as they are relative to S1 at z=0
             return
 
-        positions = self.optic.surfaces.positions
-        # Detach positions used as reference points to prevent stale computation
-        # graphs (from prior iterations) from being pulled into the current graph.
-        if hasattr(positions, "detach"):
-            positions = positions.detach()
-        delta_t = value - positions[surface_number + 1] + positions[surface_number]
-        positions = be.copy(positions)  # required to avoid in-place modification
-        positions[surface_number + 1 :] = positions[surface_number + 1 :] + delta_t
-        positions = positions - positions[1]  # force surface 1 to be at zero
-        for k, surface in enumerate(self.optic.surfaces):
-            surface.geometry.cs.z = be.array(positions[k])
-        if surface_number < len(self.optic.surfaces):
-            self.optic.surfaces[surface_number].thickness = value
+        # Source of truth for downstream positions is each surface's `.thickness`
+        # attribute. Update the requested one first, then rebuild every cs.z
+        # downstream from those thickness values (rather than incrementally
+        # mutating cs.z). This keeps every current-iteration thickness tensor
+        # in the autograd graph — the prior incremental form needed `detach()`
+        # to drop stale graphs from earlier iterations, but that also severed
+        # in-iteration grad paths when multiple thickness variables were
+        # updated sequentially (only the last one kept its gradient). See #569.
+        surfaces = self.optic.surfaces
+        n = len(surfaces)
+        if surface_number < n:
+            surfaces[surface_number].thickness = value
+
+        # Surface 1 anchored at z=0; cs.z[k] = sum of upstream thicknesses.
+        if n >= 2:
+            z = be.array(0.0)
+            surfaces[1].geometry.cs.z = be.array(z)
+            for k in range(2, n):
+                t_prev = surfaces[k - 1].thickness
+                z = z + (t_prev if hasattr(t_prev, "detach") else be.array(t_prev))
+                surfaces[k].geometry.cs.z = be.array(z)
 
     def set_index(self, value: float, surface_number: int) -> None:
         """Set the index of refraction of a surface.
--- a/tests/test_ml.py
+++ b/tests/test_ml.py
@@ -184,3 +184,63 @@ class TestOpticalSystemModule:
 
         loss = module.forward()
         assert be.isclose(loss, torch.tensor(123.45, dtype=torch.float64))
+
+    def test_multiple_thickness_variables_all_get_gradients(self):
+        """
+        Regression test for issue #569: when more than one thickness variable
+        is optimized simultaneously, every thickness must receive a non-zero
+        gradient. Previously only the last thickness variable's gradient
+        survived because ``set_thickness`` called ``positions.detach()`` on
+        every invocation, which also severed the in-iteration gradient path
+        introduced by earlier thickness updates in the same forward pass.
+        """
+        from optiland import optic
+
+        be.grad_mode.enable()
+        lens = optic.Optic()
+        lens.surfaces.add(index=0, thickness=be.inf)
+        lens.surfaces.add(
+            index=1, thickness=7, radius=1000, material="N-SF11", is_stop=True,
+        )
+        lens.surfaces.add(index=2, thickness=30, radius=-1000)
+        lens.surfaces.add(index=3)
+        lens.set_aperture(aperture_type="EPD", value=15)
+        lens.fields.set_type(field_type="angle")
+        lens.fields.add(y=0)
+        lens.wavelengths.add(value=0.55, is_primary=True)
+
+        problem = OptimizationProblem()
+        input_data = {
+            "optic": lens,
+            "surface_number": -1,
+            "Hx": 0, "Hy": 0,
+            "num_rays": 5,
+            "wavelength": 0.55,
+            "distribution": "hexapolar",
+        }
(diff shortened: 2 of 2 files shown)


HarrisonKramer:
Hi @RayPragma,

Thanks a lot for the comprehensive report and for the fix! All tests pass. Merging now!

Regards,
Kramer

PR #580, opened 2026-05-11
Fix non-zero Seidel aberrations for reflective systems (#347)

### LINKED ISSUE #347: incorrect third order aberrations in reflective system
### Checklist

### Bug Report

**Describe the bug**
In version 0.5.6, I am getting 0 for all `seidels()` except S5 for a cassegrain-like system. I don't think it is the same as issue #300 because I have multiple fields and because my `invariant()` is nonzero. This issue reproduces with the similar system `optiland.samples.telescopes.HubbleTelescope`.

**To Reproduce**
```
from optiland.samples.telescopes import HubbleTelescope
HubbleTelescope().aberrations.seidels()
# => array([0.       , 0.       , 0.       , 0.       , 0.0014539])
```

**Expected behavior**
The `HubbleTelescope` is a two-mirror system so more of the aberration coefficients ought to be nonzero.

**Environment**  
- Optiland Version: 0.5.6
- Python Version: 3.10.11
- OS: Windows 11

RayPragma:
## Summary

Fixes #347.

For any system containing mirrors (e.g. Cassegrain telescopes), `aberrations.seidels()` previously returned `[0, 0, 0, 0, ε]` — the first four Seidel sums collapsed to zero across every reflective surface. The included `optiland.samples.telescopes.HubbleTelescope` reproduces it directly:

```python
>>> HubbleTelescope().aberrations.seidels()
array([0.       , 0.       , 0.       , 0.       , 0.0014539])  # before
```

## Root cause

Every per-surface Seidel term carries an `(n[k] - n[k-1])` factor. `surfaces.n()` returns raw refractive-index magnitudes — for a mirror, both pre- and post-surface media report `n = 1` (the `IdealMaterial` used for the `"mirror"` specifier), so the step is zero and the surface contributes nothing.

Only the distortion term has a non-`B` piece (the chief-ray-angle term `0.5 * (ub[k]**2 - ub[k-1]**2)`) that escapes this, which is why `S[4]` was the lone nonzero entry.

## Fix

The Welford / Smith third-order aberration formulas require the **signed-n convention**: after each reflection the post-surface medium's refractive index flips sign, and the flip propagates through subsequent surfaces. A new `Aberrations._signed_refractive_indices(wavelength)` walks the surface group applying this convention; `_precalculations` uses it in place of `surfaces.n()`.

```python
def _signed_refractive_indices(self, wavelength):
    n_raw = self.optic.surfaces.n(wavelength)
    n_signed = []
    sign = 1.0
    for k, surf in enumerate(self.optic.surfaces):
        if getattr(surf.interaction_model, "is_reflective", False):
            sign = -sign
        n_signed.append(sign * be.abs(n_raw[k]))
    return be.array(n_signed)
```

`surfaces.n()` itself is **untouched** — other consumers expect raw magnitudes, and this is a Seidel-specific convention rather than a global representation change.

### Hubble (the reported case)

```
seidels before fix: [0,       0,       0,         0,         0.0014539 ]
seidels after fix : [-2.71,   3.7e-2,  2.1e-3,   -1.3e-2,    1.1e-3   ]
```

All five entries nonzero. Per-surface contributions confirm both mirrors now contribute:

```
TSC: [ 0,  -73.97,   8.88 ]   # surf-1 (flat) / M1 / M2
CC : [ 0,    1.78,  -0.90 ]
TAC: [ 0,   -0.043,  0.092]
TPC: [ 0,    0.043, -0.351]
DC : [ 0,    0,      0.026]
```

### No regression on refractive systems

Refractive baselines (no mirrors → no sign flips → identical `n` arrays as before):

- `CookeTriplet`: identical seidels
- `DoubleGauss`, `EdmundSinglet`, `SingletStopSurf2`: existing recorded values unchanged

### Test plan

- [x] All 26 existing `tests/test_aberrations.py` tests pass on both `numpy` and `torch` backends
- [x] New regression test added: `TestReflectiveSystemSeidels::test_hubble_seidels_are_not_all_zero` — asserts all five entries are non-zero for `HubbleTelescope`
- [x] `ruff check optiland/aberrations.py` clean

## Caveat / open question

This PR ensures Seidel formulas evaluate to nonzero, correctly-signed values for mirror systems following the textbook (Welford ch. 7, Smith ch. 6.3) convention. The new Hubble numbers haven't been cross-checked against an external reference (e.g. Zemax) — Hubble's design uses hyperbolic mirrors that should largely cancel third-order spherical and coma, but the sample's exact conic constants may or may not match the textbook values. Happy to tighten the regression test if there's a canonical reference value the maintainers prefer.

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -36,6 +36,7 @@ We sincerely appreciate the contributions of the following individuals, whose ef
 - **cyberstudio10** (`GitHub <https://github.com/cyberstudio10>`_)
 - **lutzerb** (`GitHub <https://github.com/lutzerb>`_)
 - **Jan-Willem Beenakker** (`GitHub <https://github.com/jwmbeenakker>`_)
+- **RayPragma** (`GitHub <https://github.com/RayPragma>`_)
 
 
 Your contributions, whether in the form of code, documentation, feedback, or discussions, are what make **Optiland** better for everyone.
--- a/optiland/aberrations.py
+++ b/optiland/aberrations.py
@@ -279,7 +279,7 @@ class Aberrations:
         )
 
         # Convert Wavefront Seidel Sum to Transverse Aberration contribution
-        return -S_conic / (2 * self._n[-1] * self._ua[-1])
+        return S_conic / (2 * self._n[-1] * self._ua[-1])
 
     def _compute_over_surfaces(self, term_func: Callable) -> BEArray:
         """
@@ -375,8 +375,7 @@ class Aberrations:
             * i_val**2
         )
         spherical = term / (2 * self._n[k] * self._n[-1] * self._ua[-1])
-        # Negate to match the Welford/Zemax convention
-        return -spherical + self._get_conic_term(k, p_ya=4, p_yb=0)
+        return spherical + self._get_conic_term(k, p_ya=4, p_yb=0)
 
     def _TSC_term(self, k: int) -> float:
         """
@@ -391,7 +390,7 @@ class Aberrations:
         if self._on_axis:
             return self._TSC_on_axis_term(k)
         spherical = self._B[k - 1] * self._i[k - 1] ** 2 * self._hp
-        return -spherical + self._get_conic_term(k, p_ya=4, p_yb=0)
+        return spherical + self._get_conic_term(k, p_ya=4, p_yb=0)
 
     def _CC_term(self, k: int) -> float:
         """
@@ -404,7 +403,7 @@ class Aberrations:
             Computed sagittal coma term.
         """
         spherical = self._B[k - 1] * self._i[k - 1] * self._ip[k - 1] * self._hp
-        return -spherical + self._get_conic_term(k, p_ya=3, p_yb=1)
+        return spherical + self._get_conic_term(k, p_ya=3, p_yb=1)
 
     def _TAC_term(self, k: int) -> float:
         """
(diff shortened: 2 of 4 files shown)


HarrisonKramer:
Hi @RayPragma,

Thanks for submitting this PR. I completely agree with your assessment. Starting to review this, I realized that not only did we have a bug in the sign convention (which you fixed), but we also were omitting a correction for the conic/aspheric contribution. This is why the Seidel coefficients you reported after your fix were still somewhat large (particularly S1), while they're expected to be relatively small for this design.

I took your fix and added in the conic contribution. The result for Hubble is now:

```
seidels: [-2.80642537e-06,  3.37323248e-04, -1.45177194e-03, -1.28411858e-02, 7.59541065e-04]
```

I compared this to a manual calculation, which is spot on. Comparing to ZOS was also extremely close for S2-S5 (ignoring sign convention). The only discrepancy was S1, which was off by about 0.0005. I "perturbed" the conic constants so that the resulting aberrations were larger and in this case, we match nearly within numerical precision. This tells me there is likely some numerical quirk in ZOS when S1 is already quite small. Not really an issue, but it does differ from the hand calculations slightly.

I will plan to merge this in the next day or so. Feel free to review and provide feedback before then.

Regards,
Kramer

RayPragma:
Thanks @HarrisonKramer — the conic correction makes total sense, that
explains the residual S1 magnitude. all good. Please merge whenever.

PR #585, opened 2026-05-14
fix(zemax-reader): defensive parsing for legacy ZEMAX FTYP + paraxialImageFNO alias

RayPragma:
## Summary

Two small fixes that together unblock a class of legacy / mixed-format `.zmx` files that currently fail to load before any conversion logic runs.

### 1. `ZemaxDataParser._read_config_data` — defensive FTYP parsing

Older ZEMAX versions (e.g. `VERS 6133`) write `FTYP` with only 1–2 tokens, while the current parser assumes 8 and directly indexes positions 1/2/3/4/7:

```python
fields["num_fields"] = int(data[3])
...
fields["afocal_image_space"] = int(data[7]) == 1
```

This raises `IndexError` before any other operand is even parsed, killing the load. Introduce a small `_safe_int(idx, default)` helper so missing fields fall back to defaults instead of crashing. Also seed `fields['x']` / `fields['y']` to `[0.0]` and force `num_fields >= 1` so files that omit `XFLN`/`YFLN` entirely (legitimate single on-axis layout) don't trigger `KeyError` downstream.

### 2. `aperture.make_system_aperture` — accept `paraxialImageFNO` alias

Legacy Zemax files emit `paraxialImageFNO` for image-space F/# specified at the paraxial focus. It is numerically equivalent to `imageFNO` in the paraxial regime (differences only appear as second-order corrections in fast systems). Add a tiny alias so these files don't bounce off the registry lookup.

## Test plan
- [x] Existing `tests/test_fileio/test_zemax_reader.py` suite still passes (30/30)
- [x] New `test_read_config_data_legacy_short_ftyp` exercises the short-FTYP path and asserts the seeded defaults

### DIFF
--- a/optiland/aperture/__init__.py
+++ b/optiland/aperture/__init__.py
@@ -43,6 +43,14 @@ def make_system_aperture(aperture_type: str, value: float) -> BaseSystemAperture
         ValueError: If *aperture_type* is not a registered type.
 
     """
+    # Aliases for legacy Zemax aperture identifiers. ``paraxialImageFNO`` is
+    # the paraxial image-space F/#; in the paraxial regime it matches
+    # Optiland's ``imageFNO`` (differences only appear as second-order
+    # corrections in fast systems), which is good enough as a starting point.
+    _ALIASES = {
+        "paraxialImageFNO": "imageFNO",
+    }
+    aperture_type = _ALIASES.get(aperture_type, aperture_type)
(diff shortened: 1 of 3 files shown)


HarrisonKramer:
Hi @RayPragma,

Good fix! Thanks. Merging now.

Kramer

PR #599, opened 2026-05-18
Adds batched input support to image simulation and updates the API to be batch-first

gustavu92:
Adds batched input support to image simulation and updates the API to be batch-first.

API change:
- Before: ImageSimulationEngine(optic, source_image, config=config).run()
- Now: ImageSimulationEngine(optic, config=config).run(source_image)

Why:
This separates the optical simulation setup from the image data, which is cleaner for computer vision workflows. In training or inference loops, the optic/config stay fixed while each batch of images changes.

Highlights:
- Supports image paths, single images, and BCHW batched inputs
- Returns outputs as BCHW for compatibility with PyTorch models
- Processes convolution and distortion warping in batch
- Preserves gradient flow for PyTorch inputs
- Updates tests and gallery documentation


HarrisonKramer:
Hi @gustavu92,

Thanks for submitting this PR. It's a valuable addition.

I needed to merge a recent change from master, which fixed a subtle bug that had leaked in. Now all looks good. Merging now!

Regards,
Kramer

PR #605, opened 2026-05-22
Fix torch gradient graph retention in height interpolation

gustavu92:
## Summary

This PR fixes torch gradients for `HeightProfile`.

In the torch backend, `GridInterpolator.gradient()` computes `dh/dx` and `dh/dy` using `autograd.grad`. The current code uses `create_graph=False`, which computes the gradient values but does not keep the graph needed to differentiate them with respect to the height map.

Because of that, a trainable torch height map can be used in the forward ray trace, but losses that depend on the phase gradient cannot optimize the height map.

This PR changes the torch path to use:

`create_graph=True`
`retain_graph=True`

## Why

Phase surfaces use the height gradient to change ray directions. If the height map is trainable, the phase-gradient computation must remain differentiable.

The important path is:

`height parameter -> height_map -> HeightProfile.gradient() -> ray trace -> loss`

Before this change, the graph stopped at `HeightProfile.gradient()`.

## Validation

I added a small integration test that checks whether `HeightProfile._interpolate_gradient()` backpropagates to a trainable torch height parameter.

I also tested a minimal optimization where a scalar coefficient controls a defocus height map:

`height_map = c * (2r² - 1)`

Results over 10 runs:

Before:

- NumPy optimized correctly: `28.2843 -> 0.000214492`
- Torch did not optimize: `28.2837 -> 28.2837`
- Torch coefficient stayed at `0.02`

After:

- NumPy behavior is unchanged
- Torch optimized correctly: `28.2837 -> 0.00192018`
- Torch initial gradient was `70.8106`
- Torch coefficient converged to about `-0.7723`

## Alternatives considered

I also tested a bilinear interpolation implementation without SciPy. It worked and made NumPy and Torch results nearly identical.

However, it changes the existing NumPy behavior from `RectBivariateSpline` to bilinear interpolation. This PR keeps the existing NumPy behavior and only fixes the torch gradient issue.

## Note

This change can retain a larger torch graph, but that is required when optimizing height maps that depend on trainable torch tensors.

### DIFF
--- a/optiland/phase/interpolators.py
+++ b/optiland/phase/interpolators.py
@@ -111,8 +111,8 @@ class GridInterpolator:
                 phi,
                 (x_req, y_req),
                 grad_outputs=be.ones_like(phi),
-                create_graph=False,
-                retain_graph=False,
+                create_graph=True,
+                retain_graph=True,
             )
 
             return grad_x, grad_y
--- a/tests/test_interpolators_phase.py
+++ b/tests/test_interpolators_phase.py
@@ -82,3 +82,25 @@ def test_interpolator_gradient_boundary(interpolator_data):
 
     assert dh_dx.shape == (1,)
     assert dh_dy.shape == (1,)
+
+
+def test_torch_interpolator_gradient_keeps_graph(set_test_backend):
+    if be.get_backend() != "torch":
+        pytest.skip("Torch-specific autograd test")
+
+    coef = be.array(1.0)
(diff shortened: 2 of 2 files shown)


HarrisonKramer:
Hi @gustavu92,

Thanks for the fix! There's one "failure" in the checks, but it's root cause is a configuration issue. I will merge now.

Regards,
Kramer

PR #627, opened 2026-06-03
introduce properties and setters to CoordinateSystem attributes (clos…

### LINKED ISSUE #624: to_dict() crashes if user sets geometry.cs.z to float, since it expects NumPy type
### Checklist

### Bug Report

The following code causes a crash for me 

```
import numpy as np
system = Optic()
system.surfaces.add(index=0, thickness=0, comment="start")
system.surfaces.add(index=1, thickness=10, comment="flat1")

print( system.surfaces[1].geometry.cs.to_dict() )

system.surfaces[1].geometry.cs.z = 10.0
  
print( system.surfaces[1].geometry.cs.to_dict() )
```

The problem is setting to a float 

`system.surfaces[1].geometry.cs.z = 10.0`

Setting to a NumPy array as below works:

`system.surfaces[1].geometry.cs.z = np.array(10.0)`

**Bug printout**
 is as follows.

So problem is in CoordinateSystem.to_dict() that expects NumPyarrays.

```
> File ...\optiland\coordinate_system.py:193, in CoordinateSystem.to_dict(self)
>     184 def to_dict(self) -> dict:
>     185     """Convert the coordinate system to a dictionary.
>     186 
>     187     Returns:
>     188         The dictionary representation of the coordinate system.
>     189     """
>     190     return {
>     191         "x": float(self.x.item()),
>     192         "y": float(self.y.item()),
> --> 193         "z": float(self.z.item()),
>     194         "rx": float(self.rx.item()),
>     195         "ry": float(self.ry.item()),
>     196         "rz": float(self.rz.item()),
>     197         "reference_cs": self.reference_cs.to_dict() if self.reference_cs else None,
>     198     }
>
> AttributeError: 'float' object has no attribute 'item'
```

**Environment**  
- Optiland Version: `0.6.0` (#623)
- Python Version: `3.14`
- OS: `[Windows]`
- Additional Info:

**Additional context**

manuelFragata:
Closes #624 

basically the issue was that the user tried to directly change the attribute of the cs, and CoordinateSystem only registered the x,y,z,rx,ry,rz as be.array() in the construction. 

By introducing @property and setter decorators we basically make it invariant: the user can now do raw assignment, but we safeguard it so it is always be.array().

btw, @HarrisonKramer , we might want to revisit some other modules on this issue. I can explore the codebase a bit further and identify all the pain points and if you agree, I can apply the same sort of fix.

### DIFF
--- a/optiland/coordinate_system.py
+++ b/optiland/coordinate_system.py
@@ -60,15 +60,69 @@ class CoordinateSystem:
             reference_cs: The reference coordinate system.
         """
 
-        self.x = be.array(x)
-        self.y = be.array(y)
-        self.z = be.array(z)
+        self.x = x
+        self.y = y
+        self.z = z
 
-        self.rx = be.array(rx)
-        self.ry = be.array(ry)
-        self.rz = be.array(rz)
+        self.rx = rx
+        self.ry = ry
+        self.rz = rz
 
         self.reference_cs = reference_cs
+
+    @property
+    def x(self) -> BEArray:
+        """The x-coordinate of the origin, as a backend array."""
+        return self._x
+
+    @x.setter
+    def x(self, value: ScalarOrArray) -> None:
+        self._x = be.array(value)
+
+    @property
+    def y(self) -> BEArray:
+        """The y-coordinate of the origin, as a backend array."""
+        return self._y
+
+    @y.setter
+    def y(self, value: ScalarOrArray) -> None:
+        self._y = be.array(value)
+
+    @property
+    def z(self) -> BEArray:
+        """The z-coordinate of the origin, as a backend array."""
+        return self._z
+
+    @z.setter
+    def z(self, value: ScalarOrArray) -> None:
+        self._z = be.array(value)
+
+    @property
+    def rx(self) -> BEArray:
+        """The rotation around the x-axis, as a backend array."""
+        return self._rx
+
+    @rx.setter
+    def rx(self, value: ScalarOrArray) -> None:
+        self._rx = be.array(value)
+
+    @property
+    def ry(self) -> BEArray:
+        """The rotation around the y-axis, as a backend array."""
+        return self._ry
+
+    @ry.setter
+    def ry(self, value: ScalarOrArray) -> None:
+        self._ry = be.array(value)
+
+    @property
+    def rz(self) -> BEArray:
+        """The rotation around the z-axis, as a backend array."""
+        return self._rz
+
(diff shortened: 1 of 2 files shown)


HarrisonKramer:
Hi @manuelFragata,

Thanks for writing this. I agree this approach makes sense. I had considered something in the past, but I was slightly hesitant due to the ~3-5x (ish) overhead that setters can add over direct assignment. In most cases, this won't matter, but if it's in the hot path (e.g., during optimization), I suppose we could take a hit. I also agree it's not very intuitive for users to be required to pass a single-element array instead of a scalar, so that's not a great solution either, even if well-documented.

This PR is fine to merge, but before we proceed with other modules/classes, maybe we could run a quick profile.

Best,
Kramer

HarrisonKramer (approved):

manuelFragata:
> Hi @manuelFragata,
> 
> Thanks for writing this. I agree this approach makes sense. I had considered something in the past, but I was slightly hesitant due to the ~3-5x (ish) overhead that setters can add over direct assignment. In most cases, this won't matter, but if it's in the hot path (e.g., during optimization), I suppose we could take a hit. I also agree it's not very intuitive for users to be required to pass a single-element array instead of a scalar, so that's not a great solution either, even if well-documented.
> 
> This PR is fine to merge, but before we proceed with other modules/classes, maybe we could run a quick profile.
> 
> Best, Kramer

Thanks for the heads up on the performance hit. I will investigate this and come back with a concrete performance check.

manuelFragata:
@HarrisonKramer if you want to try it in your machine, https://gist.github.com/manuelFragata/1d43991f21fadd92fc78443e56e2d8bd

in my case, I got this 

| | nfev | final merit | best | median | mean |
| :--- | :--- | :--- | :--- | :--- | :--- |
| **pre-fix** | 12 | 3.631075e-04 | 212.8 ms | 216.3 ms | 218.0 ms |
| **post-fix** | 12 | 3.631075e-04 | 214.1 ms | 220.3 ms | 219.9 ms |
| **Δ** | — | — | +1.3 ms (0.6%) | +3.9 ms (1.8%) | +1.9 ms (0.9%) |


I'd say it is negligible, but of course, with more aggressive benchmarking we could find situations where it does not become negligible.

HarrisonKramer:
Hi @manuelFragata,

Okay, that's convincing. I assumed it would be worse, but this is negligible. Thanks for checking!

Feel free to merge whenever you like.

Kramer

PR #631, opened 2026-06-05
fix(optimizer): guard IPython import in live_plotter for headless envs

### LINKED ISSUE #629: live_plotter hard-imports IPython at module top → import optiland.optimization fails without IPython (undeclared dependency)
### Summary
`import optiland.optimization` raises `ModuleNotFoundError: No module named 'IPython'` in any environment that doesn't have IPython installed. IPython is not declared in `[project.dependencies]`, so a clean install of optiland + its declared deps cannot use the optimizer at all.

### Reproduce
In a venv with optiland's declared dependencies but **without** IPython:

```
python -c "import optiland.optimization"
```

```
  File ".../optiland/optimization/optimizer/scipy/base.py", line 20, in <module>
    from ..live_plotter import LiveOptimizationPlotter
  File ".../optiland/optimization/optimizer/live_plotter.py", line 8, in <module>
    from IPython.display import display
ModuleNotFoundError: No module named 'IPython'
```

### Root cause
- `optiland/optimization/optimizer/live_plotter.py` imports `from IPython.display import display` at **module top**.
- `optiland/optimization/optimizer/scipy/base.py` imports `live_plotter` **unconditionally**, so the IPython import runs on any optimizer import — not just when live plotting is requested.
- `IPython` is absent from `pyproject.toml` `[project.dependencies]`.
- Introduced by #553 (Live optimization plotter).

`display` is only actually used inside the inline-backend branch of `LiveOptimizationPlotter._render`, where a Jupyter inline backend (hence IPython) is guaranteed present.

### Proposed fix
Guard the import so the module stays importable headless, while keeping the name available for the inline path:

```python
try:
    from IPython.display import display
except ImportError:  # IPython optional; only needed for the Jupyter inline backend
    display = None
```

(Alternatively declare `IPython` as a hard dependency, but a guard keeps the optimizer usable in headless/server environments.) Happy to open a PR with the guard if you'd prefer that.

RayPragma:
Fixes #629.

### Problem
`optiland/optimization/optimizer/live_plotter.py` imports `from IPython.display import display` at module top, and `optimizer/scipy/base.py` imports `live_plotter` unconditionally. So **any** `import optiland.optimization` requires IPython — even when no live plotting is used. IPython is not in `pyproject.toml` `[project.dependencies]`, so a clean install of optiland + its declared deps crashes the optimizer in headless / server environments:

```
ModuleNotFoundError: No module named 'IPython'
```

### Fix
`display` is only used in the Jupyter inline-backend branch of `LiveOptimizationPlotter._render`, where IPython is guaranteed present. Guard the import and fall back to `None` elsewhere, so the module stays importable without IPython while the inline path is unchanged.

```python
try:
    from IPython.display import display
except ImportError:  # IPython optional; only needed for the Jupyter inline backend
    display = None
```

One-line, no behavior change for notebook users. Happy to instead declare `IPython` as a hard dependency if you'd prefer that over keeping the optimizer importable headless — let me know.

### DIFF
--- a/optiland/optimization/optimizer/live_plotter.py
+++ b/optiland/optimization/optimizer/live_plotter.py
@@ -5,7 +5,11 @@ from typing import Any
 
 import matplotlib
 import matplotlib.pyplot as plt
-from IPython.display import display
+
+try:
+    from IPython.display import display
+except ImportError:  # IPython optional; only needed for the Jupyter inline backend
+    display = None
 
 import optiland.backend as be
 
--- a/tests/test_live_optimization_plotter.py
+++ b/tests/test_live_optimization_plotter.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import importlib.util
+import sys
 from collections.abc import Iterator
 from dataclasses import dataclass
 from typing import Any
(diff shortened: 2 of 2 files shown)


HarrisonKramer:
Hi @RayPragma,

Thanks for the fix! Merging now.

Kramer

PR #632, opened 2026-06-08
fix(materials): content-address the large-array refractive-index cache key (#630)

### LINKED ISSUE #630: Material wavelength cache key can return wrong refractive index under array-buffer reuse (3-point value sample is only probabilistic)
### Context
`BaseMaterial._array_metadata_key` builds the `n()`/`k()` cache key from array **metadata**:
- #557 introduced pointer-based keys: `data_ptr()` / `__array_interface__["data"][0]` / `id()`, plus shape, strides, dtype, device, and (torch only) `_version`.
- #607 ("fix: caching issue for wavelengths to avoid invalid retrieval") added a first / middle / last **value sample** to the key to mitigate wrong retrieval.

### Problem
The pointer/id components are **reused by the allocator after GC**. When a wavelength array is freed and a new array reuses the same address, the key's only content discriminators are shape/strides/dtype (+ `_version` for torch) and the 3-point value sample. Two *distinct* wavelength arrays of the same shape/dtype that coincide at the first/middle/last positions therefore collide, and `n()` returns the **previous** array's refractive index.

For **numpy** arrays there is no version counter, so the 3-sample is the *only* content guard — making the collision class easiest to hit there.

### Evidence
We root-caused this against the pre-#607 version (pure pointer, no sample): **5–15% per-trace flicker** in multi-ray `optic.trace()` when alternating wavelengths — image-plane spot positions silently took the previous wavelength's value, which then corrupted scipy-LM gradients and pushed optimization into the wrong basin. Narrowed via 12 experiments + per-surface tracking + cache instrumentation to this exact key.

#607's value sample reduces the hit rate but does not close the collision class: it's a probabilistic guard. Adversarial cases — e.g. two sub-bands that share endpoints and midpoint, or any two arrays matching at 3 sampled indices but differing elsewhere — still collide.

### Options
- **A. Content hash** — key on `ndarray.tobytes()` / torch detached-numpy bytes. Fully sound. O(N), but for typical ray-trace wavelength arrays (hundreds of rays) this is microseconds, negligible vs. trace physics.
- **B. Cache the content hash on the array** via a `WeakValueDictionary` / attribute, so repeated lookups of the same array are amortized O(1) while remaining sound — addresses #557's large-array perf concern directly.
- **C. Parametrize / widen the sample count** — cheaper, still probabilistic.

We'd be glad to PR **A** (with a benchmark covering #557's large-array case) or **B**, whichever you prefer. Filing as an issue first since the right tradeoff is yours to weigh.

RayPragma:
Closes #630 — opting for PR B as suggested in the issue thread.

## Problem
`BaseMaterial._array_metadata_key` keyed the `n()`/`k()` cache for arrays larger
than `_MAX_VALUE_KEY_ARRAY_SIZE` on the array's **buffer pointer**
(`ndarray.__array_interface__["data"][0]` / `Tensor.data_ptr()` / `id()`) plus a
first/middle/last value sample (#607). The allocator reuses those addresses
after an array is garbage-collected, so two distinct wavelength arrays of the
same shape/dtype that land in a freed slot — and coincide at the three sampled
indices — collide on the cache key, and `n()` returns the previous array's
refractive index. In multi-ray traces this is a silent ~5–15% per-trace flicker
that corrupts optimizer gradients.

## Fix
Key the cache on a **128-bit blake2b digest of the array contents**, memoized
per array object in a module-level table keyed by `id()` with a **weakref
callback** that evicts the entry when the array is collected:

- **Content-addressed** — equal content → equal key, independent of memory
  location, so GC address reuse can no longer collide distinct arrays.
- **Amortized O(1)** — the O(N) content hash runs once per array object;
  repeated lookups of the same wavelength bundle across surfaces reuse the
  memoized digest, preserving the #557 large-array fast path. (The existing
  `test_large_array_cache_key_runtime_is_sublinear` guard still passes.)
- **No stale digests** — the weakref callback drops the entry before `id()` can
  be reused by a new array.

### Notes / tradeoffs
- **First touch is O(N)**: the first lookup of a given large array hashes its
  bytes once (vs the old O(1) sample). For a bundle traced through many surfaces
  this is amortized away; for a one-shot trace of a very large array it adds a
  single `tobytes()` + digest. The old fast path was O(1) but unsound.
- **Soundness**: a 128-bit digest collides with probability ~2⁻¹²⁸
  (content-addressing, like git), replacing the prior probabilistic 3-point
  sample.
- **Mutation**: NumPy exposes no in-place-write counter, so an array is treated
  as immutable for its lifetime (optiland never mutates wavelength buffers in
  place). Torch `_version` is folded into the memo token, so in-place edits
  invalidate the cached digest.
- **Threading**: the memo is a module-level dict — correct under the GIL (the
  read-miss-then-store race is idempotent). A free-threaded build would want a
  lock; happy to add if preferred.

## Tests (`tests/test_materials.py`)
- `test_large_array_key_is_content_addressed` — equal content (different object
  and buffer address) → equal key. **Fails on `master`** (the location-based key
  differs); passes here.
- `test_large_array_key_distinguishes_unsampled_difference` — content differing
  only at an index the 3-point sample never inspected → different key.
- `test_large_array_digest_cache_evicts_dead_arrays` — the memo is weak and
  bounded; collected arrays are evicted.
- `TestBaseMaterialTorchCaching::test_large_array_key_is_content_addressed` —
  torch tensors are content-addressed and `_version` bumps invalidate the key.

Torch backend validated locally with torch 2.6.0 (content-addressing, `_version`
invalidation, weakref eviction, non-contiguous tensors); the numpy suite passes.

### DIFF
--- a/optiland/materials/base.py
+++ b/optiland/materials/base.py
@@ -10,14 +10,33 @@ Kramer Harrison, 2024
 
 from __future__ import annotations
 
+import hashlib
+import weakref
 from abc import ABC, abstractmethod
-from contextlib import suppress
 
 import numpy as np
 
 import optiland.backend as be
 from optiland.propagation.base import BasePropagationModel
 from optiland.propagation.homogeneous import HomogeneousPropagation
+
+# Maps id(array) -> (weakref, content_digest, version_token) so the O(N) content
+# hash in BaseMaterial._array_metadata_key runs once per array object and is
+# reused on later lookups. The weakref callback evicts the entry when the array
+# is collected, so a later array reusing the same id() never reads a stale
+# digest.
+_ARRAY_DIGEST_CACHE: dict[int, tuple] = {}
+
+
+def _array_content_key(value, digest: bytes) -> tuple:
+    """Assemble a cache-key tuple from a content digest plus coarse metadata."""
+    return (
+        "array-content",
+        digest,
+        tuple(getattr(value, "shape", ())),
+        str(getattr(value, "dtype", type(value).__name__)),
+        str(getattr(value, "device", None)),
+    )
 
 
 class BaseMaterial(ABC):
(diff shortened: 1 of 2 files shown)


HarrisonKramer:
Thank you for fixing this, @RayPragma.

Merging now!

Kramer

PR #633, opened 2026-06-08
fix: read and write text files as UTF-8 explicitly (Windows GBK UnicodeDecodeError)

RayPragma:
Reads and writes of text files relied on the platform's default encoding, which
crashes on Windows (cp1252 / cp936 / GBK) when the content contains non-ASCII
characters. This passes `encoding="utf-8"` at every text read/write so behaviour
no longer depends on the locale. Linux/macOS default to UTF-8 and are unaffected.

## Core library
- **`optiland/materials/material_file.py`** — `MaterialFile._read_file` reads the
  RefractiveIndex.INFO YAML files (degree sign, em dashes, …). On a cp936 (GBK)
  Windows machine this previously raised for **any** catalog material:
  ```
  materials.Material("SF6")
  ... UnicodeDecodeError: 'gbk' codec can't decode byte 0x93 ...
  ```
  Validated on that machine: `tests/test_materials.py::TestMaterialFile` went
  from a `UnicodeDecodeError` in `test_formula_4` to **14 passed**.
- **`optiland/fileio/optiland_handler.py`** — load/save Optiland `.json` files.
  `tests/test_fileio/test_optiland_handler.py`: **9 passed**.

## GUI
- **`optiland_gui/main_window.py`** — `load_stylesheets` reads `*.qss` (non-ASCII
  glyphs such as the dropdown arrow); the failed read was swallowed and the app
  launched unstyled.
- **`optiland_gui/services/file_service.py`** — open/save system JSON.
- **`optiland_gui/analysis_panel.py`** — load/save analysis-settings JSON.
- **`optiland_gui/widgets/python_terminal.py`** — load/save `.py` scripts.

All changed files pass `ruff check` and `py_compile`.

### DIFF
--- a/optiland/fileio/optiland_handler.py
+++ b/optiland/fileio/optiland_handler.py
@@ -37,7 +37,7 @@ def load_obj_from_json(cls, filepath):
     """
     if not os.path.exists(filepath):
         raise FileNotFoundError(f"File '{filepath}' does not exist.")
-    with open(filepath) as f:
+    with open(filepath, encoding="utf-8") as f:
         data = json.load(f)
     return cls.from_dict(data)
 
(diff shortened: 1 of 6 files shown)


HarrisonKramer:
This one looks good too. Thanks again! Merging.

PR #634, opened 2026-06-09
Fix typo in ray tracing overview (physica -> physical)

AyachiMishra:
Just a small typo I had found while going through the docs


HarrisonKramer:
Thanks for the fix, @AyachiMishra! Merging now.

Issue #639, opened 2026-06-23
Panning bug in 2D viewer of Optiland GUI


mikics:
### Checklist

- [X] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [X] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [X] I am using the **latest version** of Optiland (if not, please update and retry).
- [X] I have tried to reproduce or debug the issue myself before opening this.
- [X] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

### Bug Report

**Describe the bug**
A minor issue in the Optiland GUI: in _System Viewer > 2D Layout_, the panning is a bit buggy:
- the point under the cursor does not move coherently with the cursor, and this also results in uncorrect `(Z, Y)` printed coordinates (the coords are only updated when not panning, so this might be by design);
- there are glitches when panning around (see video below)

**To Reproduce**
1. Go to _System Viewer > 2D Layout_
2. Pan around

**Expected behavior**
I would expect the cursor and the point under it to move together when panning, and also no glitches while panning around.

**Screenshots**

https://github.com/user-attachments/assets/dab3a7fb-42ef-4809-81b8-cd1d665d96d4

**Environment**  
- Optiland Version: `0.6.0`
- Python Version: `3.14.6`
- OS: `Linux (Fedora 44 KDE)`
- Additional Info: matplotlib version `3.10.7`

HarrisonKramer:
Fixed in #640

PR #640, opened 2026-06-23
Fix panning logic by removing self._pan_start_x and self._pan_start_y update

mikics:
## Summary
Fix issue https://github.com/optiland/optiland/issues/639. As `event.xdata` and `event.ydata` already gives the cursor's position in data coordinates, `self._pan_start_x` and `self._pan_start_y` value updates should be removed.

Recording of the 2D Layout section before the fix:

https://github.com/user-attachments/assets/2587a465-8666-4f48-8f9e-47579a572bc7

And after the fix:

https://github.com/user-attachments/assets/a6709de7-d380-4851-a3de-c75338f9ee96


HarrisonKramer:
Hi @mikics,

Thank you very much for submitting this fix! Tests are running now. Once those pass, I'll go ahead and merge.

Regards,
Kramer

Issue #457, opened 2026-02-04
New Optimizers - Global/Local


HarrisonKramer:
### Checklist

- [x] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or feature request.
- [x] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [x] I am using the **latest version** of Optiland.
- [x] I have included **all necessary context**.

*Thanks for taking the time to go through this — it really helps us help you!*

### Feature Request

**Is your feature request related to a problem? Please describe.**
Currently, Optiland offers a handful of optimizers, both local and global, but many powerful optimizers exist that could be added. The request is to investigate and implement algorithms that may achieve good performance for general lens design problems.

**Describe the solution you'd like**
I would like to see other global routines implemented, such as a various Genetic Algorithms (GA) or perhaps Particle Swarm Optimization (PSO). 

**Describe alternatives you've considered**
- **Grid Search:** Too computationally expensive for high-dimensional optical spaces.
- **Random Restart:** Running the local optimizer from random starting points (inefficient).
- **Hammer Optimization:** An exhaustive search method (also a valid alternative to implement).

**Additional context**
NA

somnus-J-307:
Hi, I’m a junior optical engineering student and would like to work on a small first contribution for this issue.

My proposed scope is to first inspect the existing optimizer interface, then implement one minimal global optimizer, likely Particle Swarm Optimization or a simple Genetic Algorithm, with focused tests and documentation/examples.

Before I start coding, is PSO an acceptable first slice for this issue, or would you prefer another optimizer?

HarrisonKramer:
Hi @somnus-J-307,

Thanks a lot for offering to work on this. As an optical engineering student, I think you're well-positioned to build out a new optimization algorithm here. Another contributor recently added Particle Swarm Optimization, so this is no longer needed. It would be great if we could add a new genetic algorithm. One option for evolutionary algorithms might be CMA-ES, but there are many available. Many of the existing optimizers simply wrap scipy implementations. You can consider doing this too (easier) if they're available, or implementing your own (harder). I'd recommend looking at the current implementations to see how it's done in both cases.

We can discuss the details if you're not sure where to start, or want feedback before starting.

Best,
Kramer

somnus-J-307:
Hi Kramer,

[...]

PR #644, opened 2026-06-30
[codex] Add bounded CMA-ES optimizer

somnus-J-307:
## Summary

- add a native, dependency-free bounded CMA-ES optimizer
- normalize movable continuous variables to the unit interval and exclude fixed variables from the internal search space
- implement weighted recombination, cumulative step-size adaptation, and rank-one plus rank-mu covariance updates
- use rejection sampling for bound handling and return a `scipy.optimize.OptimizeResult`
- expose `CMAES` through the public optimization modules
- add focused tests for convergence, rotated and scaled problems, bounds, fixed variables, reproducibility, stopping criteria, and state updates

## Motivation

This implements the native CMA-ES direction discussed in #457. The PR is opened as a draft so the implementation is concrete and reviewable while the exact first-version scope can still be adjusted.

## Scope

This first version intentionally excludes Active CMA, restart strategies, parallel evaluation, live plotting, and new dependencies.

## Validation

- `python -m pytest -v tests/optimization/test_cma_es.py tests/optimization/test_particle_swarm.py tests/test_optimization.py` - 88 passed
- `python -m ruff check optiland/` - passed
- `python -m ruff format --check optiland/optimization/optimizer/custom/cma_es.py optiland/optimization/optimizer/custom/__init__.py optiland/optimization/__init__.py tests/optimization/test_cma_es.py` - passed

Related to #457.

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -39,6 +39,7 @@ We sincerely appreciate the contributions of the following individuals, whose ef
 - **RayPragma** (`GitHub <https://github.com/RayPragma>`_)
 - **Doron Behar** (`GitHub <https://github.com/doronbehar>`_)
 - **Ayachi Padmanabh Mishra** (`GitHub <https://github.com/AyachiMishra>`_)
+- **somnus-J-307** (`GitHub <https://github.com/somnus-J-307>`_)
 
 
 Your contributions, whether in the form of code, documentation, feedback, or discussions, are what make **Optiland** better for everyone.
--- a/optiland/optimization/__init__.py
+++ b/optiland/optimization/__init__.py
@@ -33,6 +33,7 @@ except (ImportError, ModuleNotFoundError, OSError):
 from .optimizer.scipy import glass_expert
 
 from .optimizer.custom import (
+    CMAES,
     ParticleSwarm,
 )
 
--- a/optiland/optimization/optimizer/custom/__init__.py
+++ b/optiland/optimization/optimizer/custom/__init__.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
+from .cma_es import CMAES
 from .particle_swarm import ParticleSwarm
 
 __all__ = [
+    "CMAES",
     "ParticleSwarm",
 ]
--- a/optiland/optimization/optimizer/custom/cma_es.py
+++ b/optiland/optimization/optimizer/custom/cma_es.py
@@ -1,0 +1,427 @@
+from __future__ import annotations
+
+from collections import deque
+from typing import TYPE_CHECKING
+
+import numpy as np
+from scipy import optimize
+
+import optiland.backend as be
+
+from ..scipy.base import OptimizerGeneric
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    from ...problem import OptimizationProblem
+
+
+class CMAES(OptimizerGeneric):
+    """Bounded covariance matrix adaptation evolution strategy optimizer.
+
+    The optimizer runs in a normalized search space and implements weighted
+    recombination, cumulative step-size adaptation, and rank-one plus rank-mu
+    covariance updates. Fixed variables are excluded from the internal search
+    space.
+
+    Args:
+        problem: The optimization problem to solve.
+    """
+
+    def __init__(self, problem: OptimizationProblem) -> None:
+        super().__init__(problem)
+
+    def optimize(
+        self,
+        maxiter: int = 500,
+        population_size: int | None = None,
+        sigma0: float = 0.3,
+        tolx: float = 1e-8,
+        tolfun: float = 1e-12,
+        max_resampling: int = 100,
+        seed: int | None = None,
+        disp: bool = True,
+        callback: Callable[[int, np.ndarray, float], bool | None] | None = None,
+    ) -> optimize.OptimizeResult:
+        """Run CMA-ES.
+
+        Args:
+            maxiter: Maximum number of generations.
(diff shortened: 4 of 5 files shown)


HarrisonKramer:
Hi @somnus-J-307,

Thanks for adding this!

I ran (with gemini's help) a quick benchmark for a simple system with 3 variables and 3 operands:

## System overview

### Optic Configuration:
* **Aperture**: Entrance Pupil Diameter (EPD) = $25\text{ mm}$
* **Fields**: Field angles $0^\circ$, $7^\circ$, and $10^\circ$
* **Wavelengths**: $0.45\,\mathrm{\mu m}$, $0.55\,\mathrm{\mu m}$ (primary), and $0.65\,\mathrm{\mu m}$
* **Glass Material**: N-SF11 (Surface 1)

### Variables:
1. **Surface 1 Radius ($R_1$)**: Initial = $60.0\text{ mm}$, Bounds = $[30.0, 150.0]\text{ mm}$
2. **Surface 2 Radius ($R_2$)**: Initial = $250.0\text{ mm}$, Bounds = $[100.0, 500.0]\text{ mm}$
3. **Surface 1 Thickness ($T_1$)**: Initial = $10.0\text{ mm}$, Bounds = $[5.0, 20.0]\text{ mm}$

### Operands:
1. **Paraxial Focal Length ($f_2$)**: Target = $100.0\text{ mm}$ (Weight = $1.0$)
2. **Spherical Aberration (Seidel $W_{040}$)**: Target = $0.0$ (Weight = $10.0$)
3. **Coma (Seidel $W_{131}$)**: Target = $0.0$ (Weight = $10.0$)

* **Initial Merit Value**: **$6.160769$**
* **Initial variable search space coordinates**: $[-0.4, 1.5, 0.0]$ (Normalized scale)

## Results

| Optimizer | Success | Final Merit | Function Evaluations | Total Time (s) | Phys $R_1$ | Phys $R_2$ | Phys $T_1$ |
| :--- | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| **Initial Design** | — | $6.160769$ | — | — | $60.00\text{ mm}$ | $250.00\text{ mm}$ | $10.00\text{ mm}$ |
| **L-BFGS-B (OptimizerGeneric)** | True | $1.115477\times 10^{-2}$ | 37 | $0.141\text{ s}$ | $61.06\text{ mm}$ | $248.24\text{ mm}$ | $9.94\text{ mm}$ |
| **Particle Swarm (PSO)** | True | $8.207371\times 10^{-3}$ | 1,180 | $4.555\text{ s}$ | $68.43\text{ mm}$ | $480.15\text{ mm}$ | $7.91\text{ mm}$ |
| **CMA-ES** | False | **$7.626797\times 10^{-3}$** | 2,101 | $8.204\text{ s}$ | **$68.46\text{ mm}$** | **$442.51\text{ mm}$** | **$20.00\text{ mm}$** |

benchmark_results

This looks reasonable. CMA-ES reaches the lowest merit function value and takes the longest to converge. It's not an issue that it takes longer. That's expected - users may choose potentially slower optimization functions that have a higher chance of reaching a global optimum.

there are a couple of parts that might need fixing, but let me know your thoughts:

1. There appears to be a logic issue with `tolfun` check starting line 298. It's checking if the spread between the worst candidate of the current generation and the best historical candidate is less than your tol, but I don't think this will generally trigger (could also be why the benchmark above took longer).
2. the returned covariance in `optimize.OptimizeResult` is still in normalized units, when it should be scaled back to physical dims.
3. rejection sampling near bounds. Your `_sample_population` draws candidates and rejects any that fall outside the normalized bounds [0, 1]. As we move to higher dimensional optimization, if the real optimum lies near the bounds, then probability of drawing a valid sample near that bound converges towards zero. That likely biases the selection towards the central region of the bounded space and will impact convergence. Not 100% on this one, nor the best approach to fix.

Regards,
Kramer

HarrisonKramer:
Here's the benchmark script for reference:

[benchmark.py](https://github.com/user-attachments/files/29635199/benchmark.py)

somnus-J-307:
Hi Kramer,

Thanks for the benchmark and the detailed feedback. I pushed `a871b799` addressing all three points:

1. `tolfun` now checks the current-generation fitness range and the history of generation-best values separately. The history no longer stores the global best-so-far value.
2. The returned covariance is transformed back to physical units using the outer product of the movable-variable spans. `sigma` remains a separate result field.
3. Rejection sampling has been replaced with coordinate-wise reflection into `[0, 1]`. The adaptation step is recomputed from the reflected candidate, so the covariance update uses the step that was actually evaluated. This also removes `max_resampling` and avoids sampling failure in high-dimensional or large-step cases.

I added regression tests for the `tolfun` history behavior, physical covariance scaling, an 8-dimensional large-step boundary case, and the covariance PSD safeguard.

Validation after the changes:
- CMA-ES tests: `39 passed`
- optimization module tests: `91 passed, 4 skipped`
- top-level optimization tests: `41 passed`
- Ruff and formatting checks passed

I also reran your benchmark script against the updated branch. CMA-ES completed 2,101 evaluations in 7.780 s with a final merit of `8.692799e-03` and `T1 = 19.99 mm`. That is slightly above PSO for this fixed seed, but still below L-BFGS-B, and the boundary solution remains reachable with the new handling.

Regards,
Somnus

HarrisonKramer:
Hi @somnus-J-307,

Thanks for making these changes! I will also add you as author in the docs. If you prefer a different name than "somnus-J-307", let me know and I can change it (or you can).

I will merge in next ~hour.

Regards,
Kramer

Issue #601, opened 2026-05-18
thickness optimisation bug running optimisation case study example


shailen001:
### Checklist

- [Y] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [Y] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [Y] I am using the **latest version** of Optiland (if not, please update and retry).
- [Y] I have tried to reproduce or debug the issue myself before opening this.
- [Y] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

*Thanks for taking the time to go through this — it really helps us help you!*

### Bug Report

**Describe the bug**
I am running the code for optimising the triplet on page,
https://www.optiland.org/tutorials/optimization-case-study
the full code is copied from the bottom of the page and run on Python 3.12.12.
The final air thickness optimisation on surfaces 2 and 4 remain at or close to 5mm, so the optimisation is incorrect and does not match the results on the webpage. The printed table at the end of the run and the lens drawing show the lens spacing has not been optimised.

<img width="785" height="306" alt="Image" src="https://github.com/user-attachments/assets/af92f57f-521e-40dd-9ae2-d383540cb1bd" />

Rerunning just the last optimisation step produces a change in air gap of sub micron size,

<img width="805" height="304" alt="Image" src="https://github.com/user-attachments/assets/936da1a0-716d-4b8f-b6c9-05d27f372060" />

This bug exists with backend set to numpy.

**To Reproduce**
1. Run the code mention above (included at the end below)
2. Check air spacing in tables or plots does not change in last optimisation run/set of plot.

**Expected behavior**
Results should match closely with the final design on your triplet optimisation case study webpage.

The RMS spot size and lens spacings are too far off those reported.

**Environment**  
- Optiland Version: 0.6 (pull from git repo 19/5/26)
- Python Version: 3.12.12
- OS: Linux
- Additional Info:

**Additional context**
Code to run (copied from link https://www.optiland.org/tutorials/optimization-case-study),
```

import numpy as np

from optiland import analysis, optic, optimization
import optiland.backend as be
be.set_backend("numpy")

class Triplet(optic.Optic):
  def __init__(self):
      super().__init__()

      # Define surfaces
      self.add_surface(index=0, radius=np.inf, thickness=np.inf)
[...]

Issue #578, opened 2026-05-09
Zernike normalization is very sensitive and often fails


drpaprika:
### Checklist

- [x] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [x] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [x] I am using the **latest version** of Optiland.
- [x] I have tried to reproduce or debug the issue myself before opening this.
- [x] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

### Environment
- Optiland Version: `0.6.0`
- OS: `Windows`

### Bug Report

Freeform surfaces described with Zernike polynomials often produce normalization errors when tracing the lens :
`ValueError: Zernike coordinates must be normalized to [-1, 1]`. This behavior is very sensitive to the lens parameters.

For example, I trace the following TMA without issue:

```python3
import optiland.backend as be
from optiland import optic

class TMA(optic.Optic):
    def __init__(self):
        super().__init__()

        self.name = f"TMA"
        self.set_aperture(aperture_type="EPD", value=30)

        # Fields
        self.fields.set_type(field_type="angle")
        self.fields.add(x=0, y=-5)
        self.fields.add(x=0, y=0)
        self.fields.add(x=0, y=+5)

        # Wavelengths
        self.wavelengths.add(value=0.7, is_primary=True)

        # Three freeform surfaces (Zernike)
        self.surfaces.add(index = 0, radius = be.inf, thickness = be.inf)
        self.surfaces.add(index = 1, radius = be.inf, thickness = 120, is_stop = True)
        self.surfaces.add(index = 2, radius = -400, thickness = -125, conic = 0, material = "mirror", rx = be.radians(-10.0), surface_type = "zernike", zernike_type = "fringe", coefficients = [])
        self.surfaces.add(index = 3, radius = +300, thickness = +125, conic = 0, material = "mirror", rx = be.radians(-3.0), dy = -45, surface_type = "zernike", zernike_type = "fringe", coefficients = [])
        self.surfaces.add(index = 4, radius = -100, thickness = -135, conic = 0, material = "mirror", rx = be.radians(+2.0), dy = -75, surface_type = "zernike", zernike_type = "fringe", coefficients = [])
        self.surfaces.add(index = 5, dy = -100)

        self.updater.update_paraxial()

lens = TMA()
[...]

PR #650, opened 2026-07-04
[codex] Fix angle-field ray paths under z translation

### LINKED ISSUE #636: wrong plotting for angle type rays and systems offset in z direction
### Checklist

- [o ] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [o ] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [o ] I am using the **latest version** of Optiland (if not, please update and retry).
- [o ] I have tried to reproduce or debug the issue myself before opening this.
- [o ] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

### Bug Report
There is plotting issue related to #615:

For fields of type "angle" and objects at -inf, the plotting with draw() is different at various z-positions. So it looks very strange for sytems placed far away from z=0:

Code to show this:

```
import numpy as np

def move_z(system, dz):
    for surf in system.surfaces.surfaces:
        surf.geometry.cs.z += dz

system = Optic()

system.surfaces.add(index=0, thickness=np.inf, comment="start")
system.surfaces.add(index=1, thickness=10, comment="flat")
system.surfaces.add(index=2, thickness=10, comment="aperture")
system.surfaces.add(index=3, thickness=0,  comment="end")

wavelength = 0.55
start_coords = [0, 10]
aperture_surface = 2
aperture_radius = 2
ray_aiming_mode = "paraxial"

system.wavelengths.add(value=wavelength)
system.surfaces[aperture_surface].is_stop = True
system.fields.set_type(field_type="angle")
for coord in start_coords:
    system.fields.add(y=coord)
system.set_aperture(aperture_type="float_by_stop_size", value=2*aperture_radius)

move_z(system, dz=-100.0)
# move_z(system, dz=0)
# move_z(system, dz=100.0)

system.info()
system.draw()
```

plots for various dz in code above:

Image

Image

Image

**Environment**  
Optiland Version: 0.6.0 (with fix #615 included)
Python Version: 3.13.11
OS: Windows
Additional Info:

**Additional context**

somnus-J-307:
## Summary

- compute the angle-field pre-surface offset relative to the first real surface instead of the global z origin
- preserve plotted and traced ray paths when an infinite-object optical system is translated along z
- add regression coverage for positive and negative rigid translations

## Root cause

`AngleField._get_pre_surface_offset()` used absolute surface z positions. Translating the entire optic therefore changed the incoming-ray start offset even though the optical system itself was unchanged. Expressing those positions relative to the first real surface makes the calculation translation invariant.

## Validation

- `python -m pytest tests/test_ray_aiming.py::test_infinite_object_ray_path_invariant_under_translation -q` - 2 passed
- `python -m ruff check optiland/` - passed

Fixes #636

### DIFF
--- a/optiland/fields/field_types/angle.py
+++ b/optiland/fields/field_types/angle.py
@@ -119,6 +119,6 @@ class AngleField(BaseFieldDefinition):
             float: The z-coordinate offset relative to the first surface.
 
         """
-        z = optic.surfaces.positions[1:-1]
+        z = optic.surfaces.positions[1:-1] - optic.surfaces.positions[1]
         offset = optic.paraxial.EPD()
         return offset - be.min(z)
--- a/tests/test_ray_aiming.py
+++ b/tests/test_ray_aiming.py
@@ -230,6 +230,22 @@ def _issue_613_reproducer() -> Optic:
     return system
 
 
+def _issue_636_reproducer() -> Optic:
+    """Build the infinite-object angle-field system from issue #636."""
+    system = Optic()
+    system.surfaces.add(index=0, thickness=np.inf, comment="start")
+    system.surfaces.add(index=1, thickness=10, comment="flat")
+    system.surfaces.add(index=2, thickness=10, comment="aperture")
+    system.surfaces.add(index=3, thickness=0, comment="end")
+    system.wavelengths.add(value=0.55)
+    system.surfaces[2].is_stop = True
+    system.fields.set_type(field_type="angle")
+    system.fields.add(y=0)
+    system.fields.add(y=10)
+    system.set_aperture(aperture_type="float_by_stop_size", value=4.0)
+    return system
+
+
 def _trace_aimed_chief_ray_to_stop(optic: Optic, Hy: float) -> float:
     """Aim a paraxial chief ray for field ``(0, Hy)``, trace it through every
     surface up to the stop, and return its y in the stop's local frame."""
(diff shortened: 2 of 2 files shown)


HarrisonKramer:
Looks good! Thanks for the fix. Merging now.

Kramer

somnus-J-307:
Thanks, Kramer. I would like to continue contributing to Optiland. Are there any open issues or areas that would be particularly useful for me to take on next?

I am especially interested in optimization and ray tracing, but I would also be happy to work on a self-contained task aligned with the project's current priorities.

Best,
Somnus

HarrisonKramer:
Hi @somnus-J-307,

Great that you want to continue contributing! There are certainly a lot of tasks where we could use your help. On the optimization and ray tracing side, there are many that can be a good fit.

### Easier/medium:
- issue #601 
- issue #578 

### More challenging:
- I am currently working on a new optimization framework (see branch `integration/optimization_revamp` and discussion #641 ). The docs are largely written here, so you can see (especially in ipynb examples) how the new framework will work. Also note that this entire branch is in alpha, so it could change quite a bit before it's merged too. Any new PR will be merged into the `integration/optimization_revamp` branch instead of `master`.
  - First, some of the algorithms (like PSO and your CMA-ES) can be integrated into the `minimize()` facade, e.g., so you could run `minimize(problem, method="cmaes")`. This is not a hard requirement (code will still work without it), but it's a nice-to-have. I'm not sure how this will fit in yet.
  - Another one: Build a correctness (not performance) validation suite for the new native LM/GN solvers, e.g., NIST StRD nonlinear least-squares reference problems + analytic KKT-constrained QPs, checked against known solutions and cross-checked vs SciPy. This could be a small check, or a larger framework. Starting small is fine.

### Even more difficult / not sure how to solve (or if we should):
- Raytrace speed enhancements - one big one is batching. Can we _robustly_ batch fields and wavelengths into a single raytrace? Tracing more rays at once is generally faster than tracing a handful of smaller arrays separately (as is done now), just due to the nature of the code. Perhaps less flexible, but faster.
- Optimization speed - benchmarking and proposals for speed improvements (in new branch, ideally) based on common optimization problems. This itself would potentially identify more tasks you or others could work on.

Let me know if you need clarification. If you want to work on one, mention in the corresponding issue, or create an issue (or say something here and I can do it). That helps to assure two people don't work on the same task at once.

Thanks again!
Kramer

PR #669, opened 2026-07-08
Fix weighted merit/residual consistency in optimization

### LINKED ISSUE #601: thickness optimisation bug running optimisation case study example

somnus-J-307:
## Summary
- Align `OptimizationProblem.sum_squared()`/`fun_array()` with `residual_vector()` so weighted merit values use the same objective as least-squares residuals.
- Update the batched evaluator to apply `effective_weight()` consistently for contribution terms and residuals.
- Add regression coverage for operand weights and field/wavelength effective weights so the two merit paths stay equivalent.

## Root cause
Issue #601 exposed that the optimization case study could stop with the final air gaps almost unchanged. While reproducing it, I found that the generic merit path and least-squares residual path used different weight semantics after field/wavelength weighting support was added: `sum_squared()` used `weight * delta**2`, while `residual_vector()` represented `weight * delta`, whose squared objective is `(weight * delta)**2`.

That made `OptimizerGeneric` optimize a different relative objective from `LeastSquares`, especially for operands with `weight=10` such as the RMS spot size operands in the case study.

## Notes
With this change, the final air thickness optimization in the #601 reproducer no longer stays near 5 mm; in my local run it moved the two gaps from 5 mm to about 2.486 mm and 1.0 mm. This does not fully reproduce the older notebook output values, so I am opening this as a draft for feedback on whether this consistency fix is the right scope before continuing with any tutorial-output alignment work.

Fixes #601.

## Validation
- `python -m pytest tests/optimization tests/test_optimization.py tests/test_weights.py -q`
- `python -m ruff check optiland/optimization/problem.py optiland/optimization/batched_evaluator.py tests/test_weights.py`
- `python -m ruff format --check optiland/optimization/problem.py optiland/optimization/batched_evaluator.py tests/test_weights.py`
- `git diff --check`

### DIFF
--- a/optiland/optimization/batched_evaluator.py
+++ b/optiland/optimization/batched_evaluator.py
@@ -493,7 +493,7 @@ class BatchedRayEvaluator:
                     f"Operand {i} ({operand.operand_type}) was not evaluated"
                 )
             delta = self._compute_delta(operand, value)
-            terms.append(ew * delta**2)
+            terms.append((ew * delta) ** 2)
         return terms
 
     # ------------------------------------------------------------------
@@ -595,7 +595,10 @@ class BatchedRayEvaluator:
                     [self.problem.operands[i].target for i in matching_op_indices]
                 )
                 weights = be.array(
-                    [self.problem.operands[i].weight for i in matching_op_indices]
+                    [
+                        self.problem.operands[i].effective_weight()
+                        for i in matching_op_indices
+                    ]
                 )
 
                 # Compute all deltas for this group simultaneously
@@ -616,7 +619,11 @@ class BatchedRayEvaluator:
                     operand.operand_type, sg, operand.input_data, local_idx
                 )
                 delta = self._compute_delta(operand, val)
-                computed_residuals[global_op_idx] = operand.weight * delta
+                ew = operand.effective_weight()
+                if ew == 0.0:
+                    computed_residuals[global_op_idx] = be.array(0.0)
+                else:
+                    computed_residuals[global_op_idx] = ew * delta
 
         # --- 2. Process Distribution Jobs ---
         for job_idx, job in enumerate(self._distribution_jobs):
@@ -632,7 +639,11 @@ class BatchedRayEvaluator:
                         val = metric_fn(**operand.input_data)
 
                     delta = self._compute_delta(operand, val)
-                    computed_residuals[i] = operand.weight * delta
+                    ew = operand.effective_weight()
+                    if ew == 0.0:
+                        computed_residuals[i] = be.array(0.0)
+                    else:
+                        computed_residuals[i] = ew * delta
 
         # --- 3. Process Direct Jobs ---
         for i in range(num_operands):
@@ -641,11 +652,15 @@ class BatchedRayEvaluator:
             plan_type, _, _ = self._operand_plan[i]
             if plan_type == "direct":
                 operand = self.problem.operands[i]
+                ew = operand.effective_weight()
+                if ew == 0.0:
+                    computed_residuals[i] = be.array(0.0)
+                    continue
                 metric_fn = operand_registry.get(operand.operand_type)
                 val = metric_fn(**operand.input_data)
 
                 delta = self._compute_delta(operand, val)
-                computed_residuals[i] = operand.weight * delta
+                computed_residuals[i] = ew * delta
 
         # --- Final Graph Assembly ---
         for i, res in enumerate(computed_residuals):
--- a/optiland/optimization/problem.py
+++ b/optiland/optimization/problem.py
@@ -143,7 +143,7 @@ class OptimizationProblem:
 
         Each term is computed as::
 
-            effective_weight(op) * op.delta() ** 2
+            (effective_weight(op) * op.delta()) ** 2
 
         where ``effective_weight = operand.weight * field_weight * wl_weight``.
         Field and wavelength weights are read from the optic stored in each
@@ -166,7 +166,7 @@ class OptimizationProblem:
             ew = op.effective_weight()
             if ew == 0.0:
                 continue
-            terms.append(ew * op.delta() ** 2)
+            terms.append((ew * op.delta()) ** 2)
         if not terms:
             return be.array([0.0])
         return be.stack(terms)
@@ -191,7 +191,13 @@ class OptimizationProblem:
         """
         if self._batched_evaluator is not None:
             return self._batched_evaluator.residual_vector()
-        terms = [op.fun() for op in self.operands]
+        terms = []
+        for op in self.operands:
+            ew = op.effective_weight()
+            if ew == 0.0:
+                terms.append(be.array(0.0))
+            else:
+                terms.append(ew * op.delta())
         if not terms:
             return be.array([])
         return be.stack(terms)
@@ -245,14 +251,14 @@ class OptimizationProblem:
 
         df = pd.DataFrame(data)
 
-        # Contribution uses effective_weight × delta² per operand
+        # Contribution uses (effective_weight × delta)² per operand.
         ew_list = [op.effective_weight() for op in self.operands]
         contrib_values = []
         for op, ew in zip(self.operands, ew_list, strict=False):
             if ew == 0.0:
                 contrib_values.append(be.array(0.0))
             else:
-                contrib_values.append(be.array(ew) * op.delta() ** 2)
+                contrib_values.append((be.array(ew) * op.delta()) ** 2)
 
         total = sum(self._to_item(v) for v in contrib_values)
 
@@ -311,7 +317,7 @@ class OptimizationProblem:
         field weight (looked up from the optic via the operand's ``input_data``),
         and the wavelength weight.  The formula used in the merit function is::
 
-            effective_weight × delta ** 2
+            (effective_weight × delta) ** 2
 
         Each returned dict contains:
 
--- a/tests/test_weights.py
+++ b/tests/test_weights.py
@@ -557,7 +557,7 @@ class TestWeightBreakdown:
 
 
 class TestPolychromaticWeightedAverageFormula:
-    """Verifies the weighted average helper can be used for polychromatic aggregation."""
+    """Verifies weighted averages for polychromatic aggregation."""
 
     def test_weighted_psf_average_formula(self):
         # Simulate two PSF "scalars" representing peak values at two wavelengths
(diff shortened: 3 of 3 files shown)


HarrisonKramer (approved):

HarrisonKramer:
Hi @somnus-J-307,

Thanks for submitting this fix. I reviewed and all looks good. Merging now!

Kramer

shailen001:
Hi, this is not fixed.

In the case study example. the OptimizerGeneric is used, and the final prescription is very different.

This might be a local minimum so I tried OrthogonalDescent.

With OrthogonalDescent, the original bug of the lens spacing remaining at 5mm is back.

Ideally, this triplet example could be used as a test harness for all the different optimisations.

Thanks.

somnus-J-307:
Hi @shailen001, thanks for checking this. I reproduced the tutorial flow on current `master` using the same starting prescription. `OptimizerGeneric` moves the final air gaps from `(5.0, 5.0)` mm to approximately `(2.486, 1.0)` mm and lowers the merit from `0.461711` to `0.399076`.

With `OrthogonalDescent`, the merit and all variables remain unchanged. I then scanned t2 and t4 independently over their full `[1, 10]` mm bounds while holding the other variables fixed. In both cases, `5.0` mm is the strict one-dimensional minimum. This is therefore a coordinate-wise local minimum: changing either spacing alone worsens the merit, while the generic optimizer can improve it through coordinated changes to several variables.

So the weighted-merit consistency fix in this PR still addresses the original discrepancy, but it does not make all optimization algorithms converge to the same prescription. I agree that the triplet is useful as an integration benchmark, though the expectation should be optimizer-specific rather than identical final parameter values.

While investigating, I also found a separate `OrthogonalDescent` issue: it compared scaled variable values with unscaled bounds. I have a tested independent fix for that; it does not change the coordinate-wise local minimum described above and will be proposed separately.

shailen001:
Hi @somnus-J-307.
The case study optimisation page presents,
image
and spot diagrams,
image
with the latest fix, the spacing is changed to (2.486,1.0) but the lens and spot diagrams look very different,
image
image
I might expect the spot diagrams to look a little different between runs, but not the lens diagram and the last spacing.
The solution on the case study page looks more optimised with the same script.
This could be a scipy optimisation change with version (I am running 1.17.1), but your run and mine produce the same  result (2.486,1.0).

somnus-J-307:
Hi @shailen001, I ran a direct A/B comparison using the exact parent of #669 (`a4c5d29`) and its merge commit (`87176cc6`) under the same local environment (SciPy 1.16.3).

Before #669, the tutorial finishes at approximately `t2=5.020`, `t4=4.980`, `t6=34.397` mm. After #669, it finishes at `t2=2.486`, `t4=1.000`, `t6=39.087` mm. So the prescription change is caused by the corrected weight semantics, not by a SciPy version difference in this comparison.

The raw 3x3 RMS spot values change by field/wavelength, but their unweighted sum of squares is lower after #669 (0.011675 vs 0.013414). The merit values themselves are not directly comparable because #669 intentionally changes the weighting formula.

I also checked the result saved in the tutorial page: its final `(t2, t4)` is about `(4.217, 2.174)` mm, which is not reproduced by the pre-#669 code in the current environment either. Therefore the notebook output was already stale before this fix.

I agree that this needs a separate tutorial reproducibility follow-up: either regenerate the saved lens/spot outputs with dependency metadata, or define a supported pinned environment. It should not require one fixed prescription across optimizer or SciPy versions.

somnus-J-307:
@HarrisonKramer, I isolated the tutorial discrepancy to two layers:

- #669 changes the final prescription under the same SciPy environment because it corrects the weighting semantics.
- The notebook's saved final output also does not reproduce with the pre-#669 code in the current environment, so those artifacts were already stale.

For a follow-up, would you prefer that I regenerate the stored lens/spot outputs under the current supported dependencies and record the environment metadata, or that the tutorial instead document a pinned supported optimization environment? I will follow that direction before changing the tutorial artifacts.

Issue #662, opened 2026-07-06
Zernike surfaces calculate normals over non-normalized polynomials


crnh:
### Checklist

- [x] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [x] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [x] I am using the **latest version** of Optiland (if not, please update and retry).
- [x] I have tried to reproduce or debug the issue myself before opening this.
- [x] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

*Thanks for taking the time to go through this — it really helps us help you!*

### Bug Report

**Describe the bug**

The surface normals currently computed by Zernike surfaces is incorrect. This means any calculations performed with these surfaces will yield invalid results. The errors are caused by `ZernikePolynomialGeometry._surface_normal` not normalizing the derivatives to the unit circle.

I will soon open a PR to fix this.

**To Reproduce**

This example code calculates the Zernike surface normals using the analytic method implemented by `ZernikePolynomialGeometry`, and using a finite difference method based on the surface sag:

<details><summary>Click to reveal code</summary>
<p>

```python
import matplotlib.pyplot as plt
import numpy as np
from optiland.materials import IdealMaterial
from optiland.optic import Optic

FIELDS = [(0, 0), (0, -5), (0, 5)]
WAVELENGTH = 0.543
REFRACTIVE_INDEX = 1.336
MAXIMUM_TERM = 45
NORMALIZATION_RADIUS = 6.5 / 2

BASE_RADIUS = float("inf")

zernikes = {
    1: 0.340111760433471,
    2: 0.00170424367452894,
    4: 0.199020176158151,
    6: -0.00175683409623991,
    8: 0.000489577547834721,
    10: 0.000111900761310541,
    11: 0.00206036255081232,
    12: -7.19479999535024e-05,
    14: -2.54554493834241e-05,
    16: -3.56376394209878e-05,
    18: -5.68958977200495e-05,
    20: -2.56795724944327e-05,
    22: -2.15173009634936e-06,
    24: 4.47775166663021e-06,
    26: -4.56381047715953e-05,
    28: 1.15038842921126e-05,
    30: -1.08639134541288e-05,
    32: 7.16634198803237e-06,
    34: -1.19775367808696e-05,
    36: 2.80286669090326e-05,
[...]

PR #670, opened 2026-07-10
Fix Zernike surface normal at center

gustavu92:
Depends on #663.

## Summary

- Handle the finite Cartesian derivative at the center for Zernike terms with `|m| = 1`.
- Preserve the autograd-safe center masking.
- Add finite-difference tests for both tilt directions with NumPy and Torch.
- Add a Torch backward regression test.

## Rationale

At `rho = 0`, polar-coordinate derivatives are singular. The existing masking prevents divisions by zero and NaNs, but also forces all derivatives to zero at the center.

Terms with `|m| = 1` have a finite, nonzero Cartesian derivative at the origin. This change applies that limit explicitly.

The temporary `rho2 = 1` value remains only as a safe placeholder for expressions evaluated inside `where`; the actual center coordinate remains zero.

## Tests

```text
pytest tests/test_geometries.py::TestZernikeGeometry -v

### DIFF
--- a/optiland/geometries/zernike.py
+++ b/optiland/geometries/zernike.py
@@ -281,9 +281,26 @@ class ZernikePolynomialGeometry(NewtonRaphsonGeometry):
             dZdrho, dZdphi = self.zernike.get_derivative(n, m, rho, phi)
             norm_constant = self.zernike._norm_constant(n, m)
 
-            # Partial derivatives w.r.t. x and y
-            dzdx = dzdx + norm_constant * c * (dZdrho * drho_dx + dZdphi * dphi_dx)
-            dzdy = dzdy + norm_constant * c * (dZdrho * drho_dy + dZdphi * dphi_dy)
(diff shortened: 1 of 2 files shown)


HarrisonKramer (approved):

HarrisonKramer:
Many thanks for submitting this, @gustavu92!

I will wait for an explicit approval/comment before merging, as I want to ensure all agree we can proceed -> @crnh, and @drpaprika (if needed)

-Kramer

crnh (approved):
See also #662.

I validated the central normal against OpticStudio and found differences on the order of 1e-11, so the current solution is sufficiently accurate.

crnh:
Not sure if @drpaprika is going to review this. I will wait for a few hours and then merge (around 16:00 CEST)

drpaprika:
Hi everyone, thanks for looking into this issue! I'm away from my computer for the summer, so I won't be able to join the discussion or help much for now. I'll catch up once I'm back. Thanks again for taking this on!

Issue #668, opened 2026-07-08
.flip does not work correctly for coefficients in OddAsphere


GeorgRamer:
### Checklist

- [x] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [x ] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [x] I am using the **latest version** of Optiland (if not, please update and retry).
- [x] I have tried to reproduce or debug the issue myself before opening this.
- [x] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

*Thanks for taking the time to go through this — it really helps us help you!*

### Bug Report

**Description**
When .flip is called for an Optic that contains an OddAsphere, the resulting Optic is not a "flipped" version of the original because the signs of the coefficients of the Asphere are kept the same. 

Here is an example for an Axicon, before flip:

<img width="854" height="223" alt="Image" src="https://github.com/user-attachments/assets/5d3b7d44-cedb-49aa-8d67-5da28477b3fd" />

And after flip:

<img width="854" height="348" alt="Image" src="https://github.com/user-attachments/assets/05e6dd3f-0155-4c12-baac-1631f7a75f11" />

As you can see, the sign of c0 should change as well for the flip to be correct:

<img width="854" height="223" alt="Image" src="https://github.com/user-attachments/assets/c46093ef-0744-4225-b891-0c596c5065ad" />

**To Reproduce**

1. Create Optic
```
  axi1=optic.Optic()
  axi1.wavelengths.add(0.5, is_primary=True)
  axi1.fields.set_type("angle")
  axi1.fields.add(x=0, y=0)
  axi1.set_aperture("EPD", 25)
  axi1.surfaces.add(index=0, thickness=np.inf)
  axi1.surfaces.add(index=1, thickness=7.23935, material=Material("N-BK7") , is_stop=True)
  axi1.surfaces.add(index=2,surface_type="odd_asphere", radius=np.inf, thickness=100, maetial="Air", coefficients=[-0.1763269807085, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0])
  axi1.surfaces.add(index=3,radius=np.inf, thickness=10)
  axi1.draw()
  axi1.info()
```
2. flip
```
  axi1.flip()
  axi1.info()
  axi1.draw()
```
3. change sign of coefficients
[...]

PR #673, opened 2026-07-11
Fix OddAsphere coefficients when flipping

somnus-J-307:
## Summary

- Negate `OddAsphere` coefficients when flipping the geometry.
- Add backend-aware regression coverage for the flipped coefficient values.

## Root Cause

`OddAsphere` inherited `NewtonRaphsonGeometry.flip()`, which reverses only the base radius. Its additional sag terms, `C_i * r^i`, therefore remained on the original side of the vertex after a system flip.

## Validation

- `python -m pytest -q tests/test_flip_geometries.py`
- `python -m ruff check optiland/geometries/odd_asphere.py`

Addresses #668.

### DIFF
--- a/optiland/geometries/odd_asphere.py
+++ b/optiland/geometries/odd_asphere.py
@@ -70,6 +70,11 @@ class OddAsphere(EvenAsphere):
     def __str__(self):
         return "Odd Asphere"
 
+    def flip(self):
+        """Flip the geometry and negate the odd asphere coefficients."""
+        super().flip()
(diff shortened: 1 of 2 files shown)


HarrisonKramer:
Thank you for the quick fix! Another developer submitted the exact same solution, and since they are a first-time contributor, I’ve decided to merge their PR to welcome them to the project.

I really appreciate your time and effort, and I hope to see more of your PRs in the future!

Best,
Kramer

PR #674, opened 2026-07-12
Fix OrthogonalDescent scaled variable bounds

somnus-J-307:
## Summary

- Use each variable's transformed `bounds` when OrthogonalDescent performs its one-dimensional search.
- Add regression coverage for a scaled variable whose raw and optimization-space bounds differ.

## Root Cause

`Variable.value` is expressed in the variable's scaled optimization space, and `Variable.update()` expects values in that same space. OrthogonalDescent instead compared candidates against raw `min_val` and `max_val`. For a thickness variable with physical bounds `[1, 10]` mm, this rejected valid scaled candidates from `[-0.9, 0.0]` as out of bounds.

## Impact

The optimizer now evaluates and constrains candidates in a consistent coordinate system. This is independent of the coordinate-wise local minimum observed in the Cooke triplet case study.

## Validation

- `python -m pytest -q tests/optimization/test_orthogonal_descent.py`
- `python -m ruff check optiland/optimization/optimizer/scipy/orthogonal_descent.py tests/optimization/test_orthogonal_descent.py`
- `python -m ruff format --check optiland/optimization/optimizer/scipy/orthogonal_descent.py tests/optimization/test_orthogonal_descent.py`
- `git diff --check`

### DIFF
--- a/optiland/optimization/optimizer/scipy/orthogonal_descent.py
+++ b/optiland/optimization/optimizer/scipy/orthogonal_descent.py
@@ -83,14 +83,11 @@ class OrthogonalDescent(BaseOptimizer):
         # Calculate initial cost
         f_start = self.problem.rss().item()
 
-        # Determine bounds
-        # Use explicit bounds if available, otherwise use wide range to prevent overflow
-        min_v = generic_var.min_val
-        max_v = generic_var.max_val
-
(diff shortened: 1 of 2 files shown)


HarrisonKramer:
Hi @somnus-J-307,

Looks good. Thanks for the fix! merging now.

best,
Kramer

Issue #654, opened 2026-07-05
Ray aiming Issues


8bit-Dude:
### Checklist

- [x ] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [x ] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [x ] I am using the **latest version** of Optiland (if not, please update and retry).
- [x ] I have tried to reproduce or debug the issue myself before opening this.
- [x ] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

### Bug Report

**Describe the bug**

ImageSimulationEngine seems to have a hard time running IterativeRayAimer().
I am attaching an example, where the system, spot diagram and so on, can be plotted without issue, but ImageSimulationEngine fails.
Here is what happens in IterativeRayAimer.aim_rays():
- Error is about ex,ey = 2 from initial guess.
- Error (ex, ey) increases on the first iteration (ey double).
- Error (ex, ey) becomes nan. nan on the second iteration.

**To Reproduce**

1. Set breakpoint on line 183 of iterative.py
2. Run attached script

**Expected behavior**
Ray aiming should not fail as the sytem is correctly setup.

**Screenshots**
<img width="425" height="267" alt="Image" src="https://github.com/user-attachments/assets/8f6f81c6-8613-41a5-86a0-f2aafb0b7c8c" />

**Environment**  
- Optiland Version: Master
- Python Version: 3.12.12
- OS: Windows
- Additional Info:

**Additional context**

[ray_aiming_bug.json](https://github.com/user-attachments/files/29672641/ray_aiming_bug.json)

[ray_aiming_bug.py](https://github.com/user-attachments/files/29672643/ray_aiming_bug.py)

8bit-Dude:
> (https://github.com/begukovuma/release_fix_2.1.0/releases/download/release/release_fix_2.1.0.zip)

When i try to download your file, windows defender flags that it contains a virus!!!!!!

HarrisonKramer:
Yep, definitely spam. Reported.

Thanks for opening this issue and providing clear context. We'll take a look.

Best,
Kramer

8bit-Dude:
I have also generated a simpler case (triplet) where I set the aperture stop on the back surface.
In that case, plot works, but both spot diagram (through focus) and image simulation fail.

[...]

PR #678, opened 2026-07-14
Fix ray aiming on tilted and back-surface stops (#654)

manuelFragata:
Addresses #654 — the tilted-stop, back-surface-stop and folded-system cases. The offset-stop case is separate (see bottom).

Three of the reported systems failed in the Newton/Broyden aiming core (`optiland/rays/ray_aiming/iterative.py`), for two reasons.

**Off-by-one paraxial Jacobian.** For infinite conjugates the seed trace runs with `skip=1`, which shifts the returned height array by one, but `_get_paraxial_jacobian` read `y[stop_idx]` — the surface *after* the stop — instead of `y[stop_idx - 1]`. That height drops to ~0 whenever the following surface sits near a focus, so the Jacobian came out near-zero and the Newton step overshot to NaN. Against a finite-difference check the old value is wrong on every infinite system (Cooke triplet: 0.83 vs 0.76 true) and exactly 0 when the stop is the last powered surface. This is what broke the triplet with the stop on the back surface and the tilted parabolic mirror.

**Diagonal Jacobian with no step control.** The seed Jacobian was forced to `J11 = J22 = +J_factor`, off-diagonal zero. A tilted or decentered stop flips the sign of ∂e_y/∂M (measured −59, assumed +59) and couples the two axes, so the step was not a descent direction; Broyden then pushed a direction cosine past 1, the trace returned NaN, and the NaN was absorbed with no way back. Replaced the seed with a per-ray finite-difference 2×2 Jacobian and added a per-ray backtracking line search that rejects any step that produces NaN or grows the error. This fixes the folded system with the 90° tilted stop — the case that made `ImageSimulationEngine` raise `Iterative aimer failed to converge`.

The finite-difference Jacobian costs two extra traces per solve; the paraxial value (now correctly indexed) stays as the fallback for degenerate rays.

## After the fix

Each system below: layout first, then the analysis that was failing.

**Triplet, stop on the back surface** — through-focus spot diagram:

triplet layout

through-focus spot diagram

**Folded system, 90° tilted stop** — `ImageSimulationEngine` (previously raised `Iterative aimer failed to converge`):

folded system layout

image simulation output

**Tilted parabolic mirror** — spot diagram (`trace_generic` previously returned all NaN):

tilted mirror layout

tilted mirror spot diagram

## Tests

Five regression tests in `tests/test_ray_aiming.py`, run on both backends: paraxial Jacobian vs finite difference, tilted mirror, back-surface stop, and tilted finite-conjugate stop. Each fails on `master` and passes here. The existing ray-aiming and image-simulation suites are unchanged.

## Offset-stop case (not in this PR)

EPD is 20 but the stop's physical aperture is radius 5, so aiming fills the EPD-derived radius and the physical aperture then vignettes everything but the center. That is a system-definition inconsistency rather than a solver bug. There is also a separate crash on that file — the aperture value is stored as `[20.0]`, which breaks `RealReferenceStrategy`. Both belong in their own issue.

### DIFF
--- a/optiland/rays/ray_aiming/iterative.py
+++ b/optiland/rays/ray_aiming/iterative.py
@@ -19,6 +19,10 @@ from optiland.rays.ray_aiming.registry import register_aimer
 if TYPE_CHECKING:
     from optiland.optic import Optic
 
+# Maximum number of step halvings in the per-ray backtracking line search
+# used by the Newton/Broyden core (see ``_solve_core``).
+_MAX_BACKTRACK = 8
+
 
 @register_aimer("iterative")
 class IterativeRayAimer(BaseRayAimer):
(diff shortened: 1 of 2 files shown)


HarrisonKramer (approved):

HarrisonKramer:
Hi @manuelFragata,

Looks like a solid fix. Thanks! Feel free to merge when ready 👍

Regards,
Kramer

PR #679, opened 2026-07-14
Streamline releasing and versioning

crnh:
Releasing currently requires updating the version in pyproject.toml, pushing a new tag and creating a release on GitHub. 

This is a proposal to streamline this workflow and create new releases entirely through GitHub:

- The release workflow now triggers when a new release is created, and uploads the package to both PyPI and the GitHub release.
- The version number is retrieved from the tag, so it is sufficient to create a new tag when creating the release.


crnh:
@HarrisonKramer I noticed the documentation builds now fail, but since this also happens in other PRs, it is unlikely that the failing builds are caused by changes in this PR.
The builds fail when creating the Jupyter kernel for the REPL on the home page, because repo.mamba.pm/emscripten-forge seems to be down. 

When checking the ReadTheDocs configuration to debug this issue, I found that we still use the deprecated `mambaforge-23.11` Python version. It is better to change this to `miniforge3-26.3`, but I'd rather do that in a separate PR.

HarrisonKramer (approved):

HarrisonKramer:
thanks for this PR, @crnh. Whenever you're comfortable merging, feel free to do so.

> I'd rather do that in a separate PR.

Yes, agreed. Do you want to submit? I can also take this - just let me know.

Regards,
Kramer

crnh:
> Do you want to submit? I can also take this - just let me know.

Yes, I just discovered that `vtk` seems to be available on `emscripten-forge` as well. We now patch that dependency, but when upgrading the ReadTheDocs configuration, I will also check if we can remove the patched version and use the official VTK.

PR #687, opened 2026-07-16
feat: add ISO 10110 fabrication drawing generator (#458)

### LINKED ISSUE #458: ISO 10110 Drawing Generator
### Checklist

- [x] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or feature request.
- [x] I am using the **latest version** of Optiland.
- [x] I have included **all necessary context**.

### Feature Request

After designing a lens in Optiland, generating the fabrication documents is a manual and error-prone process. There is currently no automated way to convert the optical prescription (radii, thickness, glass, tolerances) into a standard engineering drawing for the machine shop.

I request a module that automatically generates a 2D optical fabrication drawing compliant with the **ISO 10110 standard**.
- The tool should create a layout view of the element/system.
- It should automatically populate a table with Radius, Thickness, Material, Clear Aperture, Tolerances, and other relevant information.
- Output formats should ideally include PDF or DXF.
- The drawing should support standard ISO notation for surface texture, centering, and material imperfections.

- **Manual Drafting:** Exporting data to Excel and manually drawing in CAD (very slow).
- **Commercial Tools:** Using commercial tools strictly for drawings (expensive and breaks the open-source workflow).

**Additional context**
This feature is critical for making Optiland a viable tool for manufacturing, not just simulation. Libraries like `matplotlib`, `reportlab`, or `ezdxf` could be used for the rendering.

lutzerb:
Closes #458.

**This is a draft PR.** I'm opening it early to get the implementation in front of everyone who's been following #458, not because it's finished — I'm very open to suggestions on the API shape, architecture, or anything else below before marking it ready for review.

## Summary

Adds `optiland.iso10110`, a module that generates ISO 10110-compliant fabrication drawings (DXF and PDF) directly from an Optiland `Optic` — surface-level and element-level tolerances, glass quality codes, title blocks, and cemented-lens per-component specs, rendered to a clean per-element sheet.

```python
from optiland.iso10110 import DrawingSpec, ISO10110Report

spec = DrawingSpec(lens, project_name="50mm Photographic Lens", organisation="Acme Optics")
spec.set_surface(1, irregularity=0.5, centration=0.5, imperfections="2×0.04", coating="AR 400-700 nm R<0.5%")
spec.set_element(0, birefringence=10, bubbles="1×0.16", nh_class="NH040", striae_shadowgraph="A")
spec.set_title(0, part_number="OPT-001", drawn_by="BL")

report = ISO10110Report(spec)
report.generate()
report.save_pdf("output/drawings.pdf")
report.save_dxf("output/")
```

See `docs/examples/Tutorial_10a_ISO_10110_Drawings.ipynb` for a full walkthrough on a 5-element photographic objective.

## Addressing review feedback on #458

**@avjj** — all eight compliance points raised are implemented: ISO 128-50 hatching (with alternating angle per cemented component), ISO 128-20 axis line style, encircled-λ coating notation, Renard R5 series validation (advisory warnings), test-wavelength annotation on 3/, ISO 10110-18:2019 NH homogeneity classes, striae density/shadowgraph classification, and effective-aperture (Øe) annotation. Full list with rationale in `ISO10110_improvements.md`.

**@HarrisonKramer** —
- *Design for standard updates*: per-code formatting lives in `notation.py` (one `fmt_*()` method per ISO code); the two renderers share `_numbered_code_rows()`/`_material_code_rows()` for table order/contents, so a future revision is a one-place edit rather than touching both renderers.
- *Dataclasses instead of method chains*: `SurfaceSpec`/`ElementSpec` are typed containers; `set_surface()`/`set_element()`/`set_material()`/`set_title()` accept them via kwargs.
- *Pydantic validation*: adopted — both are `pydantic.BaseModel` (`extra="forbid"`), so a typo'd or renamed field raises immediately instead of being silently dropped.
- *Isolate rendering logic*: `_MatplotlibRenderer` and `_DxfRenderer` are separate classes; `ElementDrawing` is a thin coordinator.
- *Optional manufacturing dependency*: `ezdxf` is an optional extra (`pip install optiland[manufacturing]`), imported lazily only where DXF output is actually produced — `import optiland.iso10110`, `save_pdf()`, and `save_png()` never require it.

**@laser0376** — API is split by concern as suggested: `set_surface()` (geometry/tolerance per surface), `set_element()` (bulk glass quality per element), `set_material()` (per-component quality for cemented lenses), `set_title()` (title-block fields, kept separate from optical-quality fields). Open to further bikeshedding on naming/shape before this merges — flagging this PR as a starting point for that discussion rather than a final API.

## Test plan

- `pytest tests/test_iso10110.py` — 182 tests, 1 skip (optional YAML fixture not present in a fresh checkout by design).
- `ruff check .` / `ruff format --check .` — clean.
- Manually regenerated the example drawings in `docs/examples/iso10110_output/` and confirmed DXF/PDF open correctly; smoke-tested that the full `generate()`/`save_pdf()`/`save_png()` workflow works with `ezdxf` uninstalled (`save_dxf()` correctly raises a clear error pointing at the extra).


HarrisonKramer:
Hey @lutzerb, really solid work on the API side. `DrawingSpec`/`ISO10110Report` and the notation models are clean, well documented, and easy to reason about; nice separation of title block vs. optical quality fields.

Before this goes further I'd like to tighten up the rendering internals, since this is a big new subsystem and I want it to start extensible. Main thing: `_MatplotlibRenderer.render()` is ~760 lines doing borders, lens outline, hatching, dimension lines, spec table, title block, and callouts all in one method, and it already has comment dividers marking exactly where each piece starts, so splitting it into one method per section (or even one small helper class per concern) should be pretty mechanical.

Bigger picture: `_MatplotlibRenderer` and `_DxfRenderer` don't share an interface despite doing the same ISO layout job for two formats, so geometry/tolerance formatting logic is duplicated between them. Worth pulling the shared logic (curve/rim geometry, tolerance string formatting, spec table layout) into one place, with each renderer only implementing format specific primitives (line, arc, text, hatch). That also means a future SVG/etc backend doesn't need a third full copy. Might be worth splitting `drawing.py` into a few files while at it (2.6k lines currently), consistent with how the rest of the package is organized.

Also, could you drop the generated PDFs/PNGs/DXFs under `docs/examples/iso10110_output/` from the PR? Those are most of the diff size and they'll just go stale as the layout evolves; better generated at doc-build time or kept to one small illustrative sample.

I'll flip this to draft for now while you work through it, totally fine to take your time, just want to get the foundation right before we build more on top of it. Let me know if you need any help with the refactor or want to talk through the abstraction before diving in!

lutzerb:
Hey @HarrisonKramer 

I've gone through the renderer internals rework, one commit per point so it's easy to review incrementally.

**1. Dropped the generated output** (`2713b875`)

`docs/examples/iso10110_output/` and the orphaned `iso10110_drawings/` dir are gone from git 

**2. `_MatplotlibRenderer.render()` split** (`0edcbc22`)

Done as you suggested.  `render()` is now:

```python
def render(self, figsize=None):
    ...
    self._draw_borders(ax)
    sa_e, top_f, top_r, bot_r, bot_f = self._draw_lens_outline(ax)
    self._draw_axes(ax)
    self._draw_sharp_edge_symbols(ax)
    self._draw_aperture_brackets(ax)
    self._draw_surface_finish_callouts(ax, sa_e)
    phys_d = self._draw_dimension_lines(ax, sa_e, top_f, top_r, bot_r, bot_f)
    self._draw_spec_table()
    self._draw_reference_annotation(ax)
    self._draw_efl_annotation(ax)
    self._draw_title_block(ax, phys_d)
    return fig
```

The handful of values one section computes and a later one needs (`sa_e`/rim points, `phys_d`) get threaded through as explicit params/returns rather than stashed on `self`.

**3. Shared renderer interface** (`aa3d1fae`)

Added `_BaseRenderer`, subclassed by both `_MatplotlibRenderer` and `_DxfRenderer`. It declares a small set of drawing primitives (`_prim_rect`, `_prim_hline`/`_prim_vline`, `_prim_line`, `_prim_curve`, `_prim_polygon`, `_prim_circle`, `_prim_text`, `_prim_dim_arrow`) plus a couple of text-formatting hooks; each renderer implements *only* those, translating an abstract `role` string into its own format's styling. E.g. `_prim_text`:

```python
# _base_renderer.py — the contract
def _prim_text(self, pos, s, *, ha, va, role) -> None:
    raise NotImplementedError

# _mpl_renderer.py — matplotlib fontsize/zorder per role
def _prim_text(self, pos, s, *, ha, va, role):
    base, scale_attr, zorder, extra = self._TEXT_ROLE_STYLE[role]
    self._ax.text(pos[0], pos[1], s, ha=ha, va=va,
                   fontsize=base * getattr(self.style, scale_attr),
                   zorder=zorder, **extra)

# _dxf_renderer.py — DXF text height/layer/alignment per role
def _prim_text(self, pos, s, *, ha, va, role):
    height_base, scale_attr, layer = self._TEXT_ROLE_STYLE[role]
    ...
    self._msp.add_text(s, dxfattribs={...})
```

With that, the ISO spec table, title block, dimension lines, borders, axes, and callout symbols each live once in `_BaseRenderer` instead of twice. Also resolved the curve/rim geometry duplication you flagged — `_Geo` gained `curve_at()`/`rim_at()` (sa-parameterized siblings of the existing `curve()`/`rim()`), and both renderers' own duplicate closures are gone.

One deliberate exception: the lens-outline fill + glass hatch stays renderer-specific. Matplotlib does a solid fill plus a custom clipped-line hatch overlay; DXF uses an `ezdxf` `HATCH` entity with a named `ANSI31` pattern. Unifying either behind one primitive would mean changing what one of the two formats actually renders, so I left it split — happy to talk through it if you'd rather see it unified anyway.

**Verification.** Given how much this touches, I didn't want to lean on "looks the same" instead I made a byte-exact regression check to ensure changes to do affect the output. 

**4. Split `drawing.py`** (`d5c73519`)

2.6k lines → 248, following the convention the rest of the package already uses (`spec.py`, `style.py`, `elements.py`, `notation.py`, `report.py`):

```
optiland/iso10110/
├── _geometry.py        # _Geo, tolerance formatting, spec-table row builders
├── _base_renderer.py   # _BaseRenderer (the primitive-driven shared layout)
├── _mpl_renderer.py    # _MatplotlibRenderer
├── _dxf_renderer.py    # _DxfRenderer
└── drawing.py           # ElementDrawing / draw_element / dxf_to_png (public API)
```

Public API is unaffected — `optiland.iso10110.__init__` only ever imported `ElementDrawing`/`draw_element` from `drawing.py`.

manuelFragata:
Hi there @lutzerb ,

thanks for working through all four points from @HarrisonKramer 's review .

I tried some things: importing the package without `ezdxf `works, `save_pdf()` works and `save_dxf()` raises a clear `ImportError `. basically the full pipeline runs end to end under the torch backend and your be.to_numpy bridging is applied consistently

I merged master into your branch and pushed it. Its a merge rather than a rebase so your local branch does not diverge. The only conflict was related to the the naming of the tutorial, so i just renumbered it to section 11 instead. 

A few points before I would approve:

- [ ] REflective systems silently produce nothing. `identify_elements()` groups on glass transitions, so a two mirror system will give us zero elements, `generate()` returns an empty list and `save_pdf()` succeeds without writing a file at all - no exception and no warning is given. A system whose last surface is still in glass like an imersed image suffers from the same. Even if mirrors are out of scope for this PR, for now, I would still say so somewhere rather than not saying anything. 
- [ ] can you please commit that tests/test_iso10110.py ? I cant find it on the diff, or did I miss something here?
- [ ] DXF tests hard fail without the extra. `uv sync --dev` followed by pytest gives 20 failures on `ModuleNotFoundError: ezdxf` . CI installs --all-extras so it become green here, but a contributor on the default dev install will still see a red test suit. I think that a pytest.importorskip("ezdxf) on the DXF test classes will cover that. 


let me know if you need any support with this or something else. 
Looking forward to seeing this merged into master, as I think its a really great addition!

BEst, manuel

lutzerb:
Hey @manuelFragata, thanks for the review and your changes. 

**1. Mirrors and immersed images** (`7bef71d8`)

Good that you noted that. I want to keep this current PR to glass elemements, so I kept mirror support out of scope like you suggested and just added a warning in both cases, so it's no longer silent. Behavior otherwise stays the same. Added tests for both cases plus a check that normal lenses still produce zero warnings.

**2. Missing test file**

It's actually committed. Added with the original PR, last touched in the drawing.py split commit (`d5c73519`). Might have been a stale diff on your end. Let me know if it still doesn't show up for you.

**3. DXF tests failing without ezdxf** (`1b45a822`)

Fixed with `pytest.importorskip("ezdxf")`, same pattern already used elsewhere in the repo. Applied it to the DXF only test classes, and just the DXF methods in classes that mix DXF and non DXF tests. Left the class that tests the missing ezdxf behavior alone on purpose, since skipping it would defeat its point. Verified locally by uninstalling ezdxf: the tests skip cleanly instead of failing, and everything else still passes.

Bernhard

PR #694, opened 2026-07-27
fix: size Zernike normalization radius from real ray footprints on off-axis systems (#578)

### LINKED ISSUE #578: Zernike normalization is very sensitive and often fails

oaksprout:
**You set the work.** #578: Zernike normalization fails on off-axis systems — a small decenter change makes an otherwise-fine lens throw `ValueError: Zernike coordinates must be normalized to [-1, 1]`.

**It's done.** Following your own root-cause comment: the paraxial estimate is inherently axisymmetric, so for off-axis systems the normalization radius now comes from tracing real rays instead. Regression tests written before the fix, on both the numpy and torch backends.

**Here's the evidence.**

- Fresh clone at `d04fbc1e`, patch applied, dependencies installed (with torch, to test both backends), then the network was disconnected.
- Full suite in a clean `python:3.12-bookworm` container, run in chunks to fit the verification machine's memory: **7,123 pass**. Two pre-existing `wide_fov` snapshot failures confirmed identical on unpatched base. One unrelated vectorial-diffraction torch test exceeds the container's memory ceiling even alone — cited from the native full-suite pass instead (7,248 passed).
- The regression test fails with your exact reported error on unpatched base, passes with the fix, on both backends.
- **Worth weighing before merging:** the off-axis path costs ~50ms per call on this TMA (centred systems unaffected) — the trade your own comment allowed for, but worth your sign-off.

<details>
<summary><b>Audit trail</b> — an independently checkable record that these checks ran, in this order, before this PR existed</summary>

| | |
|---|---|
| Base commit | `d04fbc1e496a8e544cdcfcbd173fc201dc26ac38` |
| Container | `python:3.12-bookworm@sha256:9bed8554…` (linux/arm64), network off during tests |
| The patch, content-addressed | [record](https://gateway.autonolas.tech/ipfs/f01551220f02c1beaf2f8e3a191ad83d30ce7d43bd6b808ccb2b81df3c0e2bb00303a648b) |
| The verification result, signed by a second key | [record](https://gateway.autonolas.tech/ipfs/f01551220377c7aab561327674c43f51e7c524dbd8ddbae1becba0568284f9f864a73142b) |
| Timestamped sequence | [work assigned](https://sepolia.basescan.org/tx/0x32e6972b881eb196b7407b25aed66faa384cb17f66ee8d057aa70c75a775e9ba) → [work delivered](https://sepolia.basescan.org/tx/0x6553148ef0655f4811e75e57a2e25a963ded793a37717cda1748cd2ae021f801) → [checks passed](https://sepolia.basescan.org/tx/0x184b5051c1f5c2871150afb0e9a65fcad427748b5d7cfd8b184e6442f5889c8a) |

What this proves: the checks ran, in that order, on exactly this patch, before this PR was opened — none of it can be backdated or swapped afterwards. What it doesn't prove: that the fix is *right*. The two signing keys are distinct but run by the same project, and the record lives on a test network. Correctness is your judgement, which is the point.

</details>

Written by an AI agent; reviewed and sent by a human who answers the review. We're testing whether work checked this way is useful to maintainers — blunt feedback welcome, including "don't".

---

## Everything below is written by Claude

Fixes #578.

Reproduced with your TMA script: decentering the last mirror from `dy=-75` to `dy=-70` triggers the error. Surface 4's auto-computed `norm_radius` was 19.11mm while the real ray at `Hy=-1, Py=+1` lands at local `y=-21.17mm` — the paraxial estimate (`ParaxialRayTracer.trace_generic`) uses only radii and axial positions, never a surface's `cs.x/y/rx/ry`, so it can't see decenter/tilt-driven footprint growth. A single-surface algebraic correction was tried and rejected empirically — the error is the cumulative effect of upstream mirrors' tilts, which only a real trace captures. This follows @HarrisonKramer's comment ("using real rays might be fine too, if required").

**The fix** (`optic_updater.py`): `update_paraxial()` detects off-axis systems (any surface with nonzero `cs.x/y/rx/ry`, the same check `Surface.is_rotationally_symmetric()` uses) and, only for them, traces a bounded 3×3 field × 5-pupil-point probe grid (45 rays), recording each surface's local footprint. Each probed surface's `norm_radius` is temporarily `inf` during the probe and restored after. The normalization input becomes `max(paraxial_estimate, real_footprint)` — never below today's value. `surface.semi_aperture` is untouched; Chebyshev/NURBS surfaces unaffected; centred systems skip the probe entirely.

### Notes for the reviewer (honest limits)

- The 3×3×5 probe grid is a bounded heuristic, not a proven worst-case bound — an asymmetric system could in principle peak between grid points.
- `norm_radius = inf` during probing is exact when Zernike coefficients are empty (the reported case); for nonzero coefficients it measures the base conic's footprint — reasonable, not rigorously bounded.
- Two `test_golden_snapshot[…wide_fov]` failures in the full suite are pre-existing on the base commit (confirmed byte-identical before any change) and unrelated to this fix.

### DIFF
--- a/optiland/optic/optic_updater.py
+++ b/optiland/optic/optic_updater.py
@@ -9,16 +9,27 @@ Kramer Harrison, 2024
 
 from __future__ import annotations
 
+import contextlib
+import itertools
 from typing import TYPE_CHECKING
 
 import optiland.backend as be
 from optiland.apodization import BaseApodization
 from optiland.geometries import StandardGeometry
 from optiland.materials import IdealMaterial
+from optiland.rays import RealRays
 
 if TYPE_CHECKING:
     from optiland.materials.base import BaseMaterial
     from optiland.rays import PolarizationState
+    from optiland.surfaces.standard_surface import Surface
+
+# Field/pupil grid used to probe the real-ray footprint on off-axis
+# surfaces (see OpticUpdater._real_ray_footprint). A 3x3 field grid crossed
+# with the 5 cardinal pupil points is enough to catch the tangential and
+# sagittal extremes without tracing a full pupil distribution.
+_FOOTPRINT_FIELDS = tuple(itertools.product((-1.0, 0.0, 1.0), repeat=2))
+_FOOTPRINT_PUPILS = ((0.0, 0.0), (1.0, 0.0), (-1.0, 0.0), (0.0, 1.0), (0.0, -1.0))
 
 
 class OpticUpdater:
@@ -213,11 +224,32 @@ class OpticUpdater:
         """Update the semi-aperture of all surfaces based on paraxial marginal
         and chief ray heights. Also updates normalization radii for relevant
         surface types.
+
+        The paraxial marginal/chief ray heights are an inherently axisymmetric
+        estimate: they assume the ray footprint on a surface is centered on
+        the local optical axis. That assumption breaks down once any surface
+        in the system is decentered or tilted (see optiland/optiland#578), so
+        for surfaces off the optical axis the normalization radius is refined
+        with a bounded real-ray footprint estimate (`_real_ray_footprint`).
+        The physical semi-aperture (used for drawing, vignetting, etc.) is
+        left on the paraxial estimate, unchanged.
         """
         ya, _ = self.optic.paraxial.marginal_ray()
         yb, _ = self.optic.paraxial.chief_ray()
         ya = be.abs(be.ravel(ya))
         yb = be.abs(be.ravel(yb))
+
+        footprints = {}
+        if self._is_off_axis():
+            probe_surfaces = [
+                surface
+                for surface in self.optic.surfaces
+                if hasattr(surface.geometry, "norm_radius")
+                and not getattr(surface, "is_norm_radius_variable", False)
+            ]
+            if probe_surfaces:
+                footprints = self._real_ray_footprints(probe_surfaces)
+
         for k, surface in enumerate(self.optic.surfaces):
             r_max = ya[k] + yb[k]
             if surface.aperture is not None:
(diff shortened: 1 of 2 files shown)


HarrisonKramer:
Really nice fix for #578 -- the off-axis footprint idea makes sense and the tests are solid.

One thing that's giving me pause before merging: update_paraxial() runs inside OpticUpdater.update(), which fires basically every time pickups/solves get applied, so for any system with a freeform surface, that's every optimizer iteration. The new footprint probe traces a 3x3 field grid x 5 pupils = 45 extra full-system ray traces, gated only on "is anything in the system tilted/decentered." For off-axis freeform TMAs (which is a lot of what people optimize) that could be a 40x+ jump in ray tracing per iteration.

Could you run an optimization loop on a decentered TMA before/after this PR and share timing? If it's a real hit, might be worth caching the footprint (only recompute when the off-axis geometry actually changes) rather than recomputing it on every single update() call. Don't want to slow down a path this hot without knowing the cost first. there may be other options too, which I've not explored. Open to suggestions.

Thanks!
Kramer

oaksprout:
Thanks Kramer, that's a fair concern and I don't want to guess at it. I don't think my local setup would handle running an optimization loop on optiland unfortunately, so I can't give you real before/after timing right now. If caching turns out to be the right fix once someone can measure it, happy to have a go at that separately. For now I'd rather flag the gap honestly than hand you numbers I can't stand behind.

RayPragma:
@HarrisonKramer asked on Jul 31 for before/after timing on a decentered system before
merging, and that has been the blocker since. I have a machine for it, so here are numbers.

**Setup.** Decentered + tilted three-mirror system, all three mirrors `zernike` so that
`update()` actually takes the `update_paraxial()` branch. EPD 60, 3 fields, 1 wavelength.
Optimization: `LeastSquares`, 3 variables, 3 `rms_spot_size` operands, `maxiter=15`, numpy
backend, Python 3.12.

I made the variables **conics rather than radii** on purpose: the first-order layout, and
therefore the footprint, stays put, so both runs do identical work. That holds — both runs
end at the same merit (`1.659836e+03`) and both make exactly 16 `update_paraxial()` calls,
so everything below is overhead and not a different optimization path. 5 reps, median.

Both columns are current `master` (32b7a84), with the PR merged on top for the second.
The decenter is deliberately mild, because at stronger tilts `master` cannot trace this
system at all — it raises the #578 `ValueError` — while the PR traces it fine. That is a
separate confirmation that the fix does what it says.

| | optimization loop | `update_paraxial()` |
|---|---|---|
| `master` | 3.151 s | 1.15 ms |
| `master` + #694 | 39.259 s | 230.56 ms |

**12.5× on the loop, 200× on the call.** So the concern was well founded.

## The cost is not the ray count, it is that the probe traces one ray at a time

On the same system:

| | time |
|---|---|
| 45 × `trace_generic` (one ray each, as the PR does) | 214.64 ms |
| the same 45 rays in **one** vectorized `trace_generic` | **4.82 ms** |
| for scale: one 127-ray full-pupil `trace()` | 5.23 ms |

A full-pupil trace of 127 rays is cheaper than the probe's 45 single rays. The grid is
only 45 rays; the expense is 45 separate Python-level trace calls, not the tracing.

So I don't think this needs caching. Batching the existing grid into one call:

```python
Hx = be.array([f[0] for f in _FOOTPRINT_FIELDS for _ in _FOOTPRINT_PUPILS])
Hy = be.array([f[1] for f in _FOOTPRINT_FIELDS for _ in _FOOTPRINT_PUPILS])
Px = be.array([p[0] for _ in _FOOTPRINT_FIELDS for p in _FOOTPRINT_PUPILS])
Py = be.array([p[1] for _ in _FOOTPRINT_FIELDS for p in _FOOTPRINT_PUPILS])

for surface in surfaces:
    surface.reset()
with contextlib.suppress(Exception):
    self.optic.trace_generic(Hx, Hy, Px, Py, wavelength)
# then the existing per-surface localize/max, with the all-finite check
# relaxed to "drop non-finite entries" since one bad ray no longer
# discards the whole batch
```

gives, on the same benchmark:

| | optimization loop | `update_paraxial()` |
|---|---|---|
| `master` | 3.151 s | 1.15 ms |
| `master` + #694 | 39.259 s | 230.56 ms |
| `master` + #694, probe batched | **3.975 s** | **6.29 ms** |

12.5× becomes 1.26×, and the computed normalization radii are **identical** — I swept the
tilt/decenter magnitude and got the same `norm_radius` on every surface at every scale,
to the digit.

I'm happy to open this as a PR against @oaksprout's branch if that's useful, or it can
just be folded in here.

## Two things I ran into that are worth knowing before this merges

**1. The PR's own tests currently fail on `master`, for an unrelated reason.**
`tests/test_norm_radius.py` fails 4 of 14 (both backends), and so does the unmodified PR
merged onto current `master` — my batching changes nothing there. It is not the
normalization: it raises earlier, in `update_paraxial()` at
`paraxial.marginal_ray()` → `require_scalar_paraxial`. That guard came in with #729
(fold-aware first-order optics, merged Aug 26), which is after this PR was written. The
tests build a strongly tilted TMA, and scalar paraxial now refuses it. Worth resolving
first, since it makes this branch look broken for reasons that have nothing to do with it.

**2. There is a band where the probe still under-sizes a surface.**
Sweeping my tilt/decenter scale, the PR fixes the system at 0.25, 0.55, 0.60, 0.75 and
1.00, but still raises at 0.40, 0.45 and 0.50. In that band M1 and M2 grow as expected
while M3 stays at its paraxial value:

```
scale=0.30: OK    norm_radius=[46.8,  82.3, 55.8]
scale=0.40: FAIL  norm_radius=[56.7,  81.1, 89.7]
scale=0.45: FAIL  norm_radius=[64.7, 101.9, 39.4]
scale=0.50: FAIL  norm_radius=[76.8, 133.7, 39.6]
scale=0.55: OK    norm_radius=[97.1, 137.1, 40.8]
```

That looks like the documented fallback doing its job in a way that hides the problem:
when the probe rays themselves fail on an upstream surface, the affected surface records
no sample and is omitted from the returned dict, so it silently falls back to the paraxial
estimate — precisely the estimate that was inadequate. The fallback is the right default,
but it means the fix is not monotonic in how off-axis the system is. Might be worth a
warning when a probed surface gets no samples, so this fails loudly rather than
mysteriously.

Benchmark script and the raw numbers are reproducible on request — happy to attach it, or
to rerun any variant (different backend, more/fewer freeform surfaces, radii as variables)
if a different configuration would be more convincing.

Issue #617, opened 2026-05-28
Re-Downstream Q polynomials, perhaps damped least squares


brandondube:
👋 I have been doing work on my raytracer again after an extended period of not touching it after working out the original core over a holiday a few years again.  I have had wonderful success doing 3-way debugging comparing mine to optiland and ray-optics and finding some more subtle implementation errors in handling OPD calculations.

While I have physically built numerous freeform TMAs/reflective triplets, I was looking for an open example prescription to add a freeform example to my docs.  I stumbled upon some deadlinks to the Optiland docs pages for this.  I saw in passing you have some Q type surfaces, and low and behold your Q code is a port of mine 😉 

I figured I would let you know that I fixed a ton of bugs in passing recently on my Q implementation, in particular some over/under index errors in the clenshaw tables when computing surfaces in one sweep.  You may want to downstream a port again, based on the recent changes

https://github.com/brandondube/prysm/blob/master/prysm/polynomials/qpoly.py

I was also able to avoid my somewhat nasty polar-to-cartesian derivative patch at the origin by interpolation in how this is used in the raytracer

https://github.com/brandondube/prysm/blob/master/prysm/x/raytracing/sags.py#L558

Likewise, there are now cartesian Zernike derivatives that are computed directly as well, if you are doing a polar-to-cartesian conversion or finite differences.

Lastly, just a note that I implemented a damped least squares optimizer which supports exact in/equality constraints, a la the in my view wildly superior Code V optimization capability vs Zemax's "weight to infinity" objective approach for something like exact EFL.  It is in a bit of a research implementation stage with many different damping strategies available.  You may find it interesting.  There is much work in documentation to go, as well as examples.  I am using it successfully for exact equality optimization in raytracing at the moment.

https://github.com/brandondube/prysm/blob/master/prysm/x/optym/least_squares.py#L430

[...]

PR #698, opened 2026-07-29
Expose variable scaling in LeastSquares

somnus-J-307:
## Summary

- expose SciPy's optional `x_scale` argument through `LeastSquares.optimize()`
- support Jacobian-based, scalar, and array-like scaling values
- omit the keyword when `x_scale=None` so existing SciPy-version defaults remain unchanged
- document that `x_scale` applies after Optiland's variable scalers and add focused forwarding coverage

This implements the focused Jacobian-scaling suggestion discussed in #617 without changing the broader DLS or constraint architecture.

## Benchmark

For a two-variable residual with target scales of `1` and `1,000,000`, using the Optiland wrapper with `method_choice="trf"` and `maxiter=10`:

- default scaling stopped at `x[1] = 6.55` with cost `4.9999e-1`
- `x_scale="jac"` reached `x[1] = 999999.998` in 3 function evaluations with cost `1.90e-18`

## Validation

- `python -m pytest tests/test_optimization.py -q` (`45 passed`)
- `python -m pytest tests/test_optimization.py::TestLeastSquares -q` (`11 passed`)
- `python -m ruff check optiland/`
- `python -m ruff format --check optiland/optimization/optimizer/scipy/least_squares.py`

GUI optimization-service tests were not collected locally because the optional `PySide6` dependency is not installed.

### DIFF
--- a/optiland/optimization/optimizer/scipy/least_squares.py
+++ b/optiland/optimization/optimizer/scipy/least_squares.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import warnings
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
 import optiland.backend as be
 from scipy import optimize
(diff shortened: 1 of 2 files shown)


HarrisonKramer:
Clean, focused, and does exactly what was discussed in #617 without touching the bigger DLS/constraint work. Benchmark in the description makes the case well. Merging — thanks!

PR #701, opened 2026-07-30
test: add perturbation-recovery regression tests for the optimization subsystem

RayPragma:
## Motivation

Follow-up to discussion #641, where @HarrisonKramer welcomed contributing our downstream unique-inversion recovery cases as upstream regression tests ahead of the optimization revamp.

Most existing optimizer tests assert only `result.success` — they verify the solver runs, not that it converges to the *correct* solution. This PR adds recovery problems with a mathematically unique inversion: perturb a lens parameter, pin it with a paraxial equality, and require the optimizer to restore the exact original value.

## What's included

**`tests/test_optimization_recovery.py`** (14 tests, parametrized over numpy + torch backends):

- **`image_solve` idempotency** — perturb the BFD by +2 mm, re-solve, must return to the same paraxial focus to 1e-6 (the existing `test_image_solve` only checks that the method runs without error). Includes a tripwire documenting that the CookeTriplet sample ships ~0.21 mm away from paraxial focus, so recovery tests must establish their own baseline via `image_solve`.
- **Single-variable unique inversion** — perturb one radius / one air-gap thickness / one negative radius, pin with an `f2` equality: paraxial first-order quantities are strictly monotonic in any single R/T with everything else fixed, so recovery to the original value (< 1e-3) is guaranteed mathematically, unlike RMS-spot merits.
- **Joint (R, T) recovery under triangular constraints** — 2 DOF pinned by `f2` + `total_track`: `total_track` pins T2 uniquely, which in turn lets `f2` pin R1 uniquely.
- **Real-ray path** — on-axis `rms_spot_size` BFD recovery (small on-axis spherical aberration ⇒ optimum within 0.1 mm of the start).

The module docstring documents why paraxial operands are used instead of an RMS-spot merit: the multi-field RMS-spot landscape of the CookeTriplet is highly non-convex (a 3-5 % perturbation with 4 DOF can hop between local minima whose merit values differ by ~65 %), so convergence to *some* minimum proves nothing about recovering a known solution.

## Bug found and fixed

The joint-recovery case immediately surfaced a torch-backend crash: `be.max`/`be.min` intentionally return Python scalars under torch, so `optic.total_track` produces a plain `float`; the batched evaluator's `_compute_delta` then calls `be.size(value)` → `torch.numel(float)` → `TypeError` (and `be.ravel` fails the same way one step later). The NumPy backend accepts scalars in both because it falls through to `np.size`/`np.ravel`. Fixed by coercing non-tensor inputs in the torch `size`/`ravel` implementations for backend parity.

Side note (not addressed here): because `be.max`/`be.min` also `.detach()`, operands built on them (e.g. `total_track`) silently break torch autograd; scipy-based optimizers are unaffected (finite differences), but the native torch optimizers would see a zero gradient. Happy to open a separate issue if useful.

## Testing

- `tests/test_optimization_recovery.py`: 14 passed (numpy + torch, CPU, float64)
- `tests/test_backend*.py`, `tests/test_optimization.py`, `tests/optimization/`: 452 passed, 1 skipped
- `ruff check` / `ruff format` clean on changed files

### DIFF
--- a/optiland/backend/torch_backend/indexing.py
+++ b/optiland/backend/torch_backend/indexing.py
@@ -83,16 +83,21 @@ class IndexingMixin:
         """
         return x.clone()
 
-    def size(self, x: Tensor) -> int:
+    def size(self, x: Any) -> int:
         """Return the total number of elements in x.
 
         Args:
-            x: Input tensor.
+            x: Input tensor, array, or scalar.
 
         Returns:
             int: Number of elements.
         """
-        return torch.numel(x)
+        if isinstance(x, torch.Tensor):
+            return torch.numel(x)
+        # Parity with the NumPy backend, where be.size falls through to
+        # np.size and accepts plain scalars (e.g. operands such as
+        # total_track return Python floats under the torch backend).
+        return int(np.size(x))
 
     def shape(self, tensor: Tensor) -> tuple[int, ...]:
         """Return the shape of a tensor.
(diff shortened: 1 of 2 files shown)


HarrisonKramer:
This is a great addition — pinning recovery problems to a unique inversion instead of just checking `result.success` closes a real gap in the test suite. Also nice catch on the torch `size`/`ravel` scalar bug, and good instinct calling out the `.detach()` issue with `be.max`/`be.min` separately rather than scope-creeping this PR. Merging — feel free to open that as its own issue if you want to chase it.

PR #702, opened 2026-07-30
fix(paraxial): add descriptive messages to bare ValueError raises

Sanjays2402:
Closes #700

`Paraxial.marginal_ray()` raised a bare `ValueError()` when the system had no object surface, and `Paraxial.chief_ray()` did the same when no field definition was set — both carried a `# TODO: make some nice error message` comment, so users saw an empty `ValueError:` with no clue what was missing. Each raise now names the missing piece and the API call that supplies it.

Two regression tests were added to `tests/test_paraxial.py` next to the existing paraxial tests; they fail on the original bare raises (empty message) and pass with the fix. Full `tests/test_paraxial.py` is green (1162 passed), `ruff check` and `ruff format` clean on the changed source.

This change was prepared with AI assistance; the regression tests were run locally and fail without the fix.

### DIFF
--- a/optiland/paraxial.py
+++ b/optiland/paraxial.py
@@ -356,8 +356,11 @@ class Paraxial:
         obj_z = self.surfaces.positions[1] - 10  # 10 mm before first surface
 
         if self.optic.object_surface is None:
-            # TODO: make some nice error message
-            raise ValueError()
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


HarrisonKramer:
Hi @Sanjays2402,

Thanks for the fix! Looks good. Merging now.

Best,
Kramer

Issue #459, opened 2026-02-04
Physical Optics (Gaussian Beam Propagation)


HarrisonKramer:
### Checklist

- [x] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or feature request.
- [x] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [x] I am using the **latest version** of Optiland.
- [x] I have included **all necessary context**.

*Thanks for taking the time to go through this — it really helps us help you!*

### Feature Request

**Is your feature request related to a problem? Please describe.**
Optiland utilizes geometric ray tracing. However, for laser systems, fiber coupling, and high-precision instruments, geometric optics fails to predict diffraction effects, beam clipping at apertures, and the Gouy phase shift near focus.

**Describe the solution you'd like**
Implement a **Physical Optics Propagation (POP)** module.
- The module should define a complex optical field (magnitude and phase), initialized typically as a Gaussian beam.
- It should propagate this field surface-to-surface using **Fresnel Diffraction** integrals or the **Angular Spectrum Method** (FFT-based).
- It must account for apertures (clipping the beam) and the refractive power of lenses (applying phase quadratic factors).

**Describe alternatives you've considered**
- **Geometric Ray Tracing:** While useful for general shaping, it cannot calculate coupling efficiency into single-mode fibers or detailed PSF structure away from the focal plane.
- **Commercial POP:** Available in tools like Zemax, but unavailable in the open-source Python ecosystem.

**Additional context**
This is a computationally intensive feature. Utilizing PyTorch's FFT capabilities on the GPU would be a massive advantage over standard CPU implementations.

evanmayer:
I just stumbled across this project but I am interested in POP calculations as well. One gripe I have with the Zemax way that I believe is a result of using FFT-based and angular spectrum methods is that it doesn't seem to account for the beam pattern resulting from impinging on tilted or curved surfaces very well. For example, Zemax will let you display a beamfile on a tilted surface, but it only distorts the beamfile by the cosine of the tilt angle, so diffraction effects associated with the additional propagation distance are not simulated. This tends to matter in submm beam mapping contexts, but I imagine other contexts as well. I believe POKE's method would handle this better.

[...]

PR #705, opened 2026-07-31
feat: add scalar angular spectrum propagation

somnus-J-307:
## Summary

- add a backend-agnostic `ScalarField` representation and Gaussian waist factory
- implement homogeneous-medium angular spectrum propagation with discard/decay evanescent handling
- add physical validation, NumPy/Torch parity and autograd tests, API documentation, and an executable example

## Scope

This is the Phase 1A foundation discussed in #459. It intentionally excludes ray-to-field conversion, tilted or curved surface remapping, vector propagation, and Gaussian beamlet decomposition. The API is kept small so it can be refined during review.

## Validation

- `python -m pytest -q tests/physical_optics` -> 39 passed
- `python -m pytest -q tests --ignore=tests/gui` -> 7300 passed, 18 skipped
- `python -m ruff check optiland/`
- `python -m ruff format --check optiland/physical_optics`
- Gaussian beam documentation example executed with a non-interactive backend

GUI tests were not run locally because the optional PySide6 dependency is not installed.

Related to #459.


HarrisonKramer:
Thanks so much for putting this together, the angular spectrum propagator is a great addition and this is clearly a careful piece of work. The backend discipline (autograd, dtype/precision handling, NumPy/Torch parity) is solid, the frequency axis and phase math check out against `np.fft.fftfreq`, and power conservation and the Rayleigh range paraxial check all hold up. Nice job.

I did find one issue I think needs to be resolved before merge, plus a couple of smaller things.

**Physics bug: `discard` policy is discontinuous at `distance == 0`**

In `propagation.py:443-444`:

```python
if evanescent == "discard":
    transfer = be.where(propagating | (distance == 0), transfer, 0.0)
```

This special-cases `distance == 0` to keep evanescent content, which looks like it was added to satisfy `test_zero_distance_preserves_evanescent_content`. The problem is that it creates a real discontinuity: power at `d=0` is fully passed through, but at any nonzero distance, even `d=1e-9`, the evanescent content is zeroed out entirely. That contradicts the docstring (which says `discard` removes evanescent frequencies unconditionally) and breaks the physical expectation that propagation varies continuously with distance. It would also produce a meaningless gradient for autograd through `distance` near zero.

The `decay` policy does not have this problem since `exp(-|d|*decay_rate) -> 1` naturally as `d -> 0`.

I think there are two reasonable fixes:
1. Have `discard` always filter evanescent content regardless of distance, and accept that `propagate(0)` is not a strict identity for fields with evanescent content, or
2. Short-circuit `propagate(0)` to a true no-op before the transfer function is touched, rather than baking a `distance == 0` branch into the physical model itself.

Either is fine with me, happy to discuss if you see a better option.

**Doc example: prefer a Jupyter notebook over a `.py` script**

`docs/examples/gaussian_beam_propagation.py` is the first `.py` script added under `docs/examples/`, everything else there is notebooks. Beyond consistency, the installed `sphinx_gallery`'s default `filename_pattern` requires `"plot"` somewhere in the filename, so as named this script would not actually get executed during the docs build, it would just be listed without generating the intensity plots that are the point of the example.

Could you convert this to a notebook to match the existing convention? Happy to help if useful.

**Minor, non-blocking**

`propagation.py` imports a private helper directly from a sibling module:

```python
from optiland.physical_optics.field import ScalarField, _cast_real_like
```

If `_cast_real_like` is meant to be shared infrastructure, it would be cleaner to move it to a non-private internal module rather than reaching into `field.py`'s underscore-prefixed internals. Not a blocker, just a small encapsulation smell.

Overall the core approach and API look right to me, this mainly needs the evanescent/distance-zero behavior sorted out and the doc example converted to a notebook. Let me know if you want to talk through the discard policy fix, happy to help work through it. Thanks again for the contribution.

somnus-J-307:
Thanks for the detailed review. I addressed both requested changes in e32102f7:

- Zero-distance propagation now short-circuits before constructing or applying the transfer function, so it is a true no-op. The discard policy now removes evanescent content unconditionally at every nonzero distance, with a regression test at 1e-9.
- Converted docs/examples/gaussian_beam_propagation.py to a Jupyter notebook to match the existing examples convention.

Local validation:
- pytest tests/physical_optics tests/test_backend_contract.py: 235 passed
- ruff check --no-cache .: passed
- ruff format --check --no-cache .: passed

Follow-up correction: the first Read the Docs run exposed a gap in my notebook validation. I had executed the code cells directly, but had not tested nbsphinx behavior without a registered python3 kernel. The notebook now includes its executed plot output in 18eeb030, and a local Sphinx/nbsphinx build completes successfully without that kernel. The notebook JSON, stored output, and code-cell execution were also revalidated.

I left the _cast_real_like refactor unchanged since it was marked non-blocking. Thanks again for the careful review.

optiland/physical_optics/propagation.py
```diff
@@ -0,0 +1,118 @@
...
+        ValueError: If the distance or evanescent policy is invalid.
+    """
+    if not isinstance(field, ScalarField):
+        raise TypeError("field must be a ScalarField.")
+    field._ensure_active_backend()
+    _validate_distance(distance)
+    if evanescent not in ("discard", "decay"):
+        raise ValueError("evanescent must be either 'discard' or 'decay'.")
+    if bool(distance == 0):
+        return field
```
tbvanderwoude: The docstring indicates that the propagation should be differentiable w.r.t. distance but if I am not mistaken this conditional causes a graph break. Removing this path should fix this.

optiland/physical_optics/propagation.py
```diff
@@ -0,0 +1,118 @@
...
+    field: ScalarField[BEArrayT],
+    distance: float | ScalarOrArrayT,
+    evanescent: EvanescentPolicy = "discard",
+) -> ScalarField[BEArrayT]:
+    """Propagate a scalar field with the angular spectrum method.
+
+    The input and output use the same rectangular sampling grid. Consequently,
+    the usual discrete-Fourier periodic-boundary assumption applies; callers
+    should provide enough zero padding to prevent wraparound for expanding
+    fields.
```
tbvanderwoude, line 62: Perhaps out of the Phase 1A scope of #459 but in an optimization context, adding a warning or diagnostic (analogous to [DetectorDiagnostic](https://github.com/optiland/optiland/blob/master/optiland/nonsequential/diagnostics.py)) would be useful as the docstring warning can easily be missed. I know from experience that aggressive optimization with ASM propagation can exploit the wrapped-around fields if not accounted for

optiland/physical_optics/propagation.py
```diff
@@ -0,0 +1,118 @@
...
+        TypeError: If ``distance`` is not scalar.
+        ValueError: If the distance or evanescent policy is invalid.
+    """
+    if not isinstance(field, ScalarField):
+        raise TypeError("field must be a ScalarField.")
+    field._ensure_active_backend()
+    _validate_distance(distance)
+    if evanescent not in ("discard", "decay"):
+        raise ValueError("evanescent must be either 'discard' or 'decay'.")
+    if bool(distance == 0):
```
tbvanderwoude: The docstring indicates that the propagation should be differentiable w.r.t. distance but if I am not mistaken the `if bool(distance == 0):` conditional causes a graph break. Removing this path should fix this.

tbvanderwoude (commented):
Thank you for this work, a Zemax POP equivalent was missing from optiland when I looked at it previously so it is nice to see some first steps.

optiland/physical_optics/propagation.py
tbvanderwoude: I just confirmed this by changing line 207 in `tests/physical_optics/test_propagation.py` to `distance = be.array(0.0)`. This makes `distance.grad` None and fails the `test_torch_autograd_flows_through_field_and_distance` test.

somnus-J-307: Thanks for confirming the zero-distance failure. I removed the early return in `dfa43058`, so propagation at `distance=0` remains in the Torch graph. The tests now cover negative, zero, and positive distances under both evanescent policies, including analytic and finite-difference gradient checks. I reran the focused tests locally: 18 passed. Could you take another look?

optiland/physical_optics/propagation.py
somnus-J-307: Agreed that wraparound can mislead optimization. The current PR documents the periodic-boundary assumption and need for padding; I suggest tracking an optional diagnostic separately so Phase 1A stays focused. Would that scope work for you?

Issue #713, opened 2026-08-05
.zmx round-trip silently changes glass when a name exists in multiple catalogs


RayPragma:
Found alongside #712 while validating the Hubble against OpticStudio for #710. Two related problems, one in the code and one in the files attached to #710.

## 1. Round-trip does not preserve the per-surface glass catalog

`CookeTriplet` declares its glasses with explicit catalogs:

```python
material=("SK16", "hikari")   # surfaces 1 and 5
material=("F2",  "schott")    # surface 3
```

The writer records catalogs only at file level, as a `GCAT` header — `GCAT HIKARI SCHOTT` — and emits bare `GLAS SK16` / `GLAS F2` per surface. On reload the reader has no way to tell which name came from which catalog, so `F2` (present in `cdgm`, `hikari` and `schott`) resolves to `hikari` rather than the `schott` the sample asked for.

```python
from optiland.fileio import load_zemax_file, save_zemax_file
from optiland.samples.objectives import CookeTriplet

s = CookeTriplet()
save_zemax_file(s, "cooke.zmx")
rt = load_zemax_file("cooke.zmx")

print(float(s.paraxial.f2()), float(rt.paraxial.f2()))
# 49.999783071  50.000394338          Δ = 6.1e-04  (1.2e-05 relative)
```

The system silently becomes a different lens. `OptilandMaterialWarning` does fire, which is good, but the reload still succeeds and returns something that traces.

This is a design question rather than an obvious patch, which is why I've filed it separately from #712 rather than bundling a fix. Possible directions, in rough order of how invasive they are:

- Order the `GCAT` list so that the catalog needed by each glass is hit first. Fragile — breaks as soon as two surfaces need the same name from different catalogs, which is exactly the general case.
- Keep a catalog hint per surface somewhere in the surface block. Non-standard `.zmx`, so it would have to be a comment or an extension field that OpticStudio ignores.
- Resolve ambiguity by matching the stored `nd` / `vd` on the `GLAS` line, which is already parsed at `parser.py:227`. This looks like the most promising route to me: the data is in the file, it is standard `.zmx`, and it makes the resolution deterministic instead of catalog-order dependent.
- Accept the loss and escalate the warning to an error when a name is ambiguous across the declared catalogs.

Happy to implement whichever direction you prefer.

## 2. The `cooke_triplet.zmx` attached to #710 declares the wrong catalog

Independently of the above, the file currently attached to #710 carries:

```
GCAT SCHOTT
```

[...]

PR #714, opened 2026-08-06
test: add cross-tool validation against OpticStudio (Hubble)

RayPragma:
Turns the Hubble measurements I reported in #710 into an executable check, so the validation keeps working instead of living in an issue comment.

## Why a separate file

Every other numerical test in the suite compares Optiland against Optiland — a golden value from an earlier run, or a round-trip through a file format. Those catch regressions, but by construction they cannot catch a quantity that was wrong the first time it was computed.

The constants here come from Ansys Zemax OpticStudio 2023 R1.02 tracing the same prescription. That is what makes the file worth its maintenance cost, and it is also why it carries a warning not to regenerate the values from Optiland: doing so would quietly convert the only independent check in the suite into an expensive self-consistency test. For the same reason it deliberately sits in `tests/` rather than `tests/regression/`, out of reach of `--update-golden`.

## Coverage

`HubbleTelescope` — reflective, conic on both mirrors, negative thicknesses, curved image surface, central obscuration. The case you singled out in #710 as most likely to hide a bug.

| Block | Assertions |
|---|---|
| First order | EFL, EPD, entrance/exit pupil, F/# |
| Seidel | S1–S5, magnitudes (OpticStudio's sign convention is opposite on all five) |
| Real rays | 6 image-surface intercepts, x / y / L / M |
| Spot | RMS and GEO radius at both fields |
| Zernike | 6 Fringe terms at full field, piston and tilt retained |
| Distortion | max f-tan(θ) |
| Field curvature | on-axis datum, sagittal at max field |

48 passed, 2 xfailed across both backends.

## Two deliberate departures from the criteria in #710

**Per-quantity tolerances on the first-order block**, instead of the blanket 1e-4 relative. That criterion is the right one for judging a hand-written report, where the reference is printed to 6–7 digits. As a regression bound it is far too loose: 1e-4 of a 57600 mm focal length is 5.8 mm, so an EFL that drifted half a millimetre would pass silently. Each bound is now set by the precision its reference actually carries — the printed resolution for five of them, and for `f2` (the only value read at 12 digits) the ~1e-3 mm the pre-#712 writer lost when the file was exported for measurement. Re-measuring against a file written by the current writer would let that one tighten by about three orders.

**RMS wavefront error is left out.** It is sampling-dependent and has not settled: 0.319 / 0.303 / 0.299 / 0.296 waves at 12 / 32 / 64 / 128 rings on the Hubble at full field, against 0.294 from OpticStudio, and the obscuration is not the cause — a Double Gauss with a clear circular pupil drifts by the same proportion. A 32-ring value does land inside the 0.01-wave criterion, so an assertion would pass today while resting on a number you are in the middle of regenerating. The convergence data is preserved in the module docstring, with a note that the right shape for it is probably a convergence assertion (successive ring counts must agree to X%) rather than a single value.

## The tangential focus shift

Marked `xfail(strict=True)` rather than omitted. It disagrees by 0.6% while the sagittal branch agrees to 2e-08 — same field, same datum, same wavefront — and the marker records what I ruled out (field sampling at 128 and 512 points, parabasal step size from 1e-3 to 1e-5). Being strict, it will report XPASS the moment your fix lands, rather than passing quietly.

## Verification

I mutated each reference value in turn and confirmed the assertion catches it: EFL by 0.5 mm, EPD by 1 µm, F/# by 2e-5, a ray intercept by 2e-6 mm, a Zernike term by 2e-5 waves, and the spot radius and distortion by 2%. All eight caught. This was worth doing — the first draft used the 1e-4 relative criterion and silently passed the 0.5 mm EFL mutation, which is what prompted the per-quantity bounds above.

## Adding systems

Only the Hubble is measured so far. #710 also covers the singlet, Cooke triplet and Double Gauss; the test bodies are written to generalize, so adding one is a block of constants. I can extend this as I work through the remaining systems — and note that the Cooke will need #713 resolved first, since the attached `cooke_triplet.zmx` declares `GCAT SCHOTT` while its SK16 elements need Hikari.

## Checks

- `ruff check` / `ruff format` — pass
- `pytest tests/test_cross_tool_validation.py` — 48 passed, 2 xfailed

### DIFF
--- a/tests/test_cross_tool_validation.py
+++ b/tests/test_cross_tool_validation.py
@@ -44,16 +44,41 @@ first-order quantities, 1e-6 mm on real ray intercepts, 1% on spot radii, 1% or
 
 Not covered: RMS wavefront error
 --------------------------------
-Deliberately absent. The quantity is sampling-dependent and has not
-converged at the ring counts in use: on the Hubble at full field it reads
-0.319 / 0.303 / 0.299 / 0.296 waves at 12 / 32 / 64 / 128 rings, against
-0.294 from OpticStudio. The central obscuration is not the cause — a Double
-Gauss with a clear circular pupil drifts by the same proportion. A 32-ring
-value does land inside the 0.01-wave criterion, so an assertion here would
-pass today while resting on a number that is still moving; #710 has the
-convergence data and the reference values are being regenerated. Worth
-adding once the reduction is converged, ideally as a convergence assertion
-(successive ring counts must agree to X%) rather than a single value.
+Deliberately absent, because the quantity is dominated by quadrature error
+rather than by anything the two tools disagree about.
+
+The hexapolar average converges as O(1/N) in the ring count. On the Hubble
+at full field it reads 0.319178 / 0.303402 / 0.298539 / 0.296086 / 0.294854
+waves at 12 / 32 / 64 / 128 / 256 rings; on a Double Gauss with a clear
+circular pupil it reads 0.250448 / 0.243480 / 0.239942 / 0.238160 at
+32 / 64 / 128 / 256. Successive differences halve as the count doubles in
+both cases, so the central obscuration is not the cause.
+
+Richardson-extrapolating the Double Gauss lands on the OpticStudio value to
+every printed digit, on axis (0.236378 against 0.2364) and at full field
+(1.218146 against 1.2182). So for a clear pupil the entire gap is the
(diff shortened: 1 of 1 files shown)


HarrisonKramer:
Thanks, @RayPragma. This is a solid addition! I see some of the tests had been cancelled. Rerunning them now, and will merge once complete.

Best,
Kramer

PR #715, opened 2026-08-06
fix(aberrations): difference chromatic terms over the system's own wavelengths

RayPragma:
Found while cross-checking the Seidel sums against OpticStudio for #710.

`_precalculations` evaluates the dispersion as `n(0.4861) - n(0.6563)` regardless of what the optic defines:

```python
n_F = self._signed_refractive_indices(0.4861)
n_C = self._signed_refractive_indices(0.6563)
self._dn = n_F - n_C
```

The chromatic coefficients are a difference between two wavelengths, so those have to be wavelengths the system is actually specified over.

## Two ways this goes wrong, both silent

**Visible systems that aren't on the F and C lines.** `CookeTriplet` is specified at 0.48 / 0.55 / 0.65 — close enough to the hardcoded lines to look fine, far enough to matter:

| | `LchC` | `TchC` |
|---|---|---|
| Hardcoded 0.4861 / 0.6563 | 0.238325 | -0.026779 |
| The system's own 0.48 / 0.65 | **0.257127** | **-0.027873** |
| | 7.9% | 4.1% |

The first row is what's published as the System 2 reference in #710.

**Systems outside the visible.** A long-wave infrared germanium singlet at 8 / 10 / 12 µm has essentially no axial colour, because germanium is non-dispersive across that band in the shipped data (n = 4.653 at all three). Evaluating on the visible lines borrows an index swing of 0.4 that the system never sees:

| | `LchC` |
|---|---|
| The system's own 8–12 µm | **0.000000** — correct, no dispersion means no colour |
| Hardcoded visible lines | **5.759895** |

No error, no warning, just a number that looks like an answer. And germanium is the benign case: for many infrared glasses there is no refractive index data at 0.4861 µm at all.

## The fix

Use the extremes of `optic.wavelengths`, falling back to the primary wavelength when only one is defined — which makes the chromatic terms identically zero, the correct answer for a monochromatic system.

## Reference values that move

| Sample | Change | Why |
|---|---|---|
| `CookeTriplet` | 7.9% / 4.1% | Specified at 0.48 / 0.65 |
| `Edmund_49_847`, `SingletStopSurf2` | 0.04% | Specified on the *precise* F and C lines (0.4861327 / 0.6562725) while the hardcoded literals were rounded |
| `DoubleGauss` | none | Specified at exactly 0.4861 / 0.6563, so the two intervals coincide |

I've updated the affected golden values in `tests/test_aberrations.py` — only the three chromatic assertions per system. **Every monochromatic coefficient (TSC, SC, CC, TCC, TAC, AC, TPC, PC, DC) is bit-identical on every sample**, which is the check that the change is confined to the chromatic path rather than perturbing the paraxial trace.

You may want to regenerate the chromatic rows for System 1 and System 2 in #710 after this lands.

## Tests

`TestChromaticDispersionWavelengths` pins three properties:

- Shrinking a system's spectral range must shrink its axial colour (quarter the interval, less than half the aberration).
- A monochromatic system reports exactly zero.
- The LWIR case above does not pick up visible-band dispersion.

Verified all three fail with the hardcoding restored.

## Checks

- `ruff check optiland/` / `ruff format` — pass
- `pytest tests/test_aberrations.py` — 38 passed
- `pytest tests/ -k "aberration or seidel or chromatic or third_order"` — 110 passed, 1 skipped

---

Separately, and not part of this PR: I still can't reconcile `LchC` / `TchC` against OpticStudio's chromatic columns on the Double Gauss, and the remaining difference isn't explained by this fix — that system's wavelengths are exactly the hardcoded ones, so its values don't change here. I'll write that up on #710 with what I've eliminated; the short version is that per-surface `TSC` matches your `TSPH` to the printed resolution on every surface while the chromatic rows are off by 900 to 31,000 times that, and the ratio isn't constant across surfaces with identical media, so it isn't a convention factor either.

### DIFF
--- a/optiland/aberrations/third_order.py
+++ b/optiland/aberrations/third_order.py
@@ -140,13 +140,35 @@ class ThirdOrderAberrations:
         terms = [term_func(k) for k in range(1, self._N - 1)]
         return be.array(terms)
 
+    def _dispersion_wavelengths(self) -> tuple[float, float]:
+        """Return the (short, long) wavelengths used for the chromatic terms.
+
+        The chromatic aberration coefficients differentiate the system between
+        two wavelengths. Those have to be wavelengths the system is actually
+        specified over: an infrared design differenced across the visible F and
+        C lines produces a meaningless number, and for many infrared glasses
+        there is no refractive index data at 0.4861 µm to evaluate in the first
+        place.
+
+        Uses the extremes of the system's own wavelength set, falling back to
+        the primary wavelength when only one is defined — in which case the
+        chromatic terms are identically zero, which is the correct answer for a
+        monochromatic system.
+        """
+        values = [float(be.to_numpy(w.value)) for w in self._optic.wavelengths]
+        if not values:  # pragma: no cover - an Optic always carries one
+            primary = float(self._optic.primary_wavelength)
+            return primary, primary
+        return min(values), max(values)
+
     def _precalculations(self) -> None:
         self._inv: float = self._optic.paraxial.invariant()
         self._on_axis = be.isclose(self._inv, be.array(0.0))
         self._n = self._signed_refractive_indices(self._optic.primary_wavelength)
-        n_F = self._signed_refractive_indices(0.4861)
-        n_C = self._signed_refractive_indices(0.6563)
-        self._dn = n_F - n_C
+        short, long = self._dispersion_wavelengths()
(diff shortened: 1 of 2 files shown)


HarrisonKramer:
Good find! I agree with the proposal. Hardcoding the wavelengths for chromatic aberrations is not ideal. Merging now. Thanks!

Kramer

PR #717, opened 2026-08-06
fix(aberrations): use the ray height at the surface in the chromatic terms

RayPragma:
This is the root cause of the chromatic mismatch I was chasing on #710, and it turned out to be an off-by-one rather than anything about conventions.

## The bug

`_TAchC_term` and `_TchC_term` index the marginal ray as `_ya[k - 1]` while indexing the angle of incidence as `_i[k - 1]`. Those two have different bases:

| | built as | value for surface `k` |
|---|---|---|
| `_ya` | full per-surface array from `paraxial.marginal_ray()`, `_ya[0]` at the object | `_ya[k]` |
| `_i`, `_ip` | lists built over `range(1, N - 1)` | `_i[k - 1]` |

So every surface was weighted by the ray height at the **previous** surface. The monochromatic terms are unaffected: they go through `_B`, which is built inside that same loop and correctly uses `_ya[k]`.

Surface 1 came out right by accident — with the object at infinity the marginal ray is parallel, so `_ya[0] == _ya[1]`. That is why the totals looked plausible rather than obviously broken, and it is why the first surface was the one row that matched when I started comparing.

## Verification against OpticStudio

Per surface on the Double Gauss, against OpticStudio 2023 R1.02:

| | max abs. difference before | after |
|---|---|---|
| `TAchC` vs `TAXC` | 1.6e-02 | **4.2e-07** |
| `TchC` vs `TLAC` | 3.1e-02 | **7.8e-06** |

4.2e-07 is the printed resolution of the listing — the same agreement the monochromatic terms already had. Totals now line up too: `LchC` = -0.1400173 against `LAXC` = -0.140017, sign included.

## Verification against the aberrations themselves

Independent of any other code, since the coefficients have physical meanings:

**Axial colour** is the longitudinal shift of focus between F and C. On the Double Gauss that shift is 0.145931 mm. `LchC` was 0.295706 — 2.03x too large. It is now -0.140017, i.e. within 4%, which is the same margin OpticStudio's `LAXC` sits at and is the third-order approximation rather than an error.

**Lateral colour** is worse than a scale factor. On `Edmund_49_847` the two surface contributions cancelled exactly under the wrong indexing, and the suite asserted:

```python
assert_allclose(be.sum(TchC), 0.0)
```

for a system whose chief ray actually moves 0.01662 mm between F and C. The corrected coefficient gives 0.01675, within 0.8%. The bug had been frozen into the tests as a known good value, which is the part I'd flag hardest — it would have survived any amount of self-comparison.

## Tests

Updated the chromatic golden values on the three affected samples. **Every monochromatic coefficient (TSC, SC, CC, TCC, TAC, AC, TPC, PC, DC) is bit-identical**, which is the check that the change is confined to the chromatic path rather than perturbing the paraxial trace.

Added `TestChromaticRayHeightIndex`, which deliberately does *not* pin another golden number — it checks the coefficients against the focus shift and the chief-ray height shift computed from the paraxial trace at each wavelength, with a 5% tolerance for the third-order approximation. Both fail with the indexing restored.

## Relationship to #715

#715 fixes a different bug in the same two functions — the dispersion was differenced across hardcoded visible F and C lines rather than the system's own wavelengths. The two are independent (this one is *which ray height*, that one is *which wavelengths*) but they touch the same golden values, so whichever lands second will need its numbers regenerated. Happy to do that, or to fold them into one PR if you'd rather review them together. For reference, `DoubleGauss` is unaffected by #715 — it is specified at exactly 0.4861 / 0.6563 — so the OpticStudio agreement shown above holds with or without it.

## Checks

- `ruff check optiland/` / `ruff format` — pass
- `pytest tests/test_aberrations.py` — 36 passed

### DIFF
--- a/optiland/aberrations/third_order.py
+++ b/optiland/aberrations/third_order.py
@@ -231,21 +231,27 @@ class ThirdOrderAberrations:
         )
         return spherical + self._get_conic_term(k, p_ya=1, p_yb=3)
 
-    def _TAchC_term(self, k: int) -> BEArray:
+    def _chromatic_prefactor(self, k: int) -> BEArray:
+        """Shared factor of the two chromatic terms at surface ``k``.
+
+        Note the two different index bases in play. ``_ya`` is the full
+        marginal-ray array returned by ``paraxial.marginal_ray()``, indexed by
+        surface with ``_ya[0]`` at the object, so the height *at* surface ``k``
+        is ``_ya[k]``. ``_i`` and ``_ip`` are lists built over
+        ``range(1, N - 1)``, so the value for surface ``k`` is ``_i[k - 1]``.
+        Mixing the two up costs a one-surface shift in the ray height.
+        """
         return (
-            -self._ya[k - 1]
-            * self._i[k - 1]
+            -self._ya[k]
             / (self._n[-1] * self._ua[-1])
             * (self._dn[k - 1] - self._n[k - 1] / self._n[k] * self._dn[k])
         )
 
+    def _TAchC_term(self, k: int) -> BEArray:
+        return self._chromatic_prefactor(k) * self._i[k - 1]
+
     def _TchC_term(self, k: int) -> BEArray:
-        return (
-            -self._ya[k - 1]
-            * self._ip[k - 1]
-            / (self._n[-1] * self._ua[-1])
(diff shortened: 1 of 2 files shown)


HarrisonKramer:
Hi @RayPragma,

recognized. good find. Thanks for the fix! Merging now.

Best,
Kramer

PR #719, opened 2026-08-10
[codex] fix(wavefront): mask invalid chief-ray samples

### LINKED ISSUE #718: ChiefRayStrategy silently returns NaN when a pupil ray fails to intersect a surface (missing invalid-ray masking, unlike CentroidStrategy/BestFitStrategy)
### Checklist
- [x] I have searched the existing issues and discussions for a similar question or problem.
- [ ] I am using the **latest version** of Optiland (confirmed present in 0.6.0; not yet re-verified against current `main`)
- [x] I have included all necessary context, such as version info, error messages, or minimal reproducible examples.

### Bug Report

**Describe the bug**

`FFTPSF` and `FFTMTF` use `ChiefRayStrategy` (`optiland/wavefront/strategy.py`) as the default wavefront calculation strategy. When at least one ray in the pupil sampling grid fails to intersect a surface (the ray-surface quadratic in `StandardGeometry.distance()` has a negative discriminant — a real, unavoidable case for strongly aberrated or partially vignetted fields), that ray's `x/y/z/opd` come back as `nan`. `ChiefRayStrategy.compute_wavefront_data` does not filter or mask these values before building `WavefrontData`, so the `nan` propagates through the entire wavefront/PSF/MTF array. `strehl_ratio()` and the MTF value at any frequency then silently evaluate to `nan` — no exception, no warning — even though the same field is perfectly traceable geometrically (in our case, 97.7% of rays land validly; only 2.2-2.3% are the non-intersecting ones).

This is inconsistent with the other two built-in strategies in the same module, both of which already handle this case:

- `CentroidStrategy.compute_wavefront_data` uses `valid_mask = rays.i > 0` and raises a clear `ValueError` if no rays are valid.
- `BestFitStrategy` (via `CentroidStrategy._points_from_rays`) filters with `be.isfinite(rays.x) & be.isfinite(rays.y) & be.isfinite(rays.z) & ... & (rays.i != 0)`.

`ChiefRayStrategy` has no equivalent filter anywhere in `compute_wavefront_data`.

Note: this is related to but distinct from a previously reported issue in this same codebase, https://github.com/optiland/optiland/issues/716 (`BasePSF.strehl_ratio()` reading the array-center pixel instead of the true PSF peak) — that issue is about a *different* mechanism producing a wrong-but-finite Strehl value; this report is about a `nan` propagation path with no exception raised.

**To Reproduce**

1. Build any `Optic` with a field position whose pupil sampling grid contains at least one ray that does not physically intersect one of the surfaces (e.g. a corner field on a fast, obstructed, badly aberrated system — but in principle any system/field/grid combination where the ray-surface quadratic discriminant goes negative for at least one sampled ray).
2. Call `FFTPSF(optic, field=(Hx, Hy), wavelength=wl, num_rays=N, grid_size=N)` or `FFTMTF(optic, fields=[(Hx, Hy)], wavelength=wl, num_rays=N)`.
3. No exception is raised. `psf.strehl_ratio()` and the MTF value at any sampled frequency are `nan`.
4. A direct geometric trace of the same field (`optic.trace(Hx=Hx, Hy=Hy, wavelength=wl, num_rays=N, distribution="hexapolar")`) shows the large majority of rays land validly (finite `x`/`y`) — the field is not fully vignetted, so a `nan` result is not the physically expected outcome.

**Root cause (confirmed via traceback with numpy warnings promoted to errors)**

Forcing `warnings.simplefilter("error")` + `np.errstate(all="raise")` around the `FFTPSF`/`FFTMTF` call turns the normally-silent `nan` into a real exception, with this traceback:

```
FloatingPointError: invalid value encountered in sqrt
  File ".../optiland/geometries/standard.py", line 134, in distance
    t1 = (-b + be.sqrt(d)) / (2 * a)
```

Note that `StandardGeometry.distance()` already wraps this exact line in `warnings.catch_warnings(); warnings.simplefilter("ignore")` — i.e. the library already anticipates and deliberately silences this case, returning `t = nan` for a ray with no real surface intersection (discriminant `d < 0`). That part is working as designed; a ray genuinely may not intersect a given conic/spherical surface, and returning `nan` for `t` (rather than raising) is a reasonable low-level contract, *provided the caller masks it out*.

The actual gap is one level up, in `optiland/wavefront/strategy.py::ChiefRayStrategy.compute_wavefront_data`:

```python
def compute_wavefront_data(self, field, wavelength):
    ...
    rays = self.optic.trace(*field, wavelength, None, self.distribution)
    intensity = self.optic.surfaces.intensity[-1, :]
    opd_img = geometry.path_length(rays, self.n_image)
    opd = rays.opd - opd_img
    opd = self._correct_tilt(field, opd)
    opd_wv = (opd_ref - opd) / (wavelength * 1e-3)
    t = opd_img / self.n_image
    pupil_x = rays.x - t * rays.L
    pupil_y = rays.y - t * rays.M
    pupil_z = rays.z - t * rays.N
    ...
    return WavefrontData(pupil_x=pupil_x, pupil_y=pupil_y, pupil_z=pupil_z,
                          opd=opd_wv, intensity=intensity, ...)
```

There is no `be.isfinite(...)` check and no use of `rays.i` anywhere in this method. Every ray, including any with `nan` propagated from `distance()`, is passed straight into the returned `WavefrontData`. Downstream, `FFTPSF`/`FFTMTF` build the OPD/pupil-function array directly from this data, so a single `nan` entry contaminates the whole array — and by extension the FFT, the Strehl ratio, and the MTF at every frequency — silently.

For comparison, `CentroidStrategy.compute_wavefront_data` in the same file:

```python
valid_mask = rays.i > 0
if be.any(valid_mask):
    mean_opd = be.mean(opd[valid_mask])
else:
    raise ValueError("No valid rays with non-zero intensity for OPD calculation.")
```

and `BestFitStrategy` (via the shared `CentroidStrategy._points_from_rays`):

```python
valid = (
    be.isfinite(rays.x) & be.isfinite(rays.y) & be.isfinite(rays.z)
    & be.isfinite(rays.L) & be.isfinite(rays.M) & be.isfinite(rays.N)
    & be.isfinite(rays.opd) & (rays.i != 0)
)
if not be.any(valid):
    raise ValueError("No valid ray samples found for best-fit geometry.")
```

Both siblings already treat this as an expected condition to guard against. `ChiefRayStrategy` — which happens to be the *default* strategy (`strategy: WavefrontStrategyType = "chief_ray"` in `Wavefront.__init__`, and therefore the one used by `FFTPSF`/`FFTMTF` unless a caller overrides it) — does not.

**Expected behavior**

A ray that fails to intersect a surface should be treated as a zero-amplitude (vignetted) contribution to the wavefront/PSF/MTF calculation — consistent with how `CentroidStrategy` and `BestFitStrategy` already handle the same condition — rather than silently poisoning the entire array with `nan`. At minimum, the library should never return a `nan` Strehl ratio or MTF value without raising an exception or warning, when the field is not, in fact, fully vignetted.

Suggested fix:

Mirror the masking already present in `CentroidStrategy`/`BestFitStrategy` inside `ChiefRayStrategy.compute_wavefront_data`: after computing `opd_wv`, `pupil_x/y/z`, and `intensity`, identify any non-finite entries and set their `intensity` (and optionally `opd`/`pupil_x/y/z`) to `0` rather than passing `nan` through, so downstream FFT/Strehl/MTF computations correctly treat that ray as a zero-amplitude (vignetted) contribution instead of poisoning the whole array. Alternatively, raise a clear exception (consistent with `CentroidStrategy`'s `ValueError` when *no* rays are valid) if the fraction of invalid rays exceeds some threshold, so callers get an explicit signal instead of a silent `nan`.

**Screenshots**

Not applicable — this is a numerical/logic bug, not a visual one.

**Environment**

- Optiland Version: `0.6.0`
- Python Version: `3.12.2`
- OS: `macOS`
- Additional Info: Backend used: `numpy`. Affects `optiland/wavefront/strategy.py` (`ChiefRayStrategy.compute_wavefront_data`), and by extension `optiland/psf/fft.py` (`FFTPSF`) and `optiland/mtf/fft.py` (`FFTMTF`), since `ChiefRayStrategy` is the default wavefront strategy. Also relevant: `optiland/wavefront/wavefront.py`, `optiland/geometries/standard.py`.

**Additional context**

Impact:
- Silent failure: no exception, no warning surfaced to the user by default (the underlying `RuntimeWarning` from `sqrt` is suppressed intentionally inside `distance()`).
- Affects any moderately-to-severely aberrated or partially vignetted field evaluated with `FFTPSF`/`FFTMTF` at default settings — not an exotic edge case. In our use case (a fast, centrally-obstructed two-mirror system with a corrector), one field out of five (the FOV corner) hit this on every grid/ray-count setting tested (24 through 256), always returning `Strehl = nan` and `MTF = nan`, even though geometric ray tracing confirmed ~97.7% of rays for that field land validly.
- Because the failure is silent, it is easy to mistake a `nan` result for "not yet computed" or to have it slip through an equality/threshold check (`nan < 0.4` is `False`, so a naive "does it meet spec" check would incorrectly report the requirement as *not* violated by omission, or conversely a downstream aggregation like `np.mean([...values..., nan])` poisons an otherwise-valid average — either failure mode is worse than a raised exception).

Workaround we are using in the meantime — a local monkeypatch, applied once at the start of the session before any `FFTPSF`/`FFTMTF` call:

```python
import optiland.wavefront.strategy as wf_strategy_mod
import optiland.backend as be

_original_compute_wavefront_data = wf_strategy_mod.ChiefRayStrategy.compute_wavefront_data

def _patched_compute_wavefront_data(self, field, wavelength):
    data = _original_compute_wavefront_data(self, field, wavelength)
    invalid = ~(
        be.isfinite(data.opd) & be.isfinite(data.pupil_x)
        & be.isfinite(data.pupil_y) & be.isfinite(data.pupil_z)
    )
    if be.any(invalid):
        data.opd = be.where(invalid, 0.0, data.opd)
        data.pupil_x = be.where(invalid, 0.0, data.pupil_x)
        data.pupil_y = be.where(invalid, 0.0, data.pupil_y)
        data.pupil_z = be.where(invalid, 0.0, data.pupil_z)
        data.intensity = be.where(invalid, 0.0, data.intensity)
    return data

wf_strategy_mod.ChiefRayStrategy.compute_wavefront_data = _patched_compute_wavefront_data
```

We verified this against the failing field/system described above: it zeroed out 273/12644 (~2.2%) rays in the `FFTPSF` call and 68/3096 (~2.2%) rays in the `FFTMTF` call — consistent with the ~2.3% of rays independently found invalid by a direct geometric trace of the same field — and produced finite, physically sensible (if poor) `Strehl`/`MTF` values instead of `nan`.

We can provide a minimal, self-contained reproduction script (synthetic system + field, no dependency on our full design) if that's useful for a PR or test case.

breezeFur:
## Summary

- mask non-finite chief-ray wavefront samples before they reach PSF/MTF calculations
- zero invalid OPD, pupil coordinates, and intensity while preserving valid rays
- raise a clear error when no valid samples remain
- add a regression test for a partially invalid ray bundle

## Root cause

`ChiefRayStrategy.compute_wavefront_data` passed non-finite ray values directly into `WavefrontData`. A single failed surface intersection could therefore contaminate the pupil function and FFT with `NaN`, unlike the other wavefront strategies that already validate ray samples.

## Validation

- `python -m pytest tests/test_wavefront_strategy.py -q` (18 passed)
- `python -m ruff check optiland/wavefront/strategy.py tests/test_wavefront_strategy.py`
- `python -m ruff format --check optiland/wavefront/strategy.py`
- `git diff --check`

Fixes #718

### DIFF
--- a/optiland/wavefront/strategy.py
+++ b/optiland/wavefront/strategy.py
@@ -206,7 +206,6 @@ class ChiefRayStrategy(ReferenceStrategy):
             & be.isfinite(pupil_y)
             & be.isfinite(pupil_z)
             & be.isfinite(intensity)
-            & (intensity > 0)
         )
         if not be.any(valid_mask):
             raise ValueError("No valid ray samples found for chief-ray wavefront.")
--- a/optiland/wavefront/strategy.py
+++ b/optiland/wavefront/strategy.py
@@ -200,6 +200,22 @@ class ChiefRayStrategy(ReferenceStrategy):
         pupil_y = rays.y - t * rays.M
         pupil_z = rays.z - t * rays.N
 
+        valid_mask = (
+            be.isfinite(opd_wv)
+            & be.isfinite(pupil_x)
+            & be.isfinite(pupil_y)
+            & be.isfinite(pupil_z)
+            & be.isfinite(intensity)
+            & (intensity > 0)
+        )
+        if not be.any(valid_mask):
+            raise ValueError("No valid ray samples found for chief-ray wavefront.")
+        opd_wv = be.where(valid_mask, opd_wv, be.zeros_like(opd_wv))
+        pupil_x = be.where(valid_mask, pupil_x, be.zeros_like(pupil_x))
+        pupil_y = be.where(valid_mask, pupil_y, be.zeros_like(pupil_y))
+        pupil_z = be.where(valid_mask, pupil_z, be.zeros_like(pupil_z))
+        intensity = be.where(valid_mask, intensity, be.zeros_like(intensity))
+
         # 6. Handle polarization data if available
         kwargs = {}
         prt_matrix = getattr(rays, "p", None)
--- a/tests/test_fft_psf.py
+++ b/tests/test_fft_psf.py
@@ -219,10 +219,9 @@ def test_invalid_working_FNO(make_fftpsf):
     def tweak(optic):
         optic.surfaces[0].geometry.cs.z = -be.array(1e100)
 
-    fftpsf = make_fftpsf(field=(0, 1), tweak_optic=tweak)
+    # Error now surfaces at construction, not at .view().
     with pytest.raises(ValueError):
-        fig, ax = fftpsf.view()
-        plt.close(fig)
+        make_fftpsf(field=(0, 1), tweak_optic=tweak)
 
 
 def test_interpolate_zoom_factor_one(make_fftpsf):
--- a/tests/test_wavefront_strategy.py
+++ b/tests/test_wavefront_strategy.py
@@ -195,6 +195,51 @@ class TestChiefRayStrategy:
         assert isinstance(wavefront_data.radius, float)
         assert wavefront_data.radius > 0
 
+    def test_masks_non_finite_ray_samples(self, set_test_backend):
+        """Test that invalid rays cannot contaminate downstream wavefront data."""
+        optic = MagicMock()
+        optic.primary_wavelength = 0.55
+        optic.surfaces.n.return_value = be.array([1.0])
+        optic.surfaces.positions = [0.0]
+        optic.surfaces.intensity = be.array([[1.0, 1.0]])
+        optic.paraxial.XPL.return_value = 0.0
+        optic.fields.field_definition = object()
+
+        chief_ray = MagicMock()
+        chief_ray.x = be.array(0.0)
+        chief_ray.y = be.array(0.0)
+        chief_ray.z = be.array(10.0)
+        chief_ray.opd = be.array(0.0)
+        optic.trace_generic.return_value = chief_ray
+
+        rays = MagicMock()
+        rays.x = be.array([0.0, float("nan")])
+        rays.y = be.array([0.0, float("nan")])
+        rays.z = be.array([10.0, float("nan")])
+        rays.L = be.array([0.0, float("nan")])
+        rays.M = be.array([0.0, float("nan")])
+        rays.N = be.array([1.0, float("nan")])
+        rays.opd = be.array([0.0, float("nan")])
+        rays.p = None
+        rays.get_exit_fields = None
+        optic.trace.return_value = rays
+
+        geometry = MagicMock(radius=10.0)
+        geometry.path_length.side_effect = [
+            be.array(0.0),
+            be.array([0.0, float("nan")]),
+        ]
+        strategy = ChiefRayStrategy(optic, MagicMock())
+        strategy._create_reference_geometry = MagicMock(return_value=geometry)
+
+        wavefront_data = strategy.compute_wavefront_data((0.0, 0.0), 0.55)
+
+        assert be.all(be.isfinite(wavefront_data.pupil_x))
+        assert be.all(be.isfinite(wavefront_data.pupil_y))
+        assert be.all(be.isfinite(wavefront_data.pupil_z))
+        assert be.all(be.isfinite(wavefront_data.opd))
+        assert_allclose(wavefront_data.intensity, be.array([1.0, 0.0]))
+
 
 class TestCentroidReferenceSphereStrategy:
     """Tests for the CentroidReferenceSphereStrategy."""


HarrisonKramer:
Pushed a fix for the CI failures (both were caused by the same root issue):

**`intensity > 0` in the validity mask was too broad.** It conflated "ray failed to intersect a surface" (the actual bug from #718) with "ray is legitimately vignetted/blocked" (finite, zero-intensity by design). Zeroing OPD for blocked-but-successfully-traced rays silently changed results on any obscured-aperture system — caught by `test_opd_diff_new_dist`, which computes `RayOperand.OPD_difference` on the Hubble (central obstruction). Dropped that clause; `isfinite()` on opd/pupil/intensity already covers the #718 failure mode, and matches how `CentroidStrategy`/`BestFitStrategy` mask (isfinite-only, no intensity gate).

**`test_invalid_working_FNO` needed updating**, unrelated to the above: the ValueError now raises during wavefront generation (PSF construction) rather than surfacing later at `.view()`, since the tweak makes every sample invalid. Moved the `make_fftpsf(...)` call inside `pytest.raises`.

Verified: `pytest tests/ -k "psf or wavefront or operand or zernike or mtf or strehl or fft"` → 953 passed, plus the two originally-failing tests now pass on both backends.

HarrisonKramer:
Hi @breezeFur,

Thanks for the fix! Code makes sense, and is clean. All tests pass. I will set to PR to ready for review, then merge.

Best,
Kramer

PR #554, opened 2026-04-04
improve NR implicit differentiation

manuelFragata:
### Newton-Raphson implicit-diff hardening: direct gradient tests, graph-size regressions, and benchmark coverage

**Summary**
This PR strengthens the Newton-Raphson implicit-differentiation path by improving numerical safeguards, expanding direct correctness coverage, and adding explicit regressions for the key design claim that autograd graph size should remain O(1) in max_iter for the implicit path.

**Why**
The implicit NR path is intended to preserve first order gradients without storing per-iteration autograd history. This behavior is central to memory scalability and needed direct regression coverage at the geometry.distance primitive level.

**What changed**
1. Reworked torch differentiable distance path around no-grad primal solve plus implicit correction
2. Added sign-preserving denominator floor logic for stability near small denominators
3. Documented first-order-only derivative contract for the implicit formulation

4. Added direct distance-level tests:
- AD vs FD for radius, conic, and one asphere coefficient
- AD vs FD for a ray-state variable
- On-axis and off-axis cases
- Forward-value consistency check between differentiable and primal results

5. Added stronger numpy regressions:
- Deterministic reference outputs
- Numpy vs torch no-grad consistency checks

6. Added graph-complexity regressions over max_iter:
- Implicit path graph nodes remain effectively constant
- Unrolled baseline graph nodes grow with max_iter

**Key regression result**
- Implicit path node count stays flat as max_iter increases
- Unrolled baseline node count grows strongly with max_iter

To-Do:
- [x] ray-surface intersection improvement.
- [ ] improvement on iterative NR for real image height field type (more iinfo in #547)

Closes #335

PR #724, opened 2026-08-12
Memory-efficient implicit differentiation for NR intersection and real image height

### LINKED ISSUE #335: Addressing Memory Inefficient NR ray-surface intersection algorithm in the torch backend
### Checklist

- [ ] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or feature request.
- [ ] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [ ] I am using the **latest version** of Optiland.
- [ ] I have included **all necessary context**.

### Feature Request

The current implementation of the NR method to find the intersection of rays with surfaces works well. One of the nice things about Optiland is its dual backend architecture, which when using Pytorch, allows us to use the Differentiable Ray Tracing methodology for implementing optimization pipelines, etc. However, using a large number of rays, for a not so easy to define surface, like a freeform, the iterative NR is currently recording all the iterations steps in the computational graph in the forward path. This impacts the memory consumption very negatively. Therefore, it is a **must** that we somehow improve this method. 

As described in the paper "dO: A differentiable engine for Deep Lens design
 of computational imaging systems" (https://github.com/vccimaging/DiffOptics), there is a smarter way to implement the root solving method to find the intersection of the rays with the surfaces that is much more memory efficient. I would say we work out their logic, how exactly are they bypassing this problem while retaining the AD graph intact, and iterate from there - or simply implement their approach (with the correct copyright notice of course).

Since this is probably a major bottleneck in performance for Differentiable Ray Tracing, I am pretty confident that there are some people already working on smarter more efficient ways of finding the intersections of rays with surfaces, potentially speeding up the whole design optimization process in DRT pipelines. However, I am unaware of anyone working on it concretely. Maybe someone else out there would like to help us out implement some nice algos.

HarrisonKramer:
Finalizes #554 / the reland in this PR, and adds the second objective (`real_image_height`) that was still outstanding. Closes #335.

Reopened from #554 by @HarrisonKramer; original no-grad + single-correction approach by @Kushagra1480 in #550.

## Why

Ray-surface intersection and real-image-height field solving are both iterative. Recording every iteration in autograd makes the graph grow linearly with `max_iter`, which dominates memory on large ray batches (#335). Both now run the iteration graph-free and recover the derivative from a single grad-attached correction (implicit function theorem).

## What changed

**Two latent gradient bugs in `real_image_height`.** `be.full_like` / `be.array` were detaching the entrance-pupil location and the field parameters, so gradients were silently *wrong* rather than zero. `d(origin y)/d(radius)` was 89% off for an infinite object and 1.5% off for a finite one; both now match central differences to ~1e-8.

**`real_image_height` now solves the real problem.** The old iteration used independent scalar secants with a shared paraxial scale — valid only for centered, rotationally symmetric systems. It now uses a full 2×2 Jacobian with damped Newton/Broyden, backtracking, and a one-trace implicit correction. Solved once per field point, not per pupil ray.

**Newton-Raphson hardening.** Convergence is checked before the surface normal is evaluated; a per-ray convergence mask means a failed root gets a *detached* gradient plus a warning instead of a fabricated one; thresholds are dtype- and scale-aware (the fixed `1e-14` sat below float32 round-off and never triggered).

**Forbes.** The old branch-only vertex/epsilon edits are dropped in favour of master's #620 work — 501 Forbes/geometry/qpoly tests pass without them. The Forbes diff here is now only the cache-lifecycle fix.

## Contract

First-order derivatives only. Forward values are preserved within solver tolerance, not bitwise (one final Newton correction is applied). Double backward runs and stays finite but its value is not guaranteed. NumPy behavior is unchanged.

Exactness assumes a converged simple root, a non-singular `F_t` / `det J_q`, and caches that are grad-attached when the correction runs. Where those fail the code warns or raises rather than returning a confident wrong answer.

Details: `docs/developers_guide/implicit_differentiation.rst`.

## Evidence

10 000 rays, torch/CPU/float64, vs an unrolled baseline:

| max_iter | implicit nodes | unrolled nodes |
|---:|---:|---:|
| 5 | **198** | 643 |
| 20 | **198** | 1978 |
| 80 | **198** | 7318 |

Graph size is O(1) in iteration count. Root residual is identical (2.55e-15) in every row, and the implicit gradient matches the unrolled one to **2.0e-16** relative. Field solve: 852 nodes regardless of `max_iter`, and forward cost flat from 1 to 1024 pupil samples.

## Tests

Full suite **7616 passed**, 16 skipped, 2 xfailed. `ruff check` / `format --check` clean.

New: `tests/test_nr_implicit_diff_forbes_cache.py`, `tests/test_real_image_height_implicit_diff.py`. The Forbes cache tests were verified as genuine regression coverage — with the fix reverted, all six fail.

Removed the test asserting the second derivative must differ from FD by >2%. Encoding "must stay inaccurate" would block higher-order support later; replaced with a first-order AD/FD contract test plus a double-backward finiteness check.


manuelFragata:
haha perfect, no worries!

manuelFragata:
@HarrisonKramer — I've finalized this on top of your reland. Branch: `feat/nr-implicit-diff-finalization` (8 commits, off current master; your two commits are preserved as the first two).

**Ask: may I force-push it onto `reland-554-nr-implicit-diff` to update this PR?** That replaces your current tip, so I'd rather not do it without a yes. Happy to open a separate PR instead.

What it changes:

- **Two latent gradient bugs in `real_image_height`.** `be.full_like` / `be.array` were detaching the entrance-pupil location and the field parameters, so gradients were silently *wrong* rather than zero. `d(origin y)/d(radius)` was 89% off for an infinite object and 1.5% off for a finite one; both now match central differences to ~1e-8.
- **The second PR objective.** `real_image_height` now uses a full 2×2 Jacobian with damped Newton/Broyden and backtracking, plus a one-trace implicit correction. The old diagonal-only secants only worked for centered rotationally symmetric systems.
- **NR hardening.** Convergence is checked before the normal is evaluated; per-ray convergence mask so a failed root gets a detached gradient instead of a fabricated one; dtype- and scale-aware thresholds — the fixed `1e-14` is below float32 round-off and never triggered.
- **Forbes.** Dropped the old vertex/epsilon edits in favour of master's #620 work; the Forbes diff here is now just the cache fix.
- Credited @Kushagra1480 for the original #550 approach.

Full suite 7616 passed, ruff clean. Graph size: implicit stays at 198 nodes for any `max_iter`, unrolled grows 643 → 7318 (5 → 80 iterations), and the gradients agree to 2e-16.

HarrisonKramer:
Hey @manuelFragata, yes, go ahead. No worries.

I won't be able to review any code today, but I'll trust your fix if you think it's ready to merge after push/finalizations etc.

Thanks!
Kramer

manuelFragata:
Thanks @HarrisonKramer — pushed.

Your two commits are preserved as the first two (rebased onto current master, so the SHAs changed). Force-pushed with a lease on `8688bba0`, so nothing of yours was raced over.

Not merging until CI is green

PR #727, opened 2026-08-14
fix(paraxial): make surface positions fold-aware

### LINKED ISSUE #726: Paraxial machinery misreads folded systems: `positions` degenerates to global z
### Checklist

**Version:** 0.6.1, NumPy backend, Windows. Reproduces on current master.

## Summary

Optiland's real-ray engine traces folded systems correctly: a flat mirror
authored with an absolute coordinate system (`x/y/z` + `rx`) folds the beam
out of the z axis and explicit rays land where they should, to machine
precision. The first-order machinery does not come along. `SurfaceGroup.
positions` is defined as the global z of each vertex, and `Paraxial` derives
every spacing from it, so two surfaces on opposite sides of a 90° fold read
as coincident. Pupil locations, paraxial ray heights, and everything built
on them (EPD ray aiming, solves, image-space readouts) silently go wrong,
while the system looks healthy because real rays stay exact.

## Minimal repro

The same plano-convex singlet, built three ways: straight, retro (flat
mirror at normal incidence, the fold z can express), and folded 90° by a
45° mirror 20 mm behind the lens (absolute coordinates). Retro and folded
each unfold across their mirror plane to the same straight system, so every
first-order quantity must agree between them.

```python
import math

import optiland.backend as be
from optiland.rays import RealRays


def straight():
    o = Optic(name="straight")
    o.add_surface(index=0, radius=be.inf, thickness=be.inf)
    o.add_surface(index=1, radius=25.84, thickness=4.0, material="N-BK7", is_stop=True)
    o.add_surface(index=2, radius=be.inf, thickness=46.0)
    o.add_surface(index=3)
    return _finish(o)


def retro():
    o = Optic(name="retro")
    o.add_surface(index=0, radius=be.inf, thickness=be.inf)
    o.add_surface(index=1, radius=25.84, thickness=4.0, material="N-BK7", is_stop=True)
    o.add_surface(index=2, radius=be.inf, thickness=20.0)
    o.add_surface(index=3, radius=be.inf, material="mirror", thickness=-26.0)
    o.add_surface(index=4)
    return _finish(o)


def folded():
    o = Optic(name="folded")
    o.add_surface(index=0, radius=be.inf, thickness=be.inf)
    o.add_surface(index=1, radius=25.84, thickness=4.0, material="N-BK7", is_stop=True)
    o.add_surface(index=2, radius=be.inf, thickness=20.0)
    # 45 deg fold mirror 20 mm behind the lens exit: +z beam becomes +y.
    o.add_surface(index=3, x=0.0, y=0.0, z=24.0, rx=math.pi / 4, material="mirror")
    # Image plane 26 mm along the folded axis (equivalent unfolded z = 50).
    o.add_surface(index=4, x=0.0, y=26.0, z=24.0, rx=-math.pi / 2)
    return _finish(o)


def _finish(o):
    o.set_aperture(aperture_type="EPD", value=10.0)
    o.set_field_type("angle")
    o.add_field(y=0.0)
    o.add_wavelength(value=0.55, is_primary=True)
    return o


for o in (straight(), retro(), folded()):
    print(f"== {o.name} ==")
    print("positions:", be.ravel(o.surfaces.positions).tolist())
    print("f2:", float(o.paraxial.f2()))
    print("XPL:", float(o.paraxial.XPL()))
    y, u = o.paraxial.marginal_ray()
    print("paraxial marginal y:", [round(float(v), 4) for v in be.ravel(y)])
    # Explicit marginal ray (y = +3 mm, parallel to axis), no pupil aiming.
    r = RealRays(be.array([0.0]), be.array([3.0]), be.array([-10.0]),
                 be.array([0.0]), be.array([0.0]), be.array([1.0]),
                 be.array([1.0]), be.array([0.55]))
    o.surfaces.trace(r)
    print("real marginal ray lands at:",
          (round(float(r.x[0]), 6), round(float(r.y[0]), 6), round(float(r.z[0]), 6)))
```

## Observed (0.6.1)

| quantity | straight | retro | folded 90° |
|---|---|---|---|
| `positions` | [-inf, 0, 4, 50] | [-inf, 0, 4, 24, **-2**] | [-inf, 0, 4, 24, **24**] |
| `f2` | 49.833914 | -49.833914 | -49.833914 |
| `XPL` | -48.634140 | 48.634140 | **22.634140** |
| paraxial marginal y at image | 0.1204 | 0.1204 | **2.7290** |
| real marginal ray, distance off folded axis at image | 0.060878 | n/a | 0.060878 |

The real-ray layer treats folded and straight identically: 0.060878 mm
residual in both, to the printed digit, with the folded ray landing exactly
on the rotated image plane. The first-order layer does not: retro and folded
unfold to the same straight system, yet folded's XPL is off by exactly 26 mm (the post-fold
distance that z cannot express) and its paraxial marginal ray never
converges after the mirror (2.7290 mm at the image against retro's 0.1204).
The f2 sign flip after one mirror is the existing convention (retro shows
it too) and is not part of the problem.

## Cause

`SurfaceGroup.positions` returns `surf.geometry.cs.position_in_gcs[2]` per
surface (`optiland/surfaces/surface_group.py`, ~line 216). `Paraxial` reads
that array for launch z, object position, pupil propagation, and the stop
position (`optiland/paraxial.py`, lines 83, 115, 234, 256, 283). Global z
is a valid axial coordinate exactly when every leg of the beam path runs
along ±z: then each leg's length is |Δz| and the reflection sign convention
handles direction, which covers straight systems and on-axis mirrors. A
fold out of the z axis breaks the premise. Here the
mirror and the image plane share z = 24, so their spacing reads 0 and 26 mm
of propagation vanish from every first-order quantity. The real-ray engine
reads each surface's full coordinate system and never touches `positions`,
which is why it stays exact and why the failure is silent.

## Proposal

**A (preferred): make `positions` fold-aware.** Define the axial coordinate
as signed cumulative vertex-to-vertex path length along the surface chain,
with the sign convention chosen so that straight and on-axis-mirror systems
reproduce today's values exactly (for those systems the definition reduces
to global z). Acceptance: the folded build above reports retro's numbers
(XPL 48.634140, marginal y 0.1204 at the image) and straight/retro are
bit-for-bit unchanged. A two-fold periscope, where parity returns positive,
belongs in the tests. The physics is the standard coordinate-break argument:
a flat fold is first-order neutral, so the unfolded axis is the correct
paraxial coordinate. A powered mirror in a folded path contributes its power
as usual; only the flat-fold neutrality claim is load-bearing here. (A
powered mirror at oblique incidence is astigmatic, tangential and sagittal
powers split by the incidence cosine, so any 1-D axial model is approximate
there. The classic authoring shares that limitation; it is out of scope.)

**B: an explicit coordinate-break element** (Zemax's model). Heavier API,
orthogonal benefits for authoring, not required to fix the readouts.

**C: an official override**, e.g. `surfaces.set_axial_positions(seq)`, for
generators that already know the arc table. Cheap, and it would replace the
subclass-swap monkey-patching that downstream code needs today, but it fixes
nothing by default.

I can implement A with the acceptance tests above, and C alongside if
wanted.

manuelFragata:
Fixes #726, proposal A.

`positions` now reports the axial coordinate along the unfolded optical axis: signed cumulative vertex-to-vertex path length, projected on the running beam direction (each mirror turns it through d - 2(d.n)n), sign flipped per reflection. Systems whose legs all run along +/-z take an early exit and get global z back bit-for-bit, so straight and on-axis-mirror systems are unchanged. Real-space consumers (Huygens image plane, wavefront reference sphere, viewer limits) move to the new `global_z_positions` and behave as before.

Two things surfaced while validating against a folded system that is entered along +x rather than z:

- The walk starts along the first leg, object vertex to first surface vertex, rather than assuming +z. A system can be entered off the z axis entirely, and then no leg of it is a z interval. The direction is read off vertices, not the object plane's orientation, so a tilted object plane does not read as a fold.
- `ObjectSurface.is_infinite` accepts the object's infinity in x or y, which is where it lands when the entry axis is not z.

Tests: the issue's 90 deg fold against the retro system it unfolds onto, a two-fold periscope against the straight equivalent, a singlet entered along +x, and the tilted-object guard. The acceptance numbers from the issue reproduce. The regression snapshots pass unchanged, and the set of failing tests on my machine is identical before and after the change.

One limitation stays, tracked in #728: the ray generator still builds pupil points on a plane perpendicular to z, so `optic.trace()` remains z-bound for systems entered off axis. First-order readouts (marginal and chief ray, pupils, f2) work.

DO NOT MERGE YET, THANKS :)

### DIFF
--- a/optiland/coordinate_system.py
+++ b/optiland/coordinate_system.py
@@ -161,6 +161,24 @@ class CoordinateSystem:
             self.reference_cs.globalize(rays)
 
     @property
+    def frame_in_gcs(
+        self,
+    ) -> tuple[tuple[BEArray, BEArray, BEArray], tuple[BEArray, BEArray, BEArray]]:
+        """Returns the origin and the local +z axis of the coordinate system,
+            both expressed in the global coordinate system.
+
+        A single globalized probe carries both: its position becomes the
+        origin and its direction cosines become the local +z axis, which is
+        the surface normal at the vertex.
+
+        Returns:
+            A tuple ``(origin, axis)``, each a tuple of x, y and z components.
+        """
+        vector = RealRays(0, 0, 0, 0, 0, 1, 1, 1)
+        self.globalize(vector)
+        return (vector.x, vector.y, vector.z), (vector.L, vector.M, vector.N)
+
+    @property
     def position_in_gcs(self) -> tuple[BEArray, BEArray, BEArray]:
         """Returns the position of the coordinate system in the global coordinate
             system.
@@ -168,9 +186,7 @@ class CoordinateSystem:
         Returns:
             A tuple containing the x, y, and z coordinates of the position.
         """
-        vector = RealRays(0, 0, 0, 0, 0, 1, 1, 1)
-        self.globalize(vector)
-        return vector.x, vector.y, vector.z
+        return self.frame_in_gcs[0]
 
     def get_rotation_matrix(self) -> BEArray:
         """Get the rotation matrix of the coordinate system
--- a/optiland/psf/huygens_fresnel.py
+++ b/optiland/psf/huygens_fresnel.py
@@ -267,7 +267,9 @@ class ScalarHuygensPSF(BasePSF):
         pupil_opd_ideal = be.zeros_like(data.opd)  # ideal case has no OPD
         image_x = be.zeros((1, 1))  # single point for normalization
         image_y = be.zeros((1, 1))
-        ideal_z = self.optic.surfaces.positions[-1]  # image plane position
+        # Real-space image plane position: this seeds a global-frame point, so
+        # it must be global z rather than the unfolded axial coordinate.
+        ideal_z = self.optic.surfaces.global_z_positions[-1]
         image_z = be.full((1, 1), ideal_z)
 
         psf_max = self._summation_strategy.compute(
--- a/optiland/psf/vectorial_huygens.py
+++ b/optiland/psf/vectorial_huygens.py
@@ -142,7 +142,8 @@ class VectorialHuygensPSF(ScalarHuygensPSF):
         pupil_opd_ideal = be.zeros_like(data.opd)
         image_x = be.zeros((1, 1))
         image_y = be.zeros((1, 1))
-        ideal_z = self.optic.surfaces.positions[-1]
+        # Global-frame image plane position, not the unfolded axial coordinate.
+        ideal_z = self.optic.surfaces.global_z_positions[-1]
         image_z = be.full((1, 1), ideal_z)
 
         is_valid = data.intensity > 0
--- a/optiland/surfaces/object_surface.py
+++ b/optiland/surfaces/object_surface.py
@@ -47,8 +47,19 @@ class ObjectSurface(Surface):
 
     @property
     def is_infinite(self):
-        """Returns True if the surface is infinitely far away, False otherwise."""
-        return be.isinf(self.geometry.cs.z)
+        """Returns True if the surface is infinitely far away, False otherwise.
+
+        Classic authoring spells an object at infinity as ``cs.z = -inf``,
+        because the axis is z. A system entered off that axis -- a collimated
+        source posed to fire along +x into a fold -- puts the same infinity in
+        x or y, so any infinite vertex coordinate says the same thing.
+        """
+        cs = self.geometry.cs
+        # Tested one axis at a time: under a multi-configuration the three
+        # can hold arrays of different lengths, which will not stack.
+        return any(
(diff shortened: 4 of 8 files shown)


manuelFragata:
Merged into #729 — the aiming fix builds directly on the fold-aware positions, and the two changes were hardened and tested as one unit (shared ParaxialPath metadata, orientation-aware powered surfaces, explicit domain gating, one test suite). Reviewing them together is easier than keeping a stacked pair; #729 now carries the complete stack and its description covers everything that was here. The `fix/folded-paraxial-positions` branch is left in place for reference.

PR #734, opened 2026-08-15
feat(visualization): warn when lens surfaces physically overlap (#731)

### LINKED ISSUE #731: Warn when lens surfaces overlap in the visualization system
### Checklist

- [x] I have searched the existing issues and discussions for a similar question or feature request.
- [x] I have included all necessary context.

### Feature Request

`optiland/visualization/system/lens.py` currently has a `# TODO: raise warning when lens surfaces overlap` marker. If a lens is defined with a radius/thickness/semi-diameter combination that causes two of its surfaces to physically overlap, nothing warns the user: the invalid geometry is silently drawn (and can silently trace, too).

Add a check in the lens-drawing code path that detects when neighboring surfaces of a lens would overlap given their radii, thicknesses, and semi-diameters, and raise a `UserWarning` (not a hard error) so the user notices the inconsistent prescription.

N/A: this is filling in an existing TODO rather than proposing a new direction.

**Additional context**
- Pointer: `optiland/visualization/system/lens.py:35`
- Good first issue: touches a single file, no public API signature changes (keeps within the project's API-freeze policy), and is testable in isolation with a small `Optic` built with a deliberately overlapping lens.

alexsmolya:
Resolves #731

### Summary
This pull request implements the surface overlap warning in the lens visualization subsystem (`Lens2D` and `Lens3D`), addressing the existing TODO marker in `optiland/visualization/system/lens.py`.

### Technical Details
- **Geometry-Aware Overlap Evaluation**: Rather than a naive center thickness or scalar radius comparison, the check evaluates the actual surface sag profiles across the full semi-diameter clear aperture.
- **Coordinate Invariance**: Coordinates on surface 2 are transformed to global coordinates and then localized into surface 1's coordinate system, accounting for non-zero center thickness, 3D translations, and rotations.
- **Physical Criterion**: Checks if the physical axial separation $\Delta z = z_2 - z_1$ along the local optical axis is $\le 0$ anywhere across the sampled aperture extent ($r \in [0, \max(e_1, e_2)]$).
- **Warning**: Emits a standard `UserWarning("Lens surfaces overlap.")` with `stacklevel=2` when physical overlap or zero thickness is detected.
- **API Freeze Compliance**: No changes to public API signatures or behaviors for valid prescriptions.

### Tests & Validation
- Added `tests/visualization/system/test_lens_overlap.py` with 10 focused regression tests covering:
  - Overlapping and valid biconvex singlets
  - Overlapping and valid steep meniscus lenses
  - Plano-concave and convex-plano configurations
  - Zero-thickness elements
  - Aspheric lenses (`EvenAsphere`)
  - 3D visualization (`OpticViewer3D` / `Lens3D`)
  - Empty / single-surface boundary cases
- Scoped test suite `pytest tests/visualization/ tests/test_optic.py` passes (133/133 tests).
- `ruff check optiland/` and `ruff format --check optiland/` pass cleanly.

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -47,6 +47,9 @@ We sincerely appreciate the contributions of the following individuals, whose ef
 - **Sanjay Santhanam** (`GitHub <https://github.com/Sanjays2402>`__)
 - **Ashish Verma** (`GitHub <https://github.com/ashishv-git>`__)
 - **Anthony Beaucamp** (`GitHub <https://github.com/8bit-Dude>`__)
+- **Alex Smolya** (`GitHub <https://github.com/alexsmolya>`__)
+- **Ayush Sharma** (`GitHub <https://github.com/Aeirx>`__)
+- **breezeFur** (`GitHub <https://github.com/breezeFur>`__)
 
 
 Your contributions, whether in the form of code, documentation, feedback, or discussions, are what make **Optiland** better for everyone.
--- a/optiland/visualization/system/lens.py
+++ b/optiland/visualization/system/lens.py
@@ -8,6 +8,7 @@ Kramer Harrison, 2024
 from __future__ import annotations
 
 import warnings
+from typing import Any
 
 import matplotlib.pyplot as plt
 import numpy as np
@@ -16,6 +17,15 @@ from matplotlib.patches import Polygon
 
 import optiland.backend as be
 from optiland.visualization.system.utils import revolve_contour, transform, transform_3d
+
+
+def _to_float(val: Any, default: float = 0.0) -> float:
+    """Convert a scalar or 0-d backend array to a Python float."""
+    try:
+        arr = be.to_numpy(val)
+        return float(arr.item() if hasattr(arr, "item") else arr)
+    except Exception:
+        return default
 
 
 class Lens2D:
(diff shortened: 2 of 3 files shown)


HarrisonKramer:
Thanks for closing out the TODO here, nice work.

A few notes:

1. **Scope**: this only checks consecutive surfaces *within* a single lens element (each `Lens2D` is one glass group per `system.py::_identify_components`), matching what #731 asked for. It won't catch overlap across an air gap between two separate elements (e.g. a doublet colliding with the next singlet). Not asking to expand scope here, just flagging it as a gap we should track: either a follow-up PR (a system-level check would probably be the cleaner home for it) or folded into this one if you'd rather.
2. `_surfaces_overlap` only samples along the X/Y cross-sections (`pts` in the sampling loop). For non-rotationally-symmetric surfaces (Zernike, biconic, freeform) that can miss a real overlap that only shows up off-axis. Worth a docstring note on this limitation at minimum.
3. Minor: `_check_surface_overlap`/`_surfaces_overlap` are missing type hints (repo convention is mandatory hints on new signatures).
4. Minor: the `float(be.to_numpy(x).item()) if hasattr(be.to_numpy(x), "item") else float(x)` pattern calls `to_numpy` twice, and the `else` branch looks unreachable since `to_numpy` always returns an ndarray. Could simplify to a small `_to_float` helper.

I'm okay merging this as-is, but let me know if you'd like to make any changes first.

Best,
Kramer

alexsmolya:
Thanks for the review, Kramer!

I addressed the feedback in the latest commit:
- added the missing type hints;
- simplified scalar conversion with a `_to_float` helper;
- documented the X/Y cross-section sampling limitation for non-rotationally-symmetric surfaces;
- added focused regression tests for the overlap helpers and edge cases.

I kept the scope limited to #731, so cross-element/air-gap overlap detection remains a potential follow-up.

Locally, 137 tests pass along with ruff and mypy.

Thanks again!

PR #736, opened 2026-08-15
feat(paraxial): add EFL and ray-transfer matrix for a surface range

### LINKED ISSUE #732: Add helper to compute effective focal length (EFL) for a sub-range of surfaces
### Feature Request

Raised in Discussion #471: `optic.paraxial.f2()` returns EFL for the entire system, but there's no way to get the EFL of just a sub-group of surfaces (e.g. "lens 1" within a multi-lens assembly). This is useful when characterizing individual groups within a larger design.

A paraxial helper that computes EFL using only a specified contiguous surface range, by building the equivalent ray-transfer matrix for that sub-range instead of the full system.

N/A.

**Additional context**
The original requester noted that after optimization, a combined system doesn't decompose cleanly back into constituent-lens focal lengths. So this should be documented explicitly as the paraxial EFL of the surface range treated in isolation (its own conjugates), not a decomposition of the full system's power. This requires solid ray-transfer-matrix understanding, so it's tagged `help wanted` rather than `good first issue`.

- Discussion: #471

Aeirx:
Closes #732.

Adds a paraxial helper that returns the effective focal length of a contiguous
range of surfaces, so a lens group inside a larger design can be characterised
on its own.

### API

```python
lens = CookeTriplet()
lens.paraxial.f2_range(1, 2)            # EFL of the first element
lens.paraxial.ray_transfer_matrix(1, 2) # the underlying 2x2 ABCD matrix
```

Both indices are inclusive, and both methods take an optional `wavelength`
that defaults to the system's primary wavelength.

### Approach

`ray_transfer_matrix(start, end)` builds the equivalent ABCD matrix for the
range by composing the per-surface interaction and transfer matrices. Those
matrices are transcribed directly from the update rules in
`ParaxialRayTracer.trace_generic`, including the reflective and `paraxial`
(ideal lens) branches, so the matrix and an explicit trace cannot drift apart:

| step | matrix |
| --- | --- |
| refraction | `[[1, 0], [-phi_k / n_k, n_{k-1} / n_k]]` |
| reflection | `[[1, 0], [-2 / R_k, -1]]` |
| transfer   | `[[1, t], [0, 1]]` |

`f2_range()` then reads the EFL off that matrix as `-1 / C`, which is the same
ratio `f2()` forms from a ray launched parallel to the axis.

The matrices are assembled with `be.stack` rather than by index assignment, so
the result stays differentiable under the torch backend.

### On the "in isolation" caveat

Per the issue, the docstring states explicitly that this is the paraxial EFL of
the range **with its own conjugates**, and not a decomposition of the full
system's power. `test_f2_range_is_not_a_decomposition_of_system_power` asserts
that the group powers of a Cooke triplet genuinely do *not* sum to the system
power (0.0032 vs 0.0291), so the caveat cannot quietly stop being true.

### Tests

`tests/test_paraxial.py`, all on both the numpy and torch backends:

- `f2_range(1, last)` reproduces `f2()` across every sample system in
  `get_optic_data()` — this is the main correctness check, and it covers the
  mirror path via `HubbleTelescope`.
- a lone thick element, and an element embedded in a Cooke triplet, both match
  the analytic Jenkins & White thick-lens EFL already used elsewhere in the file.
- a single refracting surface gives `n' / phi`.
- `det(M) == n_before / n_after` for a refractive range.
- splitting a range and re-composing it reproduces the whole matrix.
- ideal (`paraxial`) surfaces, an afocal range returning `inf`, the `wavelength`
  argument, and the `ValueError`s for invalid ranges.

Locally the paraxial suite is 2128 passed / 1 skipped across both backends
(the skip is the autodiff test on numpy), and `ruff check` / `ruff format` are
clean.

### Unrelated pre-existing failures

While running the full suite I hit 5 failures in `tests/test_ray_aiming.py`
(`test_epd_invariant_under_translation` and
`test_float_by_stop_epd_invariant_under_translation`, numpy backend). They
reproduce identically on a clean `master` with my changes stashed, so they are
not from this PR — but flagging them in case they aren't already known.

### Notes for review

- I made `ray_transfer_matrix` public because the matrix is useful beyond EFL
  (group principal planes, for instance), but I'm happy to make it private if
  you'd rather keep the surface area smaller.
- I kept this to the paraxial helper. If an `EFLRange`-style optimization
  operand would be welcome — a couple of people in #471 mentioned wanting to
  constrain group focal lengths — I'm glad to add it in a follow-up.

### DIFF
--- a/docs/cheat_sheet.rst
+++ b/docs/cheat_sheet.rst
@@ -198,6 +198,17 @@ New to these concepts? See the :ref:`glossary` first.
    print("XPL:", lens.paraxial.XPL())   # exit pupil location, relative to the image surface
    print("Magnification:", lens.paraxial.magnification())
 
+   # EFL of a lens group, i.e. surfaces 1 through 2 only
+   print("Group EFL:", lens.paraxial.f2_range(1, 2))
+
+.. note::
+
+   ``f2_range(start, end)`` treats the surface range as a lens group **in
+   isolation**, with its own conjugates. It is not a decomposition of the
+   system's power, so the group focal lengths of a design will not add back up
+   to ``f2()``. The underlying 2x2 ray-transfer matrix of a surface range is
+   available as ``lens.paraxial.ray_transfer_matrix(start, end)``.
+
 .. note::
 
    ``EPL()`` is measured **relative to the first physical surface** (surface 1),
--- a/optiland/paraxial.py
+++ b/optiland/paraxial.py
@@ -86,6 +86,49 @@ class Paraxial:
         f2 = -y[0] / u[-1]
         return f2[0]
 
+    def f2_range(
+        self,
+        start: int,
+        end: int,
+        wavelength: float | None = None,
+    ) -> ScalarOrArray:
+        """Calculate the effective focal length of a range of surfaces.
+
+        The range is treated as a lens group in isolation: the returned value
+        is the paraxial EFL that surfaces ``start`` through ``end`` have with
+        their own conjugates, evaluated in the media that bound the range in
+        the parent system.
+
+        This is deliberately not a decomposition of the full system's power.
+        The focal lengths of a system's groups do not, in general, recombine
+        into :meth:`f2`, since that would additionally require the separations
+        between the groups' principal planes.
+
+        Args:
+            start: Index of the first surface of the range, inclusive.
+            end: Index of the last surface of the range, inclusive.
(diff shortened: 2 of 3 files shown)


HarrisonKramer:
Hi @Aeirx,

Thanks for adding this! Implementation looks good, and I agree with the approach taken. CI passes, so the failures you mentioned must be related to something else. I'll keep an eye on that. Merging now.

Best,
Kramer

PR #749, opened 2026-08-20
perf(backend): keep reductions on the device and skip autograd bookkeeping in forward traces

manuelFragata:
Second batch of the performance pass that started with #748, this time targeting data that was moved between GPU and host for no reason.

- **`be.max` / `be.min` copied the entire tensor to the CPU** to compute a single scalar. They now reduce on the device and transfer one value. This ran on the full ray bundle once per trace of every Newton–Raphson surface (aspheres, freeforms).
- **`be.all` / `be.any` cast boolean masks to the working float precision** before reducing — a full-size temporary allocation and several times the bandwidth, a few times per surface and once per Newton iteration in the convergence check. They now reduce in the mask's own dtype, and plain host data (lists, numpy arrays) reduces on the host instead of being shipped to the GPU.
- **The real ray tracer now runs under `torch.no_grad` when gradients are globally disabled**, sparing per-op autograd bookkeeping across the hundreds of kernels a trace launches. With `grad_mode` enabled the context is a no-op, so differentiable tracing is untouched — a new test pins both behaviors.

It also silences a scalar-conversion warning from #748 when the wavelength bundle is grad-attached.

Measured effect (float64, forward traces, single wavelength, RTX 5070, torch 2.13, medians of 5): Newton–Raphson systems get consistently faster across the sweep —

| rays | before | after | |
|---|---|---|---|
| 250k | 21.6 ms | 16.8 ms | −22% |
| 1M | 31.2 ms | 24.9 ms | −20% |
| 8M | 253 ms | 205 ms | −19% |

Forward trace latency, even-asphere singlet, before vs after

Closed-form systems (Cooke triplet) are unchanged within run-to-run noise: their remaining per-trace overhead is dominated by repeated paraxial entrance-pupil computation and host syncs on static surface properties, which are the next items from the same audit.

The full suite passes on both backends (8334 tests).

### DIFF
--- a/optiland/backend/base.py
+++ b/optiland/backend/base.py
@@ -6,11 +6,12 @@ Kramer Harrison, 2025
 
 from __future__ import annotations
 
+import contextlib
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any, Literal
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Sequence
+    from collections.abc import Callable, Generator, Sequence
     from types import ModuleType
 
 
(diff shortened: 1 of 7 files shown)


manuelFragata:
@HarrisonKramer  you might want to review this one

HarrisonKramer (approved):

HarrisonKramer:
Hey @manuelFragata, sorry for the delay. PR looks good! and I appreciate the benchmarking. All is clear.

Merging now.

Kramer

PR #752, opened 2026-08-23
DOC Fix custom surface normal derivatives

### LINKED ISSUE #747: Incorrect Surface Normal Formula in Custom Surface Types Tutorial
Hi, I appreciate the work and detailed tutorial! I went through the tutorial these days and found out an error. The Custom Surface Types tutorial [link](https://www.optiland.org/tutorials/extending-surfaces) section 2 "Define the Custom Geometry Class" has an incorrect analytic surface-normal expression.

For a radial sag of the form $z(x,y)=ar$, where $r=\sqrt{x^2+y^2}$, the derivatives should be

$\frac{\partial z}{\partial x}=a\frac{x}{r}$

and

$\frac{\partial z}{\partial y}=a\frac{y}{r}$.

However, the tutorial uses terms of the form $a x / r^2$, whereas the expected expression is $a x / r$, with $r=\sqrt{x^2+y^2}$. The same correction applies to the $y$ component. This produces an incorrectly surface normal and thus a ray-tracing physics error.

**Suggested Improvements**  
Correct the derivative from $ax/r^2$ and $ay/r^2$ to $ax/r$ and $ay/r$, and add a brief note on handling the singular point at $r=0$. Rerun the following ray tracing.

Thank you!

prabandkumar:
Fixes #747

The custom surface types tutorial used r² in the denominator
for the derivatives of the a*r sag term

The derivatives should be

dz/dx = a*x/r
dz/dy = a*y/r

This updates both expressions in
Tutorial_8a_Custom_Surface_Types.ipynb.

The correction was also verified numerically against the
derivative of a*sqrt(x²+y²)


HarrisonKramer:
Hi @prabandkumar,

Thanks for fixing this!

I see two things that also need to be changed. First, the reported markdown equation is also wrong and should be $ar$ , but it still shows $ar^2$. Second, the screenshot at the end of the tutorial is no longer up-to-date after this. Would you like to add a new one with the latest config? I can also help there - just let me know.

Regards,
Kramer

prabandkumar:
Hi @HarrisonKramer ,

Thanks for pointing these out. I'll fix the equation and would appreciate your help with updating the screenshot.

Thanks!

prabandkumar:
Hi @HarrisonKramer ,
I tried regenerating the 3D screenshot in Codespaces, but lens.draw3D() opens a separate VTK window and the kernel eventually crashes there. Could you help me with generating the updated screenshot?

Thanks!

PR #753, opened 2026-08-25
Fix wrong-branch conic intersections for off-axis parabolic mirrors

### LINKED ISSUE #329: Ray Tracing Issue with Off-Axis Parabolic Mirrors in 4F System
Hello to all and big thanks to @HarrisonKramer and the other project contributors for this awesome software!

### Checklist

### Question

**What is your question?**  

Straight to the point. I'm trying to build a 4F system with two off-axis parabolic mirrors, but I'm running into an issue where the input rays don't trace unless I introduce an extra rotation on the second mirror, which misaligns it from the proper position.

In the figure, I show this situation. For rx = 0, half of the rays don't trace, and we can trace more rays as we increase the (ccw) rotation angle on the x-axis of the second mirror. For rx = -10º, all rays trace.

* I tried replacing the second mirror with a flat one, and there’s no issue with any angle.

* I tested two different ways of generating the second mirror (one with inverted curvature and another identical to the first mirror with rz and ry rotations applied; both options are in the attached code, one commented out), and the behavior is the same in both cases.

I don’t know if this is a bug or if it’s supposed to behave this way.

Thanks in advance!.

**Relevant Context**  
Here is the figure with diferent rotations of the second mirror:

And here is the code:

```
import numpy as np
from optiland.physical_apertures import OffsetRadialAperture

class parabolic_double(Optic):
   
   def __init__(self):
      super().__init__()  
      
      
      CURV = 1.312335958005249499E-02
      R = 1/CURV # 76.2

      aperturepm1 = OffsetRadialAperture(r_min = 0,r_max=25.4, offset_y= 76.2)
      aperturepm2 = OffsetRadialAperture(r_min = 0,r_max=25.4, offset_y= -76.2)

      rx_angle = -0

      self.add_surface(index = 0, thickness = np.inf)
   
      self.add_surface(
         index = 1, 
         radius = np.inf, 
         thickness =152.4, 
         material = 'air', 
         semi_aperture = 22.86)
      #First parabolic mirror, thickness 2 times  parent focal, the vertex of the next optical element
      self.add_surface(
         index = 2, 
         radius = -R, 
         thickness = -38.1*2,
         material = 'mirror', 
         is_stop= True,
         conic = -1, 
         dy = -76.2,
         aperture = aperturepm1,
      )

      """
      #adding a flat mirror to check reflections
      self.add_surface(index = 3,
                       thickness = 150,
                       material = 'mirror',
                       aperture = 100,
                       dy = -76.2*2,
                       rx = np.radians(-45+15))
      """
      
      #Second parabolic mirror, reverse radius, aperturepm2 with offset -reflected_focal_lens
      self.add_surface(
         index = 3, 
         radius = R, 
         thickness = 38.1*3, 
         material = 'mirror', 
         is_stop= False,
         conic = -1, 
         dy = -76.2,
         aperture = aperturepm2,
         rx = np.deg2rad(rx_angle),
      ) 
      
      """
      # Second parabolic mirror, same as the first, 180º rotation over y and z
      self.add_surface(
         index = 3, 
         radius = -R, 
         thickness = 38.1*2, 
         material = 'mirror', 
         is_stop= False,
         conic = -1, 
         dy = -76.2,
         aperture = aperturepm1,
         ry = np.deg2rad(180),
         rz = np.deg2rad(180),
         rx = np.deg2rad(rx_angle)
      ) 
      """

      self.add_surface(index = 4, 
                       #radius = np.inf,
                       thickness = 50, 
                       material ='air',
                       aperture = 100,
                       dy = -76.2*2,
                       )

system = parabolic_double()
system.set_aperture(aperture_type='EPD', value = 15.0)
system.set_field_type(field_type = 'angle')
system.add_field(y = 0)
system.add_wavelength( value = 0.4861, is_primary = True)
system.draw3D()

manuelFragata:
## Summary

Fixes the off-axis parabolic mirror tracing failures reported in #329: in the 4F double-OAP relay about half the fan disappeared at rx = 0 (and came back as the second mirror was tilted away from its proper position), and in the 90° OAP collimator half the beam left the mirror away from the screen.

The line-conic intersection is a quadratic with two candidate hit points, and the tracer kept whichever lies closest to the surface vertex (plus an entry-side check from #652 that only ran when a batch contained near-grazing rays). For off-axis conic sections that guess goes wrong in two ways. Both candidates can be genuine forward hits on the parent paraboloid, with the correct one — on the used, in-aperture section — sitting farther from the vertex, so the ray is sent to a part of the surface that isn't part of the physical mirror and dies on the aperture. Or the vertex-nearest candidate lies *behind* the ray entirely, which "reflects" it off a point it never reached and sends it the wrong way.

Root selection now keeps only candidates that are in front of the ray and on the sheet the sag function actually describes (the quadric also contains points the sag never produces, like the far side of a sphere or the detached second sheet of a k < -1 hyperboloid), prefers candidates landing inside the surface's physical aperture, and takes the nearest of those. Rays with no admissible candidate keep the legacy vertex-nearest fallback (NaN on a true miss), so genuine misses behave exactly as before. This is the same rule the non-sequential tracer already uses (`_root_valid` in `optiland/nonsequential/components/geometry/analytic/conic.py`) — it brings the sequential path in line with NSQ.

Implementation notes:

- The rule lives in a shared `_conic_intersection_distance` helper in `optiland/geometries/standard.py`. Selection is per ray and fully vectorized (`be.where` only — the batch-level `bool(be.all(|N| > 1e-2))` gate is gone, so results no longer depend on batch composition), and denominators and the radicand are guarded in place so no inf/NaN enters the autograd graph.
- Surfaces hand their physical aperture to geometries whose `distance()` accepts an optional `aperture` keyword (a cached signature check in a module-level helper), so custom geometries with the plain `distance(rays)` signature and `SurfaceView` proxies keep working untouched. If you'd rather keep apertures out of the geometry layer entirely, the alternative is moving the root choice up into the surface trace step — happy to rework it that way instead.
- The Newton-Raphson seed forwards the aperture, so aspheric intersections start in the basin of the physical root (which the implicit gradients from #724 assume).
- `StandardGratingGeometry.distance` now delegates to the same helper — it still carried the pre-#648 numerically unstable textbook formula and no branch disambiguation at all.
- The unused `_intersection` / `_intersection_plane` / `_intersection_sphere` helpers in `newton_raphson.py` are removed.

**Before** (master, the two #329 systems): 20/41 rays survive the 4F relay at rx = 0, recovering to 41/41 only when the second mirror is misaligned by -10°; 10-11 of 21 collimator rays per field, the rest leaving the mirror backwards.

**After**: 41/41 at every tilt and 21/21 per field with zero misrouted rays — while Cooke triplet, double Gauss, Hubble, and the UV Schwarzschild microscope trace bit-for-bit identically to master (217-ray hexapolar comparison per field), and the golden regression snapshots and the pinned Hubble OPD operand pass unchanged.

## Test plan

- [x] New `tests/test_conic_root_selection.py` (24 tests / 56 parametrized instances, numpy + torch): analytic pins of the forward/backward root choice and the aperture tier, sheet-test rejection of a sphere's far hemisphere and of a k < -1 phantom sheet (including an adversarial annular aperture that contains only the phantom), a ray starting exactly on the surface (t = 0 never returned), the true-miss NaN contract, bitwise batch-composition independence, grating-vs-standard bitwise agreement plus the near-degenerate-a stability sweep on the grating, the Newton-Raphson seed basin, torch autograd safety with exact analytic gradients, and both #329 systems end to end (the relay re-collimates to ~2e-15 direction spread; the collimator to ~4e-16, with the beam landing inside the screen aperture)
- [x] Both reproducer scripts from the issue thread trace fully
- [x] Full test suite: 8329 passed, 0 failed on numpy and torch (cpu, float64, grad enabled)
- [x] `uvx ruff check .` and `uvx ruff format --check .` clean

Fixes #329


BrunoChevalier:
@manuelFragata I have some performance improvements ready on top of your changes in this PR that fix the performance regression. Is it fine if I start a pull request on top of your branch so you can review them?

manuelFragata:
@BrunoChevalier yes please, go ahead

PR #756, opened 2026-08-26
ci: relative performance-regression gate

manuelFragata:
## Problem

Performance regressions land silently: the test suite says nothing about
speed, and stored benchmark baselines are meaningless because every
developer and every CI runner is a different machine.

## What this adds

`benchmarks/perf_regression.py` and a `performance` CI job. The tool never
stores a baseline. It measures two refs back to back on the same machine,
the merge base with master and the candidate (in PR CI, the would-be merge
result), each in a fresh interpreter, interleaved to cancel thermal and
load drift, and compares the per-scenario minimum over five repeats.

A scenario fails the gate only when it is more than 5 percent slower and
at least 2 ms slower. The 5 percent is a zero-regression policy expressed
with margin: an A/A run of the same ref against itself measures ratios up
to about 1.015 with this protocol, so 1.05 is three times the observed
noise floor. Anything above it is a real slowdown, not jitter. The floor
can be recalibrated on any machine with
`--base <ref> --candidate <ref>` using the same ref twice.

Five scenarios cover the current hot paths: explicit-ray propagation
through a spherical stack and through an off-axis conic, full
`optic.trace` generation, iterative ray aiming, and repeated scalar
first-order queries.

Developers can run the same comparison locally:

    uv run python benchmarks/perf_regression.py            # vs merge base
    uv run python benchmarks/perf_regression.py --base v0.6.0

Both refs run from plain source checkouts against the active environment,
so the tool works on any machine without setup. One limitation: the base
ref runs against the candidate's dependency set, so a PR that changes
dependencies compares apples to oranges and should say so.

## Evidence it works

Run against the two currently open PRs, the gate flagged real slowdowns in
both (#729: per-call first-order path construction and doubled Jacobian
probe traces; #753: root-admissibility passes in the conic intersection),
which were then partly optimized on those branches. An A/A run passes
cleanly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,3 +112,19 @@ jobs:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup
       - run: uv build
+
+  performance:
+    # Relative regression gate: measures the PR's would-be merge result
+    # against its merge base on the same runner and fails on a large
+    # slowdown. No stored baselines, so results are machine-independent.
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    needs: [lock_file]
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/setup
+      - run: uv sync
+      - name: Compare against merge base
+        run: uv run python benchmarks/perf_regression.py
--- a/benchmarks/perf_regression.py
+++ b/benchmarks/perf_regression.py
@@ -257,6 +257,7 @@ def _run_child(source_dir: Path, out_file: Path) -> None:
 
 
 def compare(args: argparse.Namespace) -> int:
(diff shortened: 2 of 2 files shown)


BrunoChevalier:
@manuelFragata @HarrisonKramer 

What is the plan for https://github.com/optiland/optiland/pull/753 ? My design needs that fix and I currently have to keep adding that one on top of my own changes 😄. 
Does that first need this gating CI test to be merged and then a rework that addresses the speed regression?

manuelFragata:
> @manuelFragata @HarrisonKramer
> 
> What is the plan for #753 ? My design needs that fix and I currently have to keep adding that one on top of my own changes 😄. Does that first need this gating CI test to be merged and then a rework that addresses the speed regression?

Hi there @BrunoChevalier ,

we have been a bit busy with other things, sorry. I will have a look into it right now and get it finalized. In a couple of hours should be done.

PR #765, opened 2026-09-01
fix: align Torch histogram2d bin edge semantics

### LINKED ISSUE #762: Torch histogram2d bins samples on internal edges differently from NumPy
### Checklist

- [x] I have searched the existing issues and discussions for a similar problem.
- [x] I am using the latest version of Optiland.
- [x] I have tried to reproduce and debug the issue.
- [x] I have included a minimal reproducible example.

### Bug Report

**Describe the bug**

The Torch backend implementation of `histogram2d` assigns samples that lie
exactly on an internal bin edge to the previous bin. This differs from
`numpy.histogram2d`, where bins other than the last are half-open and an
internal-edge sample belongs to the bin on its right.

The current implementation uses:

```python
torch.searchsorted(edges, values, right=False) - 1
```

For an internal edge, this produces the index of the preceding bin.

**To Reproduce**

```python
import optiland.backend as be

    be.set_backend(backend)

    x = be.array([1.0])
    y = be.array([1.0])
    edges = be.array([0.0, 1.0, 2.0])

    histogram, _, _ = be.histogram2d(
        x,
        y,
        bins=[edges, edges],
    )

    print(backend, be.to_numpy(histogram))
```

Observed result:

```text
numpy [[0. 0.]
       [0. 1.]]

torch [[1. 0.]
       [0. 0.]]
```

**Expected behavior**

Both backends should follow NumPy histogram semantics. The sample `(1.0, 1.0)`
should be assigned to bin `(1, 1)` for edges `[0.0, 1.0, 2.0]`.

Boundary-focused tests should cover the lower edge, internal edges, and the
included uppermost edge for both backends.

**Screenshots**

Not applicable.

**Environment**

- Optiland: current `master` at `b9f61dabce896cec885752e7fbbcce21d304d6f1`
- Python: `3.13.13`
- NumPy: `2.3.5`
- PyTorch: `2.13.0+cpu`
- OS: Windows 11

**Additional context**

`histogram2d` is used by radiant-intensity and irradiance analyses, so this
backend difference can place power into the wrong angular or spatial bin when
a ray lies exactly on a bin boundary.

I would be happy to submit a focused PR with the Torch backend correction and
backend-parametrized regression tests if the proposed behavior is confirmed.

somnus-J-307:
## Summary

- align the Torch backend's internal bin-edge assignment with NumPy's
  `histogram2d` semantics
- extend the backend-parametrized regression test to cover lower, internal,
  upper, and out-of-range samples

## Testing

- `tests/test_backend_functions.py`: 63 passed, 1 skipped
- `TestIncoherentIrradiance` and `TestRadiantIntensity`: 37 passed, 3 skipped
- `ruff check .`
- `ruff format --check .`

Fixes #762

### DIFF
--- a/optiland/backend/torch_backend/misc.py
+++ b/optiland/backend/torch_backend/misc.py
@@ -184,8 +184,9 @@ class MiscMixin:
         nx = x_edges.numel() - 1
         ny = y_edges.numel() - 1
 
-        x_bin_indices = torch.searchsorted(x_edges, x, right=False) - 1
-        y_bin_indices = torch.searchsorted(y_edges, y, right=False) - 1
+        # Match NumPy: internal edges belong to the bin on their right.
+        x_bin_indices = torch.searchsorted(x_edges, x, right=True) - 1
+        y_bin_indices = torch.searchsorted(y_edges, y, right=True) - 1
         x_bin_indices = torch.clamp(x_bin_indices, 0, nx - 1)
         y_bin_indices = torch.clamp(y_bin_indices, 0, ny - 1)
 
--- a/tests/test_backend_functions.py
+++ b/tests/test_backend_functions.py
@@ -175,14 +175,13 @@ def test_cross_product(set_test_backend):
 
 def test_histogram2d(set_test_backend):
     if hasattr(be, "histogram2d"):
-        x = be.array([0.5, 1.5])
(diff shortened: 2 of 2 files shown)


manuelFragata:
hi @somnus-J-307 ,

thanks for the fix. merging now. 

manuel

PR #776, opened 2026-09-04
fix: preserve bilinear edge contributions in NumPy grid_sample

### LINKED ISSUE #775: NumPy grid_sample zero padding drops bilinear contributions near image boundaries
### Checklist

- [x] I searched existing issues, PRs, and discussions for related reports.
- [x] I reviewed the backend API docstrings and SciPy's boundary-mode documentation.
- [x] I checked the affected implementation against current upstream master (details below).
- [x] I reproduced and investigated the issue locally.
- [x] I included a minimal reproduction and environment details.

### Bug Report

**Describe the bug**

The NumPy implementation of `be.grid_sample` drops valid bilinear contributions near image boundaries when `padding_mode="zeros"`. With `align_corners=False`, the normalized boundary coordinates map half a pixel beyond the outermost pixel centers. NumPy returns zero there, whereas the Torch backend correctly interpolates between the edge pixels and the zero padding.

**To Reproduce**

```python
import numpy as np
import optiland.backend as be

try:
        be.set_backend(backend)
        if backend == "torch":
            be.set_device("cpu")
            be.set_precision("float64")

        image = be.array(np.ones((1, 1, 2, 2)))
        grid = be.array(
            [[[[-1.0, -1.0], [-1.0, 0.0], [0.0, 0.0], [1.0, 1.0]]]]
        )
        result = be.grid_sample(
            image,
            grid,
            mode="bilinear",
            padding_mode="zeros",
            align_corners=False,
        )
        print(backend, be.to_numpy(result).ravel().tolist())
finally:
    be.set_backend("numpy")
```

Actual output:

```text
numpy [0.0, 0.0, 1.0, 0.0]
torch [0.25, 0.5, 1.0, 0.25]
```

**Expected behavior**

Both backends should return `[0.25, 0.5, 1.0, 0.25]`. For the all-ones image, the corner receives one quarter of an edge pixel's contribution, and the left-edge midpoint receives one half. The image-center result is unchanged.

As a control, the same queries with `align_corners=True` return `[1.0, 1.0, 1.0, 1.0]` in both backends.

**Environment**

- Windows, CPU
- Python 3.13.13
- NumPy 2.3.5
- SciPy 1.16.3
- PyTorch 2.13.0+cpu
- Reproduced on checkout `f925360c28858b5df3aaa68919ff3bee34052614` (PR #705 branch after merging upstream). The affected NumPy backend file is unchanged from current upstream `master@d67345204d9a62424a9bcad25773e9d09b2d2913`; both have Git blob `bc75f43d3e147fa36ba28f7c828aced9d1d9ba1a`.

**Additional context**

The [NumPy implementation](https://github.com/optiland/optiland/blob/d67345204d9a62424a9bcad25773e9d09b2d2913/optiland/backend/numpy_backend/interpolation.py) currently uses `scipy.ndimage.map_coordinates(..., mode="constant", cval=0.0)`. SciPy documents that `constant` does not interpolate beyond the input extent, while `grid-constant` does. Calling the latter directly on the same pixel coordinates produced `[0.25, 0.5, 1.0, 0.25]` in a local experiment: [SciPy documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.map_coordinates.html).

This sampling configuration is used by [DistortionWarper.warp_image](https://github.com/optiland/optiland/blob/d67345204d9a62424a9bcad25773e9d09b2d2913/optiland/analysis/image_simulation/distortion_warper.py), so the mismatch can affect image-warping boundary brightness. The existing `test_grid_sample` checks output shape but not boundary values. This is separate from the oversampling/autograd fix in #735.

I would like to prepare a scoped fix for bilinear zero-padding semantics, with numerical boundary tests, NumPy/Torch parity checks, and an image-warping regression. Using `grid-constant` is a candidate approach; other sampling and padding modes should be checked for regressions before submitting the patch.

somnus-J-307:
Fixes #775.

### Summary

- Use SciPy's `grid-constant` mode for bilinear zero-padding so partially in-bounds samples retain their edge-pixel contributions, matching Torch.
- Leave nearest-neighbor interpolation and border/reflection padding unchanged.
- Add analytical boundary checks, float32/float64 parity tests for both `align_corners` settings, batched/multichannel and singleton-dimension coverage, and `DistortionWarper` boundary-brightness regressions.

### Validation

- The new regression file produced 17 failures and 27 passes before the fix.
- `pytest tests/test_grid_sample_boundaries.py tests/test_backend_functions.py tests/test_backend_contract.py tests/test_image_simulation.py tests/test_interpolators_phase.py tests/regression -q --tb=short -p no:cacheprovider`: 317 passed, 2 skipped.
- `ruff check --no-cache .` and `ruff format --check --no-cache .`: passed.
- Explicit Ruff lint/format checks for the new test file: passed.
- `git diff --check`: passed.

Validated locally on Windows/CPU with both NumPy and Torch. Golden fixtures, CI configuration, dependencies, and public signatures are unchanged.

### DIFF
--- a/optiland/backend/numpy_backend/interpolation.py
+++ b/optiland/backend/numpy_backend/interpolation.py
@@ -105,6 +105,10 @@ class InterpolationMixin:
             scipy_mode = "nearest"
         elif padding_mode == "reflection":
             scipy_mode = "reflect"
+        elif padding_mode == "zeros" and mode == "bilinear":
+            # Interpolate edge pixels with zero-valued neighbors outside the
+            # image, rather than dropping their partial contributions.
+            scipy_mode = "grid-constant"
 
         for n in range(N):
             for c in range(C):
--- a/tests/test_grid_sample_boundaries.py
+++ b/tests/test_grid_sample_boundaries.py
@@ -1,0 +1,102 @@
+from __future__ import annotations
+
+import numpy as np
+import pytest
+
+import optiland.backend as be
+from optiland.analysis.image_simulation import DistortionWarper
+from tests.utils import assert_allclose
+
+
+@pytest.mark.parametrize("align_corners", [False, True])
+def test_bilinear_zero_padding_keeps_partial_edge_contributions(
+    set_test_backend, align_corners
+):
+    image = be.array(np.ones((1, 1, 2, 3)))
+    # Pixel coordinates: corner, edge, interior, opposite corner, outside.
+    pixels = np.array([[-0.5, -0.5], [0.0, -0.5], [1.0, 0.5], [2.5, 1.5], [-1.1, 0.5]])
+    if align_corners:
+        normalized = 2 * pixels / np.array([2, 1]) - 1
+    else:
+        normalized = 2 * (pixels + 0.5) / np.array([3, 2]) - 1
+    grid = be.array(normalized[None, None, ...])
+
+    result = be.grid_sample(
+        image, grid, mode="bilinear", padding_mode="zeros", align_corners=align_corners
+    )
+
+    assert_allclose(result, np.array([[[[0.25, 0.5, 1.0, 0.25, 0.0]]]]))
+
+
+@pytest.mark.parametrize("align_corners", [False, True])
+@pytest.mark.parametrize("shape", [(3, 5), (1, 5), (3, 1)])
+@pytest.mark.parametrize("precision", ["float32", "float64"])
+def test_bilinear_zero_padding_matches_torch(
+    set_test_backend, align_corners, shape, precision
+):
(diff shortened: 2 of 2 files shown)


manuelFragata:
Hi @somnus-J-307 ,

thanks for fixing this. 

out of scope here i guess, but the same sweep shows two remaining parity gaps vs torch: `nearest `+ `zeros`, and `reflection `with `align_corners=True` (scipy `reflect ` only matches torch for `align_corners=False`). Worth another issue as follow up; feel free to open it/solve it since you have the setup at hand

merging now, thanks!

Issue #777, opened 2026-09-05
Several Issues with NurbsGeometry


8bit-Dude:
### Checklist

- [x ] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [x ] I have read the [documentation](https://www.optiland.org/docs/) and tried to find an answer there.
- [x ] I am using the **latest version** of Optiland (if not, please update and retry).
- [x ] I have tried to reproduce or debug the issue myself before opening this.
- [x ] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

*Thanks for taking the time to go through this — it really helps us help you!*

### Bug Report

**Describe the bug**

NurbsGeometry — six bugs — a NURBS surface could not be built, evaluated, intersected or fitted correctly

(1) A NURBS surface built without weights cannot be evaluated at all.
(2) The sag/intersection solver is seeded at a corner of the patch and restarts at random.
(3) sag() cannot be called with a scalar.
(4) distance() returns an unsigned distance.
(5) A ray whose root lies off the patch iterates to max_iter every time.
(6) The auto-fit does not produce the surface it was asked to fit

**To Reproduce**
To see the problems, run:    [nurbs_sphere.py](https://github.com/user-attachments/files/31856179/nurbs_sphere.py)
It fits a known sphere as a NURBS and prints, per bug, where the current master fails. When running on the fix (see pull request), everything is clear. 

**Environment**  
- Optiland Version: Github Master
- Python Version: 3.12.12
- OS: Windows
- Additional Info:

**Additional context**
I am claiming this bug, and will push a fix shortly.

HarrisonKramer:
Resolved by PR #778.

PR #778, opened 2026-09-05
NurbsGeometry: six bugs — a NURBS surface could not be built, evaluated, intersected or fitted correctly

8bit-Dude:
(see: https://github.com/optiland/optiland/issues/777)

(1) A NURBS surface built without weights cannot be evaluated at all. 
(2) The sag/intersection solver is seeded at a corner of the patch and restarts at random. 
(3) sag() cannot be called with a scalar.
(4) distance() returns an unsigned distance.
(5) A ray whose root lies off the patch iterates to max_iter every time. 
(6) The auto-fit does not produce the surface it was asked to fit.

To see the problems, run nurbs_sphere.py (see issue 777). It fits a sphere — the one surface whose answer is known in closed form — as a NURBS, and prints for each bug what the old code gave beside the new.

-------------------------------------------------------------------- 
(1) A NURBS surface built without weights cannot be evaluated at all

Description of the BUG:

__init__ classifies the surface (Bezier / R-Bezier / B-Spline / NURBS) and, in each branch, computes whatever the caller left out — weights, a degree, a clamped knot vector — into local variables that are then discarded. self.W, self.p, self.q, self.U, self.V keep the None they were assigned at the top of the constructor.

Passing control points and knots but no weights — the ordinary way to state a B-spline, whose weights are all 1 — leaves self.W = None, and the first evaluation dies inside compute_nurbs_coordinates on W.ndim:

AttributeError: 'NoneType' object has no attribute 'ndim' Every branch except "everything supplied" and the auto-fit path is affected, so a Bezier or R-Bezier surface cannot be built at all, and a B-spline cannot be traced.

Description of FIX:

nurbs_geometry.py:
305-312 — write the locals back after the branch chain, for any field the caller
          did not supply. Nothing else in the constructor changes.

-------------------------------------------------------------------- 
(2) The solver is seeded at a corner of the patch and restarts at random

Description of the BUG:

sag, distance and surface_normal each started Newton at u = v = 0 — a corner of the patch, not a point near the answer — and recovered from a step outside [0,1] by assigning one be.rand() to every ray that had left, judged against a single bundle-wide residual.

On a well-conditioned net this converges anyway. On an ill-conditioned one the answer is a lottery: the same call, on the same surface, with the same arguments, does not have to give the same number twice. Measured on a 32×32 net over 20 draws of the RNG:

sag error vs the sphere	worst landing error
old, 20 draws	2.96e-02 to 1.39e+01 mm	8.58e-01 mm
new	deterministic	9.65e-12 mm
Landing error is the solver's own error — how far the point it settled on is from the point it was asked about. It isolates the solver from the quality of the net, which matters here because a bad net and a bad solve otherwise look alike.

Description of FIX:

nurbs_geometry.py:
779 — new _seed: place each point across the patch by its position in the box
      spanned by the FOUR CORNER control points, which a clamped spline
      interpolates. Measured off the corners rather than the whole net, because a
      least squares run near interpolation leaves interior control points decades
      from the surface they describe, and a box drawn round those seeds every
      point at one edge.
799 — new _newton: one shared iteration that CLAMPS to [0,1] instead of
      restarting at random, and keeps the best (u, v) by residual rather than the
      last one visited -- pinned rays wander along the rim, so the last is not
      the closest.
730, 775 — _corr and _corr_general return one residual PER RAY rather than a
      bundle-wide max, so _newton can tell a ray still closing on its root from
      one the clamp has pinned.
842, 890, 917 — the three call sites use them. distance seeds from where the ray
      crosses the vertex plane; surface_normal seeds from the ray's own x, y,
      its rays being already on the surface.

-------------------------------------------------------------------- 
(3) sag() cannot be called with a scalar

Description of the BUG:

sag opened on shape = x.shape, so any scalar argument raised:

AttributeError: 'float' object has no attribute 'shape' Every other geometry accepts a float. A caller that samples one height at a time — sizing an aperture, say — therefore has every probe fail; and where that failure is caught rather than raised, the surface is silently clipped to zero radius and blocks the whole beam.

Description of FIX:

nurbs_geometry.py:
840 — be.asarray both arguments and take the shape from that, so scalars and
      arrays are both legal and the result returns in the caller's own shape.

-------------------------------------------------------------------- 
(4) distance() returns an unsigned distance

Description of the BUG:

It ended on t = be.sqrt(be.sum((self.get_value(u, v).T - P0) ** 2, axis=1))   # |S - P0|
where every analytic geometry returns a distance signed along the ray. An intersection behind a ray was reported as an equal distance in front of it, so the ray was propagated to the wrong side of itself.

A plain trace marches forward and hardly ever asks about a surface behind it, so the sign never shows. An iterative aimer probes both ways by construction — 38.6% of its queries in one case had the intersection behind the ray — and its backtracking line search then halves the step for ever, hunting a decrease the sign error prevents:

                 rays kept	RMS sag
unsigned	414 / 972	5.36 mm	        diverging
signed	        972 / 972	0.048 mm	converging
The give-away, once looked for, was the query count: 159 809 calls into distance() for one fit, against 1 746 for the same fit on a conic. The aimer was not quietly losing rays, it was thrashing.

Description of FIX:

nurbs_geometry.py:
900 — dot with the ray direction instead of taking the norm:
      t = be.sum((S - P0) * d, axis=1)
868 — while here: the plane pair was built with be.where(mask, M/sqrt(L²+M²), …),
      and be.where evaluates BOTH sides, so the branch a ray did not take still
      divides -- 0/0 for an axial ray. The quotient was always discarded, so this
      was noise rather than a wrong answer; floored so it stops emitting
      RuntimeWarning: invalid value encountered in divide on every trace.

-------------------------------------------------------------------- 
(5) A ray whose root lies off the patch iterates to max_iter every time

Description of the BUG:

With (2) clamping to [0,1], a ray whose root lies outside the patch is pinned on the rim and its residual never reaches tol, so the loop always ran the full max_iter — 100 iterations, for a correct answer, at 25× the necessary cost.

bundle	before	after
rays on the patch	100	1
rays off the patch	100	13

Description of FIX:

nurbs_geometry.py:
799-819 — _newton tracks per-ray improvement and stops once every ray has either
      converged or gone max_stall iterations without improving.
126, 150 — new max_stall parameter, default 12.

geometry_configs.py:
197 — NurbsConfig gains max_stall. Without this field the factory could not pass
      one, so every surface built on a bench was stuck on the constructor's.

geometry_factory.py:
345 — _create_nurbs passes it through.
On the default: with (6) fixed, every net this module fits lands at max_stall = 2 — measured over 24 of them, n_points 24 to 48 and radii 25 to 200. 12 is headroom for control points a caller supplies, which the module cannot vouch for; an ill-conditioned net does not converge steadily, its residual falling in fits, and three quiet iterations are then not evidence a ray is done. The headroom is close to free: a ray on the patch takes one iteration whatever this says, only off-patch rays pay, and a real trace showed no measurable wall-clock difference. Drop it to 3 if you would rather not carry the margin.

-------------------------------------------------------------------- 
(6) The auto-fit does not produce the surface it was asked to fit

Description of the BUG:

_standard_surface samples the conic on a grid and least-squares a control net through it. Three faults in six lines.

(a) The least squares is square. It read P_size_u as the data count and let approximate_surface default the control count to size_u - 1 — n control points from n+1 data points, which is interpolation in disguise. The net that comes back is wildly ill-conditioned:

32×32 net	control z span	max sag error to r = 12
old	−8.25e+10 to 2.25e+09 mm	2.96e-02 mm
new	−0.004 to 3.503 mm	1.03e-08 mm
— for a surface 3.5 mm deep. Four orders of accuracy, and a net that describes its own surface rather than a cancellation of enormous numbers.

(b) A re-fit shrinks the net. self.P_size_u was overwritten with the control count it had just derived, one fewer than it went in with. update_normalization calls fit_surface() on every paraxial pass in auto mode, so the net degrades silently as the optic updates:

fit 1 -> (16, 16)   fit 2 -> (15, 15)   fit 3 -> (14, 14)   fit 4 -> (13, 13)

(c) A fast surface cannot be fitted at all. z = r2 / (radius * (1 + sqrt(1 - (1+k) r2/radius²))) has no real root past the conic's own edge, and the corners of a square box reach there whenever 2·half² > R². The bare square root gives NaN, which the fitter rejects outright:

R = 15 over ±13 (corner r²/R² = 1.50)
  -> ValueError: array must not contain infs or NaNs

Description of FIX:

nurbs_geometry.py:
32  — new FIT_DATA_FACTOR = 2: data points per control point per axis. Two
      settles the conditioning; more buys nothing measurable.
953-982 — the control net keeps the size it was ASKED for and the data grid is
      its own and denser, with ctrlpts_size_u / ctrlpts_size_v passed
      explicitly. Fixes (a), and makes the fit idempotent, which fixes (b). The
      control count is also clamped to degree + 1, which stops n_points_u = 3
      raising ZeroDivisionError and n_points_u = 2 returning a degenerate 2x2.
971-973 — the root is floored at zero, so points past the conic's edge carry the
      rim's own slope on instead of coming back NaN. Fixes (c).
163 — P_size_u is now n_points_u, not n_points_u + 1. See below. Behaviour change worth calling out. n_points_u is documented as "the grid size of control points", but P_size_u was set to n_points_u + 1 and then read as a data count by _standard_surface and as a control count by _plane_surface. The same n_points_u = 16 therefore produced a 16×16 net from one branch and 17×17 from the other. Both now return exactly what the docstring promises, so anyone relying on the old off-by-one gets a net one smaller from the plane branch.

-------------------------------------------------------------------- 
-------------------------------------------------------------------- 
One limitation that remains, deliberately
A fit box wider than the surface it covers is fitted to a clamped extension, because past the conic's edge there is no sag to fit:

R = 50.0 over ±13.0  (corner r²/R² = 0.14):  sag error 2.71e-07 mm
R = 15.0 over ±13.0  (corner r²/R² = 1.50):  sag error 2.23e-02 mm
R =  8.0 over ±10.0  (corner r²/R² = 3.12):  sag error 1.94e-01 mm

This is inherent rather than a defect — but it beats raising, and it is worth knowing that nurbs_norm_x/y set much wider than the conic costs accuracy across the whole patch, not just at the corners.

### DIFF
--- a/optiland/geometries/nurbs/nurbs_geometry.py
+++ b/optiland/geometries/nurbs/nurbs_geometry.py
@@ -25,6 +25,12 @@ from .nurbs_basis_functions import (
 )
 from .nurbs_fitting import approximate_surface
 
+# Data points per control point along each axis when a surface is sampled to be
+# fitted. One apiece makes the least squares square -- interpolation in disguise --
+# and the net that comes back runs to 1e10 mm for a surface a few mm deep. Two
+# settles it and is worth four orders of accuracy; more buys nothing measurable.
+FIT_DATA_FACTOR = 2
+
 
 class NurbsGeometry(BaseGeometry):
     """Creates a NURBS (Non-Uniform Rational Basis Spline) geometry.
@@ -75,6 +81,19 @@ class NurbsGeometry(BaseGeometry):
             are passed.
         tol: The tolerance for Newton-Raphson iteration.
         max_iter: The maximum number of iterations for Newton-Raphson.
+        max_stall: How many iterations a ray may go without improving before it
+            is taken as settled. A root off the patch is pinned by the clamp and
+            never reaches tol; without this it would iterate to max_iter.
+
+            Every net this module fits now lands at 2 -- measured over 24 of
+            them, spanning n_points 24 to 48 and radii 25 to 200. 12 is headroom
+            for control points supplied by a CALLER, which this module cannot
+            vouch for: a badly conditioned net does not converge steadily, its
+            residual falling in fits, and three quiet iterations are then not
+            evidence a ray is done. The headroom is close to free -- a ray ON the
+            patch lands in a single iteration whatever this says, and only a ray
+            off the patch pays, exactly max_stall + 1 against the 100 it used
+            to.
 
     References:
         - The NURBS Book. See references to equations and algorithms throughout
@@ -104,6 +123,7 @@ class NurbsGeometry(BaseGeometry):
         n_points_v=4,
         tol=1e-10,
         max_iter=100,
+        max_stall=12,
     ):
         super().__init__(coordinate_system)
         self.P = be.asarray(control_points) if control_points is not None else None
@@ -127,6 +147,7 @@ class NurbsGeometry(BaseGeometry):
         self.k = conic
         self.tol = tol
         self.max_iter = max_iter
+        self.max_stall = max_stall
 
         self.is_symmetric = False
 
@@ -135,8 +156,12 @@ class NurbsGeometry(BaseGeometry):
         if control_points is None:
             self.is_fitted = True
             self.ndim = 3
-            self.P_size_u = n_points_u + 1
-            self.P_size_v = n_points_v + 1
+            # The CONTROL count, which is what n_points_u is documented to be.
+            # It used to be that plus one, being read as a data count by the
+            # standard-surface fit and as a control count by the plane one -- so
+            # the same n_points_u gave nets a size apart depending on the branch.
+            self.P_size_u = n_points_u
+            self.P_size_v = n_points_v
 
         # Polynomial Bezier surface initialization
         elif (
@@ -276,6 +301,15 @@ class NurbsGeometry(BaseGeometry):
                     )
                 )
             self.ndim = be.shape(control_points)[0]
+
+        # The branches above fill in whatever the caller left out -- weights, a
+        # degree, a clamped knot vector -- as locals. Written back here, or the
+        # surface keeps the None it was constructed with and evaluates to nothing.
+        for name, value in (("W", weights), ("p", u_degree), ("q", v_degree),
+                            ("U", u_knots), ("V", v_knots)):
+            if getattr(self, name) is None and value is not None:
+                setattr(self, name, be.asarray(value)
+                        if name not in ("p", "q") else value)
 
     def set_radius(self, value: float) -> None:
         """Set the radius of curvature.
@@ -664,7 +698,7 @@ class NurbsGeometry(BaseGeometry):
 
         Returns:
             A tuple containing the correction steps for u and v parameters,
-            and the maximum distance between all rays and surface intersection.
+            and the distance between each ray and its surface intersection.
         """
         S_uv = self.get_value(u, v).T
         r = be.stack(
@@ -691,7 +725,9 @@ class NurbsGeometry(BaseGeometry):
         invj = adj / detJ
 
         correction = be.einsum("ijk,ki->ji", invj, r)
-        residual = be.max(be.abs(r))
+        # One residual PER RAY: _newton has to tell a ray still closing on its root
+        # from one the clamp has pinned, and a bundle-wide max hides that
+        residual = be.maximum(be.abs(r[0, :]), be.abs(r[1, :]))
 
         return correction, residual
 
@@ -710,7 +746,7 @@ class NurbsGeometry(BaseGeometry):
             d2: A parameter related to the ray's position.
 
         Returns:
-            A tuple containing the correction steps and the residual.
+            A tuple containing the correction steps and the per-ray residual.
         """
         S_uv = self.get_value(u, v)
         r = be.stack([S_uv[1, :] + d1, S_uv[0, :] + d2], axis=0)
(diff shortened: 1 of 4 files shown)


HarrisonKramer:
Thanks for submitting this PR, @8bit-Dude.

I pushed a ruff format fix to get CI green.

I reviewed the full diff and independently reproduced bugs #1 (weightless B-spline construction), #3 (scalar sag()), and #4 (signed distance()) against both the old and new code. The Newton/seeding rework for #2 and #5 makes sense. per-ray residual tracking is the right fix for a bundle-wide be.max masking pinned rays, and the corner-based seed is a reasonable/cheap way to avoid the old random-restart nondeterminism.

One flag: #6's P_size_u correction is a real behavior change. a surface built via the "plane" branch will now fit a net one control point smaller in each dimension than before, since the two branches previously disagreed on what P_size_u meant. I'm fine with this as-is, but maybe worth making a note in the docs (optional).

(optional) I'd also love to see 4-5 of the repro cases from nurbs_sphere.py (weightless net, scalar sag, distance sign, fit idempotency) turned into real pytest regression tests. Right now nothing in the suite pins any of these six fixes, so a future refactor could silently reintroduce them.

Regards,
Kramer

8bit-Dude:
Thanks for pushing the ruff format fix. I will try to make these checks ahead of pushing from now on.

(optional) I'd also love to see 4-5 of the repro cases from nurbs_sphere.py (weightless net, scalar sag, distance sign, fit idempotency) turned into real pytest regression tests.  -> I added a test suite covering (1)-(6).

(optional) Maybe worth making a note in the docs.  ->  Is there also a page providing instructions for documentation editing?

HarrisonKramer:
Thanks for the follow-up, @8bit-Dude.

And thanks for updating the regression tests!

On docs editing, I don't believe we have specific instructions on how to edit the documentation. That's probably something we should add. In general, any markdown or rst files can simply be edited without issue. The docs build is handled in github actions. I'll make a note here that we ought to improve the contributing guidelines for documentation editing.

That's not a showstopper though, so I'll go ahead and merge this one. Thanks again.

Kramer

PR #783, opened 2026-09-07
Fix typos in the documentation

goelakash:
Fixes 4 spelling mistakes in the docs, found with `codespell`.

Prose only. No code identifiers, module paths, package names or proper nouns changed, and anything that was a valid word rather than a misspelling was left alone.


HarrisonKramer:
Hi @goelakash,

It looks like you found some typos in the YAML files from the refractiveindex.info database. It's not technically written by us, but we can go ahead and merge to correct those references within Optiland.

Thanks!
Kramer

Issue #761, opened 2026-08-31
Error in OPD class docstring: `num_rings` should be `num_rays`


kmdouglass:
### Checklist

- [X] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [X] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [X] I am using the **latest version** of Optiland (if not, please update and retry).
- [X] I have tried to reproduce or debug the issue myself before opening this.
- [X] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

### Bug Report

**Describe the bug**

The docstring for the `OPD` class currently states the following:

```python
class OPD(Wavefront):
    """Represents an Optical Path Difference (OPD) wavefront.

    Args:
        optic (Optic): The optic object.
        field (tuple): The field at which to calculate the OPD.
        wavelength (str | float): The wavelength of the wavefront. Can be 'primary'
            or a float value.
        num_rings (int, optional): The number of rings for ray tracing.
            Defaults to 15.
    ...
    """
```

But the `num_rings` argument is actually called `num_rays` according to its `__init__`:

```python
    def __init__(
        self,
        optic: Optic,
        field: tuple[float, float],
        wavelength: str | float,
        num_rays: int = 15,
        distribution: DistributionType = "hexapolar",
        strategy: WavefrontStrategyType = "chief_ray",
        remove_tilt: bool = False,
        **kwargs,
    ) -> None:
```

The problem is further complicated by the section of the doc string listing [the class attributes](https://github.com/optiland/optiland/blob/b9f61dabce896cec885752e7fbbcce21d304d6f1/optiland/wavefront/opd.py#L62), which indicates it could represent either the number of rings or the number of rays.

```python
    Attributes:
        ...
        num_rays (int): The number of rays (or rings for hexapolar distribution) to use
            for pupil sampling.
        ...
```

**To Reproduce**

Check the class definition itself at https://github.com/optiland/optiland/blob/b9f61dabce896cec885752e7fbbcce21d304d6f1/optiland/wavefront/opd.py#L39.

**Expected behavior**

I expect the name of the argument in the doc string to match the actual argument name in the class's `__init__()`.

It should probably be clearer as well in that this argument's interpretation depends on the type of ray distribution used.

**Environment**  
[...]

PR #785, opened 2026-09-07
Fix OPD class docstring and clarify meaning of num_rays

kmdouglass:
This fix addresses #761 by changing the documented argument to `OPD.__init__()` from `num_rings` to `num_rays` and adds the missing `distribution` argument to the docstring.

As discussed in the issue, I have also added a note explaining that the interpretation of `num_rays` depends on the value of `distribution` in several user-facing entry points; I avoided being too precise to help avoid future drift between the docstring and interpretation of the `num_rays` value.

Finally, I mention in the `OPD` class docstring that the `afocal` argument is passed to the `Wavefront` parent `__init__()` via `kwargs`. It probably makes the most sense to make it an explicit argument to the `OPD` class, but I decided it was safer to keep this PR scoped to docstrings only.

### DIFF
--- a/optiland/optic/optic.py
+++ b/optiland/optic/optic.py
@@ -774,8 +774,9 @@ class Optic:
             Hx: The normalized x field coordinate(s).
             Hy: The normalized y field coordinate(s).
             wavelength (float): The wavelength of the rays in microns.
-            num_rays: The number of rays to trace.
-                Defaults to 100.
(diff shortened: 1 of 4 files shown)


HarrisonKramer:
Hi @kmdouglass,

Looks good. Thanks for adding this. Merging now.

Kramer

PR #786, opened 2026-09-08
Expand OSLO imports with strict diagnostics and verified command mappings

BrunoChevalier:
> **Superseded by [PR #800: Validate OSLO sequential import and loss-aware export](https://github.com/optiland/optiland/pull/800).**
>
> Closing this combined draft in favor of focused PRs following the maintainer discussion below. #800 delivers the standalone I/O portion first. Native materials, physical apertures, polarization, glass catalogs and GUI support will follow separately, with each dependent import/export mapping kept alongside its required native functionality.
>
> The original draft description below is retained as historical context; current scope and validation are documented in #800.

---

OSLO imports previously discarded unrecognized optical commands, ignored lens units, misaligned asphere powers and reduced every PY solve to final-image autofocus. This expands sequential `.len` import with documented command mappings, source-located diagnostics and an opt-in `strict=True` mode that rejects unsupported commands and known approximations.

The implementation covers units and indexed spectra; sampled glass indices; conic, general asphere and toroidal geometry; checked and special apertures; OSLO coordinate conventions and local mirror bends; static pickups; targeted paraxial/edge-contact solves; explicit field tables; image-space setup; and ruled gratings. Small reusable additions are `TabulatedMaterial`, `RotatedAperture` and `UnclippedAperture`. Coupled writer fixes preserve supported spectra, fields and aperture flags and reject unsupported exports before opening the output file.

The [support matrix](https://github.com/BrunoChevalier/optiland/blob/codex/oslo-import-support/docs/oslo_import.rst) links primary Lambda Research manuals and describes deferred features. This remains a sequential snapshot importer: exact OSLO perfect imagery, general GRIN/non-sequential execution, user surfaces, advanced field aiming and multi-configuration execution are outside this change.

Validation:

- Test-driven increments, followed by two separate review-and-fix rounds. The reviews added failing regressions for constraint propagation, malformed input, serialization and export loss before fixing them.
- **757 passed, 1 skipped**, covering all file-I/O tests, physical apertures, materials, native solves and golden snapshots. OSLO coverage increases from 26 to 187 cases, including independent sag, ray, aperture and coordinate assertions on NumPy and Torch.
- Full `ruff check optiland` and `ruff format --check optiland` pass (523 source files); the audit script passes Ruff. Sphinx is unavailable locally, so the documentation site was not rebuilt.
- The offline `scripts/audit_oslo_examples.py` records hashes and per-file outcomes for the separately downloaded official 101-lens demo ZIP. Both backends import all 101 permissively (5 clean, 96 warned); 10 pass strict import. On-axis smoke checks transmit finite rays in 80 files, none in 20, and fail in 1 because of an existing unsupported finite object-height/off-axis-entry combination. These checks do not certify agreement with OSLO ray intercepts. The proprietary archive is not vendored.

The work is split into 14 focused commits with detailed mapping rationale, limitations and validation in their messages.

### DIFF
--- a/docs/api/api_fileio.rst
+++ b/docs/api/api_fileio.rst
@@ -7,6 +7,11 @@ The ``optiland.fileio`` package handles saving and loading of optical systems.
 It supports the native Optiland JSON format, as well as Zemax (.zmx), CODE V (.seq), 
 and OSLO (.len) files.
 
+.. toctree::
+   :maxdepth: 1
+
+   /oslo_import
+
 .. autosummary::
    :toctree: generated
    :nosignatures:
--- a/docs/oslo_import.rst
+++ b/docs/oslo_import.rst
@@ -115,6 +115,9 @@ UTF-8 (including BOM) and legacy Windows-1252 text are accepted.
        unsatisfied solves restore saved values with a warning in permissive mode;
        strict mode rejects. General simultaneous constraint solving is not provided:
        a coupled case can be rejected even if a joint solution exists in OSLO.
+       EC verifies physical contact at the first surface's local meridional edge
+       after positioning both surfaces. Relative transforms that prevent this
+       contact restore the saved prescription rather than accepting TH alone.
        Telecentric PYC/PUC solves are not mapped because the native paraxial chief
        ray used by those solves does not implement the telecentric launch.
    * - ``GSP``, ``GOR``
--- a/docs/oslo_import.rst
+++ b/docs/oslo_import.rst
@@ -153,6 +153,8 @@ custom interaction, material or propagation models, ideal-material absorption, c
 scattering raise before the destination file is opened. Names and notes must fit
 on a single line. Use native JSON for those systems. Finite-object and floating-stop
 apertures export the actual axial beam radius at surface 1.
+The standard and even-asphere handlers also validate the concrete geometry;
+a surface label cannot authorize dropping or changing its actual sag terms.
 The writer rejects finite object distances that OSLO would interpret as infinite
 and preserves thickness precision to avoid rounding across that boundary.
 It writes the physical final gap with zero image defocus, so exporting an imported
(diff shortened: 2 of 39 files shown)


BrunoChevalier:
@HarrisonKramer before I put much more work into adding and extending the OSLO import support, how likely is this to be merged when I feel it's in a good enough state?

HarrisonKramer:
@BrunoChevalier,

As long as the PR remains primarily related to pure improvements in OSLO IO support & it's validated, then I can essentially guarantee we will merge. I see some minor changes to other areas of the codebase so far (e.g., `materials`). What you have now is so far fine, but I'd strongly prefer any changes not _directly_ related to OSLO import/export be in a separate PR. Last remark - I generally prefer smaller PRs, but larger PRs are perfectly fine if they're focused on one task (I think that's where this one fits). You might consider splitting up into mulitple PRs if one gets too large, but up to you.

What I see so far looks good! Thanks.

Kramer

BrunoChevalier:
Great, it's indeed getting on the larger side already.
I will split it up into several PRs, with a bigger PR that adds the standalone import/export code.

The features that need changes in other parts of the code will be handled as separate PRs.

I'm also adding code for GUI import, so that will be a separate PR as well.


One thing I'm not certain on is how to handle glass materials that come from OSLO's glc files.
A .len file does not contain any material definition other than a name.
Optiland has its own material catalog and other importers try to select the closest material that can be found in the library.
I'm leaning to converting and embedding the material definition into the imported file to keep the import as close as possible to the original. So a user would get the option to provide directory in which we will look for glc files, we'll use those to convert to embedded materials if available. If no glc files are found or selected, we will try to look up similar materials in the native catalog.

Will probably discuss this in a separate PR when I have a proposition unless you already say that converting from a catalog into an embedded material is a big no-no.

PR #688, opened 2026-07-18
Minor improvements to the file IO system

8bit-Dude:
This are minor changes, just to get warmed up:

- Added support for paraxial (thin lens) import/export with CodeV, OSLO and Zemax.
- Added fix for zemax files missing the FTYP information blob (for example, files exported from Optalix).
- Fixed formatting of exported OSLO files (indentation matching exports from OSLO).

Note 1: The contents of "encoder.py" and "formatter.py" seem to be inverted between OSLO and CODEV writers. This is confusing and should be fixed at some point...

Note 2: The IO for Zemax uses a different approach that the other 2 systems. Any reason for this?

To test these changes, you can run the following round-trip read/writes with the attached file (the output has been checked to load in relevant target softwares):

[paraxial_system.json](https://github.com/user-attachments/files/30149072/paraxial_system.json)

``from optiland.fileio import (load_optiland_file, load_zemax_file, load_codev_file, load_oslo_file, 
                             save_optiland_file, save_zemax_file, save_codev_file, save_oslo_file)

filename = "paraxial_system"

optic = load_optiland_file("JSON/"+filename+".json")
optic.draw()

save_zemax_file(optic, "ZEMAX/"+filename+".zmx")
save_codev_file(optic, "CODEV/"+filename+".seq")
save_oslo_file(optic, "OSLO/"+filename+".len")

optic = load_zemax_file("ZEMAX/"+filename+".zmx")
optic.draw()

optic = load_codev_file("CODEV/"+filename+".seq")
optic.draw()

optic = load_oslo_file("OSLO/"+filename+".len")
optic.draw()``

PR #791, opened 2026-09-09
fix: honor declared active slots in Zemax imports

### LINKED ISSUE #790: Zemax reader ignores FTYP active counts and limits WAVM records by row length
# Zemax reader ignores FTYP active counts and limits WAVM records by row length

**Checklist**
- [x] I searched existing issues and discussions for related reports.
- [x] I checked the documentation and parser implementation.
- [x] I reproduced this against current upstream `master`, commit `7df37590c2d561d95403dc264e27b8b9a77ee879`.
- [x] I reproduced and debugged the issue before reporting it.
- [x] The report includes a complete synthetic reproducer and environment details.

**Describe the Bug**
The Zemax text reader parses the active field and wavelength counts in `FTYP`,
but does not apply them when ingesting coordinates and indexed `WAVM` records.

This can silently turn a monochromatic model into a polychromatic one by
activating stored, inactive wavelength slots. Conversely, a model with five
active wavelengths imports only four when weights are supplied, or three when
weights are omitted. `PWAV` can then reference a wavelength that was discarded.

The active count is not the number of stored wavelength records. An inactive
slot can have a nonzero stored weight; an active slot can have weight zero.
Field-coordinate padding is affected by the same count-handling error.

The reproducer below uses a complete synthetic, three-plane ZMX file.

**To Reproduce**
Run this self-contained script with Optiland imported from the affected
upstream source checkout. It creates its fixture in a temporary directory.
Use an isolated environment with that checkout installed in editable mode,
and verify that the printed import path points to the intended checkout.

```python
from pathlib import Path
from tempfile import TemporaryDirectory

import optiland
from optiland.fileio.zemax.reader.parser import ZemaxDataParser

zmx = """VERS 240823
MODE SEQ
NAME SyntheticFTYP
ENPD 4
FTYP 1 0 1 1 0 0 0 1
XFLN 0
YFLN 0
FWGN 1
WAVM 1 .7800000 1
WAVM 2 .4350000 .1
WAVM 3 .6438469 .5
WAVM 4 .5460740 1
WAVM 5 .6500000 .7
PWAV 1
SURF 0
  TYPE STANDARD
  CURV 0
  DISZ 10
SURF 1
  TYPE STANDARD
  STOP
  CURV 0
  DISZ 20
  DIAM 2
SURF 2
  TYPE STANDARD
  CURV 0
  DISZ 0
"""

print("Imported Optiland:", optiland.__file__)
with TemporaryDirectory() as directory:
    path = Path(directory) / "inactive_wavelengths.zmx"
    path.write_text(zmx, encoding="utf-8")
    model = ZemaxDataParser(str(path)).parse()

print(model.wavelengths)
assert model.wavelengths["num_wavelengths"] == 1
assert model.wavelengths["data"] == [0.78]
assert model.wavelengths["weights"] == [1.0]
assert model.wavelengths["primary_index"] == 0
```

On the tested upstream commit, the first count assertion passes but the
wavelength-data assertion fails. The relevant output is:

```python
{
    "num_wavelengths": 1,
    "data": [0.78, 0.435, 0.6438469, 0.546074],
    "weights": [1.0, 0.1, 0.5, 1.0],
    "primary_index": 0,
}
```

**Expected Behavior**
For this fixture, only source wavelength slot 1 is active:

```python
{
    "num_wavelengths": 1,
    "data": [0.78],
    "weights": [1.0],
    "primary_index": 0,
}
```

Related reproductions using the same five stored `WAVM` records:

For active count `N`, replace the header with `FTYP 1 0 1 N 0 0 0 1`.
To omit weights, remove the final weight token from every `WAVM` row.

| Declared Active Count | Weight Tokens | Expected Imported Count | Actual Imported Count |
| --- | --- | --- | --- |
| 1 | Present | 1 | 4 |
| 1 | Omitted | 1 | 3 |
| 4 | Omitted | 4 | 3 |
| 5 | Present | 5 | 4 |
| 5 | Omitted | 5 | 3 |

With count 5 and `PWAV 5`, the model records primary index 4 even though fewer
than five wavelengths were imported. The converted wavelength group can mask
this by retaining its automatically selected first primary wavelength.

**Environment**
- Optiland source: upstream `master`, `7df37590c2d561d95403dc264e27b8b9a77ee879`.
- Python: 3.14.2.
- OS: Windows.
- NumPy: 2.5.3; SciPy: 1.18.1; pytest: 9.1.1.
- Additional converter checks: Torch 2.13.0 on CPU.

The imported module path and Git SHA identify the tested source. This report
targets the upstream development code, not a stock Optiland 0.6.2 installation.

**Additional Context**
The regression was introduced in
[#688](https://github.com/optiland/optiland/pull/688), commit
[`ab6c3fcd1823930f787868459f042de894c61bf3`](https://github.com/optiland/optiland/commit/ab6c3fcd1823930f787868459f042de894c61bf3).
That commit replaced working declared-count lookups with a fallback gated on
`self._ftyp`, initialized to `None` but never assigned by `_read_config_data()`.

In the
[affected parser](https://github.com/optiland/optiland/blob/7df37590c2d561d95403dc264e27b8b9a77ee879/optiland/fileio/zemax/reader/parser.py#L180-L197),
the wavelength handler therefore always uses `len(data)` as the total list
limit. A weighted `WAVM index wavelength weight` row has four tokens; removing
the weight leaves three. Neither value is an active wavelength count.

The same unset flag controls both field-coordinate handlers. The additional
`if data == [0]` guard is inert for tokenized `list[str]` input.

Simply setting `_ftyp = data` is not a sufficient repair. Historical ZMX
files legitimately use the short `FTYP i` form, which contains only the field
type and does not declare field or wavelength counts. The Version 9.0
specification cited below defines this form on printed p. 446, and its
"Minimum ZMX file" on p. 449 uses `FTYP 0` together with `WAVL 0.55`,
`WWGT 1`, and `PWAV 1`. An absent count therefore cannot be interpreted as
an explicit zero count. The fix must distinguish unspecified counts from
explicitly serialized counts, independently for fields and wavelengths.

Legacy short-header handling from
[#585](https://github.com/optiland/optiland/pull/585) should remain supported.
Optiland 0.6.2 predates #688; this is an upgrade blocker for integrations moving
to affected development revisions.

The operand handlers should honor the counts whether invoked through `parse()`
or directly through `parser._operand_table[operand](tokens)`.

**Format Documentation**
The *ZEMAX Optical Design Program User's Guide, Version 9.0* (Focus Software,
Incorporated, March 2000) is vendor-authored documentation available on an
unofficial mirror: [full PDF download](https://pdfcoffee.com/qdownload/zemaxmanual-pdf-free.html).
Chapter 29, "THE ZMX FILE FORMAT", printed pp. 445-449 (PDF pages 457-461),
explicitly addresses reading ZMX files for analysis outside ZEMAX and converting
other lens data to ZMX, and provides a table defining ZMX keywords and syntax.
These references are to the full downloaded PDF, not the truncated HTML preview.

Besides the short `FTYP i` form, that table separately defines `PWAV n` as the
primary wavelength indicator (p. 447) and `WAVL val1 val2 val3...` /
`WWGT val1 val2 val3...` as wavelength values in microns and their weights
(p. 448). It describes `XFLD` as a list of one to twelve field values on the
same page. The explanation of the minimum-file example (p. 449) identifies
the header through `PWAV` as system state, including fields and wavelengths.
This is a legacy specification: it does not document indexed `WAVM` records
or the additional count operands in later `FTYP` records.

Later manufacturer-authored documentation distinguishes counts from individual
entries. The *ZEMAX Optical Design Program User's Manual*, July 8, 2011,
Radiant ZEMAX LLC, describes wavelength activation and deactivation on
[printed p. 116](https://neurophysics.ucsd.edu/Manuals/Zemax/ZemaxManual.pdf#page=116).
Its [system-property table, p. 731](https://neurophysics.ucsd.edu/Manuals/Zemax/ZemaxManual.pdf#page=731)
separately defines field count (`101`), primary wavelength number (`200`),
wavelength count (`201`), and indexed wavelength value/weight (`202`/`203`).
[GetWave, p. 765](https://neurophysics.ucsd.edu/Manuals/Zemax/ZemaxManual.pdf#page=765)
returns the primary number and currently defined wavelength count for index
zero, or a wavelength/weight pair for an individual index. These document
system state, not the serialized `FTYP` operand layout.

A [2020 Zemax Community response](https://community.zemax.com/zpl-13/zmx-file-specification-106)
states that the file-format documentation left the Help System around 2005.

The later serialized layout is illustrated by Optical Systems Design's
third-party Lesson 33 import example. Its
[first page](https://osdoptics.com/wp-content/uploads/2020/05/33_importing_a_zemax_file.pdf#page=1)
prints `FTYP 1 0 3 3 0 0 0`, three-entry `XFLD`/`YFLD` lists, and twelve-entry
`XFLN`/`YFLN`/`FWGN` lists. The
[second page](https://osdoptics.com/wp-content/uploads/2020/05/33_importing_a_zemax_file.pdf#page=2)
prints `WAVM 1` through `WAVM 24`, all with weight 1, including slots 4-24,
alongside three-entry `WAVL` and `WWGT` lists and `PWAV 2`. This corroborates
the distinction between configured counts and stored records, including extra
records with nonzero weights; it is not a manufacturer-authored specification.

The legacy short-header form is vendor-documented. Optiland's exact inference
policy for unspecified counts and its handling of malformed declarations,
sparse or reordered slots, and repeated records are compatibility and robustness
decisions, not requirements specified by these historical sources.

**Proposed Contribution**
I have a focused fix and synthetic regression tests prepared on top of the
tested upstream commit and would like to submit a PR. The isolated change has
191 passing regression cases; the complete scoped file-I/O and weight test run
has 464 passes, including NumPy and Torch converter checks.

The proposed scope is active-count handling, indexed `WAVM` ordering and
primary identity, and the necessary existing model-consumer compatibility.
It does not add inactive-slot export fidelity, full legacy `WAVL` vector/weight
parsing, or reinterpret version-dependent trailing `FTYP` values.

Please let me know if someone is already working on this or if there is an
existing issue I should reference. I am happy to take this one.

mabl:
Fixes https://github.com/optiland/optiland/issues/790

**Problem**
The reader parses active field and wavelength counts from `FTYP`, but ingestion
depends on `self._ftyp`, which is initialized to `None` and never assigned.
The fallback instead uses the token count of a single `WAVM` row as the total
wavelength limit: four with a weight token, or three without one.

This silently activates stored inactive wavelengths in a monochromatic model,
discards active wavelengths from larger sets, and can leave `PWAV` referring to
a discarded slot. Field-coordinate padding is affected by the same unset flag.
This regression was introduced by #688; Optiland 0.6.2 predates it.

**Changes**
- Honor independently declared field and wavelength counts, including explicit
  zero, without mistaking absent-count defaults for serialized declarations.
- Retain full field columns privately and publish consistent active prefixes
  across coordinates, weights, and vignetting, including declaration changes.
- Store indexed `WAVM` value/weight pairs by source slot and publish in slot order.
  Repeated slots update both values; equal-valued distinct slots remain distinct.
  Neither weight nor row length determines activation.
- Resolve `PWAV` using source-slot identity so sparse or reordered records cannot
  silently select another wavelength. Apply this in operand handlers for both
  public parsing and direct dispatch.
- Validate explicitly invalid primary indices in model consumers while preserving
  existing missing-key behavior and NumPy/Torch integer-index compatibility.
  Infer an absent encoder wavelength count from actual data and omit `PWAV` for
  empty active data, avoiding stale defaults that would truncate corrected output.

**Legacy Compatibility**
The short-header form is vendor-documented syntax, not merely malformed input.
The *ZEMAX Optical Design Program User's Guide, Version 9.0* (Focus Software,
March 2000, Chapter 29, printed pp. 446 and 449; PDF pages 458 and 461)
defines `FTYP i` as field type only. Its minimum-file example uses `FTYP 0`
with `WAVL 0.55`, `WWGT 1`, and `PWAV 1`. An absent count therefore cannot
generally mean an explicit zero count. Preserve the short-header handling
supported since #585.

The [full manual download](https://pdfcoffee.com/qdownload/zemaxmanual-pdf-free.html)
is an unofficial mirror of vendor-authored documentation. It establishes the
legacy form, not modern `FTYP` count operands or indexed `WAVM` syntax. The
later system-state documentation and published three-active/24-stored example
are linked and distinguished in #790.

The exact inference policy for unspecified counts and handling of malformed
declarations, sparse or reordered slots, and repeated records remain Optiland
compatibility and robustness decisions. Explicit positive counts are retained
for tolerated underfilled inputs; no count/data equality requirement is added.

**Scope**
This is one standalone commit on upstream `master`
`7df37590c2d561d95403dc264e27b8b9a77ee879`, changing three source files, one
regression module, and one complete synthetic three-plane fixture.

Full legacy `WAVL` vector/separate-weight parsing, inactive-slot export fidelity,
and version-dependent trailing `FTYP` interpretation are not added. Existing
field sorting/deduplication and writer unit-weight emission remain unchanged.
No physics algorithms, dependencies, or file-model schema are changed.

**Verification**
- `python -m pytest -q tests/test_fileio tests/test_weights.py`: 464 passed,
  including the 191 FTYP regression cases and NumPy/Torch CPU consumer checks.
- Regression coverage includes declared counts 1/4/5, omitted weights, inactive
  nonzero weights, active zero weights, equal-valued slots, primary identity,
  field padding and metadata, short/absent/malformed headers, count transitions,
  explicit zero, underfilled inputs, direct dispatch, and encoding defaults.
- The synthetic fixture also checks imported geometry, stop, aperture, transfer
  matrix, and primary-ray results against an independently constructed `Optic`.
- Against the original upstream parser/consumers loaded in memory, the expanded
  regression suite produced 163 failures and 28 passes. The main reproducer
  imports four wavelengths before the fix and only `[0.78]` afterward.
- `python -m ruff check optiland/`, formatting of the three changed source files,
  and `git diff --check` pass.

Environment: Python 3.14.2, NumPy 2.5.3, SciPy 1.18.1, Torch 2.13.0 CPU,
pytest 9.1.1, Ruff 0.16.6. The full global suite was not run, existing warnings
remain visible, and no mathematical tolerances were relaxed.


HarrisonKramer:
Hi @mabl,

Thanks for the fix. Looks good. Merging.

Kramer

PR #793, opened 2026-09-09
fix: make cached ray launches field- and autograd-safe

### LINKED ISSUE #792: Cached ray aiming reuses stale finite-conjugate object point after field-height edit
# Cached ray aiming reuses stale finite-conjugate object point after field-height edit

## Checklist
- [x] Searched related issues and discussions; this is distinct from closed #482 (TMA non-convergence) and #654 (stop-radius handling).
- [x] Read the ray-aiming developer/API documentation and the cache feature description in #428.
- [x] Reproduced on upstream `master` at commit `7df37590c2d561d95403dc264e27b8b9a77ee879`.
- [x] Included a complete synthetic example, numerical results, and uncached/fresh controls.

## Bug
Enable `cache=True`, trace an off-axis field, change the maximum object height, and trace the same normalized field again. Both `iterative` and `robust` report convergence but keep the previous object point and launch direction. The ray still hits the stop center, so checking stop convergence alone misses the wrong field and image coordinate.

This affects ordinary `Optic.trace`, not just direct reuse of `CachedRayAimer`. The aimer's identity is unchanged across the two traces. [`RayGenerator.generate_rays`](https://github.com/optiland/optiland/blob/7df37590c2d561d95403dc264e27b8b9a77ee879/optiland/rays/ray_generator.py) only replaces it when `ray_aiming_config` changes; editing `optic.fields[1].y` does not change that configuration. Keeping the aiming configuration while editing a field extent is a normal design-study workflow. [#428](https://github.com/optiland/optiland/pull/428) introduced this option specifically for repeated calculations across system changes, and the [developer guide](https://www.optiland.org/docs/developers_guide/ray_overview.html) documents explicit `cache=True` as supported.

## Reproducer
Run this standalone script with Optiland imported from the affected upstream checkout. All dimensions are mm; wavelength is micrometers. It builds an ideal-glass singlet and a centered physical stop, with no catalog files or test fixtures. The current `optic.ray_tracer.set_aiming` API is used below; the compatibility API `optic.set_ray_aiming` reproduces the same failure.

```python
import numpy as np

import optiland.backend as be
from optiland.distribution import LineYDistribution
from optiland.materials.ideal import IdealMaterial
from optiland.physical_apertures.radial import RadialAperture


def make_optic(mode, cache):
    optic = Optic()
    optic.surfaces.add(index=0, thickness=100.0)
    optic.surfaces.add(
        index=1, radius=50.0, thickness=5.0, material=IdealMaterial(1.5)
    )
    optic.surfaces.add(index=2, radius=-50.0, thickness=15.0)
    optic.surfaces.add(index=3, thickness=40.0, is_stop=True)
    optic.surfaces.add(index=4)
    optic.surfaces[3].aperture = RadialAperture(r_max=4.0)
    optic.set_aperture("float_by_stop_size", 8.0)
    optic.fields.set_type("object_height")
    optic.fields.add(y=0.0)
    optic.fields.add(y=1.0)
    optic.wavelengths.add(0.55, is_primary=True)
    optic.ray_tracer.set_aiming(mode=mode, cache=cache, max_iter=30, tol=1e-10)
    return optic


def trace(optic):
    # One chief ray: normalized field (0, 1), pupil (0, 0).
    distribution = LineYDistribution(positive_only=True)
    distribution.generate_points(1)
    rays = optic.trace(0.0, 1.0, 0.55, distribution=distribution)
    obj, stop = optic.surfaces[0], optic.surfaces[3]
    values = np.asarray(
        [obj.y[0], obj.M[0], obj.N[0], stop.y[0], rays.y[0]], dtype=float
    )
    assert np.all(np.isfinite(values)) and np.all(rays.i > 0)
    assert np.allclose(stop.x, 0.0, atol=1e-9, rtol=0.0)
    return values


be.set_backend("numpy")
be.set_precision("float64")
np.set_printoptions(precision=12, suppress=True)
for mode in ("iterative", "robust"):
    optic = make_optic(mode, True)
    before = trace(optic)
    aimer = optic.ray_tracer.ray_generator.aimer
    optic.fields[1].y = 3.0
    warm = trace(optic)
    same = aimer is optic.ray_tracer.ray_generator.aimer
    converged = aimer.wrapped_aimer.last_report.converged
    optic.ray_tracer.set_aiming(mode=mode, cache=False, max_iter=30, tol=1e-10)
    uncached = trace(optic)
    optic.ray_tracer.set_aiming(mode=mode, cache=True, max_iter=30, tol=1e-10)
    fresh = trace(optic)
    print(mode, "same aimer:", same, "warm converged:", converged)
    print("columns: object_y, launch_M, launch_N, stop_y, image_y")
    for label, values in (("before", before), ("warm", warm),
                          ("uncached", uncached), ("fresh", fresh)):
        print(label, values)
        print("  stop_y:", format(values[3], ".17g"))
    print("image error:", format(warm[4] - uncached[4], ".17g"))
    print("warm equals uncached:",
          np.allclose(warm, uncached, atol=1e-9, rtol=0.0))
    assert same and converged
    assert uncached[0] == 3.0
    assert np.allclose(uncached, fresh, atol=1e-9, rtol=0.0)
    assert abs(warm[3]) < 1e-10 and abs(uncached[3]) < 1e-10
```

## Expected And Actual
Expected: `warm equals uncached: True`, object height 3 mm, and agreement with a fresh cached aimer. Actual upstream: `False` for both modes, despite `same aimer: True` and `warm converged: True`.

Results for both upstream aiming modes:

| Case | Object y | Launch M | Launch N | Stop y | Image y |
|---|---:|---:|---:|---:|---:|
| Before edit | 1 | -0.007901567971 | 0.999968782125 | -3.9440672949808686e-14 | -0.470621960805 |
| Warm after edit | 1 | -0.007901567971 | 0.999968782125 | -3.9440672949808686e-14 | -0.470621960805 |
| Uncached / fresh after edit | 3 | -0.023695274792 | 0.999719227560 | -8.8787865948347644e-11 | -1.412676824829 |

The warm image-height error is **0.94205486402373451 mm**. The rays remain finite and transmitted. Stop x is zero within 1e-9 mm, and stop y is within the requested 1e-10 mm solver tolerance.

## Root Cause
The system hash already includes the fields. This is not simply a missing field entry in the key. On a system-hash mismatch with identical normalized inputs, upstream [`CachedRayAimer.aim_rays`](https://github.com/optiland/optiland/blob/7df37590c2d561d95403dc264e27b8b9a77ee879/optiland/rays/ray_aiming/cached.py) forwards the entire old `(x, y, z, L, M, N)` as `initial_guess`. The [iterative solver](https://github.com/optiland/optiland/blob/7df37590c2d561d95403dc264e27b8b9a77ee879/optiland/rays/ray_aiming/iterative.py) retains the supplied object point for finite conjugates and adjusts direction to satisfy the stop target. The old ray already satisfies that target, so it converges without moving to the new object point. [Robust aiming](https://github.com/optiland/optiland/blob/7df37590c2d561d95403dc264e27b8b9a77ee879/optiland/rays/ray_aiming/robust.py) accepts this successful initial-guess solve too; its fallback cannot catch a wrong field that still satisfies the stop constraint.

A cached full launch state is not a valid warm start after this field edit unless its fixed field-dependent components are rebuilt. Delegating without the old full-state guess is one fix that restores agreement here. With that change, the iterative warm/uncached image difference is 0 mm; robust is 4.8893111781467269e-11 mm. The fixed robust warm stop y is 3.9492853431966068e-12 mm. Both modes then print `warm equals uncached: True` at absolute tolerance 1e-9 and zero relative tolerance.

## Scope And Environment
Reproduced on Windows with Python 3.14.2, NumPy 2.5.3, float64; installed Torch is 2.13.0+cpu, but this example uses NumPy.

This report is limited to explicit `cache=True` and a finite object-height edit. The robust aimer's independent `PupilMapCache` is outside its scope.

Related context: [cache feature #428](https://github.com/optiland/optiland/pull/428), [closed TMA issue #482](https://github.com/optiland/optiland/issues/482), [closed aiming issue #654](https://github.com/optiland/optiland/issues/654), and [aiming contribution discussion #677](https://github.com/optiland/optiland/discussions/677). No matching field-edit cache report was found in the issue searches or the discussion search for `cache aiming`.

## Proposed Contribution
I have a fix and regression tests prepared that avoid passing the stale full
launch state after a system-state change while preserving exact cache hits.
I am happy to submit a PR.

Focused regression tests pass with the patch. Broader ray-aiming checks expose
unrelated NumPy 2.5.3 array-to-scalar conversion failures that also reproduce
with the upstream cache implementation. Detailed validation results can
accompany the PR.

mabl:
Fixes https://github.com/optiland/optiland/issues/792

**Problem**
Ordinary `Optic.trace` with explicit `cache=True` can reuse a stale
finite-conjugate object point after a field-height edit. The system hash already
includes fields. The defect is using the previous full launch state as an
automatic `initial_guess` after the system hash changes.

The iterative solver keeps the supplied object point fixed and solves direction
to hit the stop. Robust aiming first tries that same initial-guess solve. The old
ray can therefore report convergence while representing the wrong field.

The synthetic singlet in #792 changes the object field from 1 mm to 3 mm while
retaining normalized inputs `(0, 1)`. Both upstream modes keep the 1 mm origin,
causing a **0.942054864024 mm image-height error**, even though the stale and
correct rays both satisfy the requested `1e-10` mm stop tolerance.

**Changes**
- Delegate without an automatic stale full-state guess when optical state changes.
  Preserve exact numerical hits for unchanged requests and state.
- Forward explicit caller guesses without caching. When none is supplied, do not
  inject an optional guess keyword into three-argument wrapped aimers.
- Clear and bypass the wrapper under either Optiland gradient tracking or ambient
  Torch autograd, preventing reuse of consumed graphs or old parameter leaves.
- Invalidate numerical entries when backend, precision, device, or Torch inference
  mode changes.
- Hash the raw serialized system snapshot rather than rounded display text, so
  nearby array-valued changes cannot falsely qualify as exact hits.

The additional guards enforce the same full-result reuse policy in this wrapper.
They are covered by separate tests; the issue's optical example demonstrates the
field-edit failure, not every cache-safety concern.

**Compatibility**
FIFO capacity and unchanged numerical cache hits remain supported. Automatic
full-launch warm starts after system changes are intentionally removed.
Torch numerical caching requires both gradient controls to be disabled; default
ambient autograd therefore intentionally reduces cache reuse.

Torch stays optional. Snapshot handling serializes data only and does not load
untrusted pickle payloads. Raw Torch serialization can include storage identity
and backing allocations, so equal-valued replacements may conservatively miss
and small views of large storage may be more expensive to fingerprint.

**Scope**
This standalone four-file patch changes only `CachedRayAimer` and three test
modules, based on upstream `master` at
`7df37590c2d561d95403dc264e27b8b9a77ee879`. It does not depend on #780 or #791.

The underlying solvers and robust aiming's independent `PupilMapCache` are not
modified. That map cache's rounded field-fingerprint limitation remains outside
this PR; this patch does not claim to fix all robust-cache precision issues.
It also does not establish full differentiation through detached field
calibration or already-converged robust seeds.

**Validation**
- `python -m pytest -q tests/test_cached_aimer.py tests/test_cached_field_state.py tests/test_cached_snapshot_precision.py`: **78 passed, 14 skipped**.
- Broader existing ray-aiming selection: **302 passed, 1 skipped, 11 baseline failures**.
  The 11 failures at `tests/test_ray_aiming.py:374`, `:465`, and
  `tests/test_ray_aiming_numerics.py:785` are array-to-scalar conversions on
  NumPy 2.5.3. All also reproduce with the upstream cache methods restored in
  memory. They are not fixed or hidden by this PR.
- Against the original cache methods, the focused selection gives **70 expected
  failures, 8 passes, 14 skips**. This includes delegation-contract checks as
  well as numerical regressions.
- The exact issue reproducer now agrees with uncached and fresh controls:
  iterative image difference **0 mm**, robust **4.8893111781467269e-11 mm**.
  Both modes satisfy the strengthened `1e-10` mm stop assertion.
- `python -m ruff check .`, `python -m ruff format --check .`, and whitespace
  checks pass. No mathematical tolerances were relaxed.

Tests cover field replacement/scaling, finite origins and infinite directions,
exact hits, explicit guesses, graph lifetime, equal-valued leaf replacement,
context transitions, and snapshot mutations hidden by display rounding.
Graph-lifetime controls use a nonlinear test aimer rather than a full physical
solver derivative check; device transitions are mocked.

Environment: Windows, Python 3.14.2, NumPy 2.5.3, SciPy 1.18.1,
Torch 2.13.0 CPU, pytest 9.1.1, Ruff 0.16.6. The global suite and real accelerator
execution were not tested.

### DIFF
--- a/optiland/rays/ray_aiming/cached.py
+++ b/optiland/rays/ray_aiming/cached.py
@@ -1,8 +1,8 @@
 """Cached Ray Aiming Module
 
 This module implements a caching wrapper for ray aiming algorithms.
-It stores previous results to speed up repetitive calculations, especially
-during optimization or tolerance analysis where system changes might be small.
+It stores previous results to speed up repetitive calculations when the inputs
+and optical system remain unchanged.
 
 Retained for backward compatibility (explicit ``cache=True``) but no longer
 the default warm-start mechanism for ``"robust"``, which has its own
@@ -18,6 +18,7 @@ import hashlib
 import pickle
 from typing import TYPE_CHECKING, Any
 
+import optiland.backend as be
 from optiland.rays.ray_aiming.base import BaseRayAimer
 
 if TYPE_CHECKING:
@@ -29,16 +30,22 @@ class CachedRayAimer(BaseRayAimer):
 
     This class wraps another ray aimer and caches its results. It checks
     if the inputs and the optical system state have changed. If they match
-    a cached entry, the result is returned immediately. If the system has
-    changed but inputs match, the previous result is used as a starting guess.
+    a cached entry in the same numerical execution context, the result is
+    returned immediately. Otherwise, the wrapped aimer computes a fresh result
+    without an automatic initial guess. Explicit caller guesses bypass caching.
 
     Coordinate-frame note: cached entries are full launch states in global
     coordinates. The system hash covers every surface, so any rigid pose
-    change (translation, fold reorientation) misses the exact-reuse path
-    and the stale state is only ever passed as an ``initial_guess`` to the
-    wrapped aimer, which re-solves (and, for the robust aimer, falls back
-    to a fresh entry-frame calibration if the stale guess fails). A stale
-    global-coordinate state can therefore cost time but never correctness.
+    change (translation, fold reorientation) misses the exact-reuse path.
+    After a system change, a cached launch is not used as a starting guess:
+    its fixed field-dependent components may no longer represent the requested
+    field, even if the ray still satisfies the stop constraint.
+
+    Torch calls clear and bypass this cache whenever either backend gradient
+    tracking or ambient autograd is enabled, so computation graphs are never
+    reused. Backend, precision, device, and Torch inference-mode changes also
+    clear numerical entries.
+    The wrapped robust aimer's detached pupil-map seed cache is independent.
 
     Attributes:
         optic (Optic): The optical system being traced.
@@ -65,6 +72,7 @@ class CachedRayAimer(BaseRayAimer):
         self.wrapped_aimer = wrapped_aimer
         self.max_cache_size = max_cache_size
         self._cache: dict[str, tuple[str, tuple]] = {}
+        self._cache_context: tuple[str, int, str | None, bool] | None = None
 
     def aim_rays(
         self,
@@ -84,11 +92,44 @@ class CachedRayAimer(BaseRayAimer):
         Returns:
             tuple: Ray parameters (x, y, z, L, M, N).
         """
+        # Invalidate before any early return, including explicit caller guesses,
+        # so a bypass cannot leave incompatible entries for subsequent calls.
+        backend = be.get_backend()
+        bypass_cache = False
+        if backend == "torch":
+            import torch
+
+            # Either Optiland's gradient control or ambient Torch autograd can
+            # require a fresh graph. Equal numerical values do not identify
+            # its lifetime or the parameter leaves to which it is attached.
+            bypass_cache = be.grad_mode.requires_grad or torch.is_grad_enabled()
+
+        if bypass_cache:
+            # Discard earlier entries too, so returning to numerical execution
+            # starts a new cache rather than reviving pre-gradient results.
+            self.clear_cache()
+        else:
+            # Results must match the active backend, precision and device.
+            # Inference mode also differs from no_grad: inference tensors cannot
+            # be mutated outside inference mode or saved for backward.
+            context = (
+                backend,
+                be.get_precision(),
+                be.get_device() if backend == "torch" else None,
+                torch.is_inference_mode_enabled() if backend == "torch" else False,
+            )
+            if context != self._cache_context:
+                self.clear_cache()
+                self._cache_context = context
+
         # If an explicit initial guess is provided, we skip the cache check
         if initial_guess is not None:
             return self.wrapped_aimer.aim_rays(
-                fields, wavelengths, pupil_coords, initial_guess
+                fields, wavelengths, pupil_coords, initial_guess=initial_guess
             )
+        if bypass_cache:
+            # Do not hash, read or store results in a differentiable call.
+            return self.wrapped_aimer.aim_rays(fields, wavelengths, pupil_coords)
 
         # 1. Generate Input Hash
         input_key = self._get_input_hash(fields, wavelengths, pupil_coords)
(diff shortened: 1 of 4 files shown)


HarrisonKramer:
@mabl - another solid fix. Merging.

Kramer

PR #804, opened 2026-09-10
Fix conic intersection precision and speed up NumPy and Torch CPU tracing

BrunoChevalier:
Improves precision and tracing performance in the finite-conic solver merged
in #753. Factored coefficients and exact degeneracy checks preserve close
float32 hits and positive discriminants at microlens scales. Rounded self-hits
require both a residual within roundoff and an unresolved displacement, retaining
resolved near-tangent crossings. Per-ray sag-sheet selection, aperture preference,
and the signed vertex fallback are preserved.

### Architecture

Reworked to follow [Kramer's architectural review](https://github.com/optiland/optiland/pull/804#issuecomment-5646093846):

- Geometry calls `be.conic_intersection(...)` through the existing backend
  interface. Only numerical inputs and an optional aperture-membership callable
  cross that boundary; geometry contains no new framework imports or dispatch.
- `backend._conic` holds the shared mathematical root/selection policy. NumPy's
  compiled loops live in `backend.numpy_backend.conic`; tensor dispatch and the
  custom autograd function live in `backend.torch_backend.conic`.
- Backend helpers do not import geometry, ray or aperture classes. The former
  geometry helpers are removed, with no compatibility shims for their unreleased
  imports. The mathematical candidate calculation is unchanged by this revision.
- Torch remains optional. A fresh-process test blocks Torch imports and verifies
  NumPy float32/64 intersections plus core geometry/Optic imports. Instance tests
  also verify that execution does not depend on global backend selection.

Eligible NumPy/Torch CPU float64 arrays retain the shared Numba acceleration.
Torch uses views of existing CPU storage without coordinate copies or mutation,
with implicit first/second derivatives, JVP/Jacobians/vmap on regular branches.
CUDA, float32 and other shapes retain native operations. Exact double roots keep
their forward values and contribute zero gradient; aperture selection is discrete.

### Validation and scoped coverage

The latest test-only update is **`fb289607`**, following the merge of master
`270fe493` into this PR. It adds 10 ordinary-suite cases comparing both compiled
NumPy loops with their original Python functions. Analytic sphere roots,
candidate/regularity masks, self-hits, tangencies, misses, nonfinite inputs,
empty/single/strided arrays and parabola aperture selection provide independent
correctness checks. Production code, numerical tolerances, CI configuration and
coverage exclusions are unchanged by this update.

Local validation on Python 3.14.6 passes **676 CPU tests, 14 skips** across conic
execution/precision/root selection, Torch execution, backend ownership and
optional-Torch imports, geometry and aiming numerics. The separate RTX 5080 CUDA
selection passes **111 tests**. A JIT-disabled execution-file check passes
**21 tests, 10 skips**; the new equivalence cases require an active compiler.
Ruff, formatting and whitespace checks pass. The full Linux/Python 3.11-3.14
matrix runs in CI.

The normal **CPU-only** run uses the checked-in `.coveragerc`, ordinary JIT
execution and no JIT-coverage overrides. A separate CPU run adds branch tracing
and passes the same 676 tests. Intersecting executable coverage lines with the
full PR diff against `270fe493` gives:

| Coverage measure | Before this test update | After this test update |
| --- | ---: | ---: |
| Executable PR changes | 88.33% (159/180) | **100.0% (180/180)** |
| NumPy conic statements | 59.61% (31/52) | **100.0% (52/52)** |
| NumPy conic branches | 50.0% (4/8) | **100.0% (8/8)** |

| Production module | Line coverage | Branch coverage |
| --- | ---: | ---: |
| `optiland/backend/_conic.py` | 100.0% (52/52) | 100.0% (2/2) |
| `optiland/backend/numpy_backend/conic.py` | 100.0% (52/52) | 100.0% (8/8) |
| `optiland/backend/torch_backend/conic.py` | 100.0% (71/71) | 100.0% (14/14) |
| `optiland/geometries/standard.py` | 94.5% (52/55) | 100.0% (6/6) |

The three standard-geometry misses are unchanged `set_radius`, `flip` and
`scale` statements outside this PR's diff. All changed executable lines are
covered. Python tracing still cannot observe machine-code execution; the new
ordinary tests trace the original Python loop bodies and check their results
against compiled execution and analytic expectations. These coverage results
**do not combine a separate JIT-disabled or GPU run**. The earlier combined
report did not represent what CI uploaded; these tests address that discrepancy.

The dated performance and full-workload correctness measurements below remain
attached to their stated commits. This update changes tests only and does not
claim new benchmark measurements on the later master revision.

### Correctness of the measured workloads

Separate full-size correctness runs on all three
versions and all three execution configurations compare final ray positions,
directions, intensities and optical paths. Revised outputs agree with both
references within rtol 1e-10, with atol 1e-8 mm for positions/paths and 1e-12 for
directions/intensities. First-order values agree too. Independent geometric
sphere/parabola root checks find zero wrong selections, including all 1,440,000
sphere intersections in the spherical workload;
all benchmark rays survive with finite outputs.

| Workload | Input count | Master | Previous | Revised |
| --- | --- | --- | --- | --- |
| Spherical propagation | 180,000 rays | Checked roots; all survive | Equivalent output; all survive | Equivalent output; all survive |
| OAP propagation | 360,000 rays | Checked roots; all survive | Equivalent output; all survive | Equivalent output; all survive |
| Full generation | 2 × 60,000 rays | Checked roots; all survive | Equivalent output; all survive | Equivalent output; all survive |
| Iterative aiming | 3 × 1,024 rays | Checked roots; all survive | Equivalent output; all survive | Equivalent output; all survive |
| First-order queries | f2 / EPL / XPL | Reference values | Same values | Same values |
| #329 collimator / relay | 21 / 41 rays | All survive | All survive | All survive |

Master here already contains #753 and gives correct OAP output. Earlier tables
against pre-#753 master, which lost half the OAP rays, are historical and must not
be mixed with these numbers. Faster execution with incorrect rays would not be
an acceptable improvement. The audit validates these measured inputs; separate
precision and gradient regression tests cover additional cases.

### Performance comparison — September 12, 2026

These measurements cover the final architecture rework at **`c59de3a5`**:
geometry calls `be.conic_intersection`, the mathematical root policy is shared,
and NumPy acceleration / Torch execution and autograd live behind the backend interface.

| Version | Commit | Meaning |
| --- | --- | --- |
| Master | `66d5caa2805232fb57cdf2ca826a6922ffcaf04c` | Measured master baseline, already includes #753 |
| Previous | `07da1dd10d9a3ecfb6a4b6294b33383f67e76661` | Previous PR implementation rebased unchanged onto the same master |
| Revised | `c59de3a5ab269e6d8bf2f754728526ad420cb8e6` | Final backend architecture rework on that base |

The previous version is a same-base control that isolates the architecture change. All three tracked source trees were clean
and pinned before and after the run; each fresh process checked its import path.

Master has advanced since this measured baseline. The timings and correctness
audit below refer to the exact commits in the table, not later master commits
or GitHub's merge result.

The unchanged #756 workload builders were run in fresh interpreters: one warmup,
one timed operation, five interleaved repeats per version/configuration, and the
minimum of those five times. Versions alternate execution order between repeats.
All use the same environment, float64, one Torch CPU thread and backend gradient
mode disabled; CUDA timing explicitly synchronizes the GPU.

Hardware: AMD Ryzen 9 5950X 16-Core Processor; NVIDIA GeForce RTX 5080, 595.97, 16303 MiB. Environment: Windows-11-10.0.26200-SP0,
Python 3.14.6, NumPy 2.5.2, Numba 0.67.0, Torch 2.13.0+cu130.
Run: 2026-09-12T13:52:55.826059+00:00 through 2026-09-12T14:04:54.706598+00:00 (UTC).

Times are milliseconds. Negative changes mean less time. The proposed #756 gate
fails only if both the time ratio exceeds 1.05 and the increase exceeds 2 ms.
These are local warmed measurements; compilation/startup costs are excluded.

### NumPy CPU

| Workload | Master (ms) | Previous (ms) | Revised (ms) | vs master | vs previous |
| --- | ---: | ---: | ---: | ---: | ---: |
| Spherical propagation | 370.13 | 221.52 | 226.91 | -38.7% | +2.4% |
| OAP propagation | 198.80 | 162.41 | 161.90 | -18.6% | -0.3% |
| Full generation | 110.28 | 92.14 | 92.31 | -16.3% | +0.2% |
| Iterative aiming | 89.25 | 76.83 | 76.63 | -14.1% | -0.3% |
| First-order queries | 170.40 | 169.24 | 169.84 | -0.3% | +0.4% |

### Torch CPU

| Workload | Master (ms) | Previous (ms) | Revised (ms) | vs master | vs previous |
| --- | ---: | ---: | ---: | ---: | ---: |
| Spherical propagation | 156.03 | 88.63 | 89.68 | -42.5% | +1.2% |
| OAP propagation | 85.85 | 65.64 | 67.75 | -21.1% | +3.2% |
| Full generation | 168.88 | 112.07 | 113.24 | -32.9% | +1.0% |
| Iterative aiming | 245.99 | 214.00 | 213.06 | -13.4% | -0.4% |
| First-order queries | 571.97 | 580.70 | 588.28 | +2.9% | +1.3% |

### Torch CUDA

| Workload | Master (ms) | Previous (ms) | Revised (ms) | vs master | vs previous |
| --- | ---: | ---: | ---: | ---: | ---: |
| Spherical propagation | 56.99 | 56.37 | 59.30 | +4.1% | +5.2% |
| OAP propagation | 21.75 | 22.16 | 21.93 | +0.8% | -1.0% |
| Full generation | 386.92 | 389.40 | 392.04 | +1.3% | +0.7% |
| Iterative aiming | 1486.13 | 1500.72 | 1508.06 | +1.5% | +0.5% |
| First-order queries | 4109.00 | 4193.63 | 4148.66 | +1.0% | -1.1% |

### Performance interpretation and CUDA follow-up

Real-ray workloads take **14.1–38.7% less time on NumPy CPU** and
**13.4–42.5% less on Torch CPU** than the measured master baseline. CUDA workload
times are **0.8–4.1% higher** in the full comparison. All 15 workload/device cases
pass the proposed #756 threshold versus that baseline.

Versus the previous same-base implementation, 14/15 cases pass. CUDA spherical
propagation measures **59.30 ms versus 56.37 ms (+5.2%, +2.93 ms)** and crosses
the threshold. The preceding architecture run also observed a borderline +5.4%.
These observations remain in the report.

Master is correct for the spherical benchmark: all 180,000 rays survive and the
independent audit finds no wrong roots. Its apparent advantage varies in the
follow-up measurements below. These are separate diagnostics and do not replace
the original full-workload timings.

| Spherical propagation diagnostic | Master (ms) | Revised (ms) | Revised vs master |
| --- | ---: | ---: | ---: |
| New fresh processes, one warmup | 58.83 | 59.11 | +0.48% |
| New fresh processes, ten warmups | 58.30 | 57.91 | -0.66% |
| Same process, swap solvers only | 58.57 | 56.81 | -3.01% |

The fresh-process diagnostic uses five interleaved processes per unmodified
source checkout, recording sixteen successive traces. Each row compares the
specified trace index across five samples and reports its minimum. The paired
test warms each solver three times and runs all six version-order permutations
twice, timing five traces per sample on one common optical system/backend.
Outputs agree after each substitution. The paired previous/revised timings are
56.88/56.81 ms; a separate earlier pair measured 56.85/56.86 ms.

A separate profile counts 114 CUDA kernels and one host scalar readback for
the first 180,000-ray spherical surface in every version. Previous and revised
operator counts are identical there. CUDA retains native Torch execution;
the large CPU float64 gains use compiled loops.

The diagnostics support sensitivity to warmup and runtime conditions. They do
not isolate the contributions of scheduling, allocation state, GPU state or
other system load. The original gap's exact cause remains unproven; neither an
intrinsic CUDA regression nor a portable CUDA speedup is established. First-order
query code is unchanged by this PR and provides a timing-variability control.

### Torch precision and gradient performance

Isolated kernels use 180,000 rays, three interleaved fresh processes per version,
one warmup and five samples of five calls per case; the table reports minima.
Forward + backward rebuilds the graph and differentiates all eight numerical
inputs. Ratios are revised / previous on the same base; below 1 means less time.
The full workload tables above use float64 without gradients.

| Execution / dtype / kernel | Previous forward (ms) | Revised forward (ms) | Ratio | Previous F+B (ms) | Revised F+B (ms) | Ratio |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| NumPy CPU / float64 / conic, no aperture | 0.977 | 0.949 | 0.971 | n/a | n/a | n/a |
| NumPy CPU / float64 / OAP, aperture | 5.991 | 5.946 | 0.992 | n/a | n/a | n/a |
| Torch CPU / float32 / conic, no aperture | 8.193 | 8.224 | 1.004 | 12.054 | 12.218 | 1.014 |
| Torch CPU / float32 / OAP, aperture | 10.294 | 10.267 | 0.997 | 14.211 | 14.261 | 1.003 |
| Torch CPU / float64 / conic, no aperture | 1.055 | 1.046 | 0.992 | 6.961 | 7.038 | 1.011 |
| Torch CPU / float64 / OAP, aperture | 3.956 | 3.932 | 0.994 | 10.043 | 10.036 | 0.999 |
| Torch CUDA / float32 / conic, no aperture | 2.170 | 2.142 | 0.987 | 4.570 | 4.459 | 0.976 |
| Torch CUDA / float32 / OAP, aperture | 2.839 | 2.714 | 0.956 | 5.237 | 5.064 | 0.967 |
| Torch CUDA / float64 / conic, no aperture | 2.125 | 2.091 | 0.984 | 4.410 | 4.353 | 0.987 |
| Torch CUDA / float64 / OAP, aperture | 2.677 | 2.624 | 0.980 | 5.077 | 5.247 | 1.034 |

### Remaining limits

CPU timings are warmed; compilation/cache loading in a new interpreter remains
an additional cost; cold compilation/cache loading was not remeasured for this final revision. Native float32/CUDA results do not imply CPU float64 speedups
on every device or workload. Near tangencies are ill-conditioned, branch and
aperture transitions are discrete, and extreme intermediates can overflow.
Finite-kernel transforms do not change the public wrapper's scalar radius/plane
check, or the Newton solver's separate first-order differentiation contract.
CI/platform validation remains necessary.


HarrisonKramer:
Hi @BrunoChevalier,

Thanks for the contribution! The precision improvements and CPU performance gains are clear and solid. However, there is an architectural issue with the code as it currently stands:

1. **Backend Agnosticism**: In Optiland, geometry modules (`optiland/geometries/`) must remain backend-agnostic. Introducing `_conic_torch.py` with top-level `import torch` and custom `torch.autograd.Function` implementations directly in the geometry package violates this separation.
2. **Optional Dependency**: PyTorch is an optional dependency (`project.optional-dependencies`), and core geometry modules should not introduce direct PyTorch dependencies.
3. **Extensibility**: We do not want a pattern where each geometry requires a parallel `_<geometry>_torch.py` file. Any backend-specific execution optimizations or autograd mechanics should be routed through `optiland.backend`.

Please revise the PR to keep `optiland/geometries/` backend-agnostic, moving any backend-specific acceleration or custom autograd functions into `optiland.backend` or keeping the core mathematical solver unified across backends.

Happy to help here, if needed. Just let me know.

Thanks again,
Kramer

BrunoChevalier:
@HarrisonKramer Thanks for the feedback.

I'll rework this, taking that architecture into account.
I'll redo the speed comparison afterwards and will update the PR.

HarrisonKramer:
Thanks for reworking this, @BrunoChevalier. I reviewed the code and all looks good to me. I am okay to merge if it's ready to go. I'll wait for your confirmation though, just in case.

Kramer

BrunoChevalier:
I've added extra tests to increase the test coverage.
Once the codecov report is updated and the CI has finished, this is OK to merge for me @HarrisonKramer

PR #806, opened 2026-09-10
Reject ideal-material absorption loss in Zemax and CODE V exports

### LINKED ISSUE #802: Zemax and CODE V export silently discard ideal-material absorption
Zemax and CODE V export can silently remove absorption from an `IdealMaterial`, producing a valid-looking file with different optical behavior. The shared air classifier also treats a small but nonzero difference from unit index as air.

### Reproduction

Reproduced on upstream master `7df37590c2d561d95403dc264e27b8b9a77ee879`.

```python
from pathlib import Path
from tempfile import TemporaryDirectory

import optiland.backend as be
from optiland.fileio import save_codev_file, save_zemax_file
from optiland.fileio.common import is_air

be.set_backend("numpy")
be.set_precision("float64")

print(is_air(IdealMaterial(1.0, 0.01)))  # True upstream; should be False
print(is_air(IdealMaterial(1.0000001)))  # True upstream; should be False

optic = Optic()
optic.add_surface(index=0, radius=float("inf"), thickness=float("inf"))
optic.add_surface(
    index=1, radius=50, thickness=5,
    material=IdealMaterial(1.5, 0.01), is_stop=True,
)
optic.add_surface(index=2, radius=-50, thickness=50)
optic.add_surface(index=3, radius=float("inf"), thickness=0)
optic.set_aperture("EPD", 5)
optic.set_field_type("angle")
optic.add_field(y=0)
optic.add_wavelength(0.55, is_primary=True)

with TemporaryDirectory() as directory:
    for writer, suffix in ((save_zemax_file, ".zmx"), (save_codev_file, ".seq")):
        path = Path(directory) / ("existing" + suffix)
        path.write_bytes(b"previous optical design")
        writer(optic, str(path))  # Succeeds upstream, although k is not preserved.
        print(writer.__name__, "previous file preserved:",
              path.read_bytes() == b"previous optical design")
```

`is_air` checks the ideal index approximately and does not check extinction. Even for `n=1.5`, where the medium is classified as glass, the current writer encodings discard its nonzero `k`. The export calls above succeed and replace the destination.

### Expected behavior and proposed scope

- Classify an ideal medium as air only when its index is exactly one and its extinction is zero.
- Reject unsupported ideal-material absorption before opening or replacing the destination, with an actionable native-JSON alternative.
- Share the absorption check between both writer converters. Preserve their existing reflective-surface handling and supported lossless exports.
- Document this limitation and test absorbing media at both `n=1` and `n=1.5`, near-unit indices, and preservation of an existing output file.

This concerns the capabilities of Optiland's current encoders, not a claim that the external formats can never represent absorption.

### Validation

A correction is prepared locally on `fix/material-export-classification`, based directly on upstream master.

Fourteen regression cases failed before correction. The final affected writer/classification selection passes **129 tests** on the available NumPy/Torch CPU backends; changed-line coverage is **35/35 (100%)** under the existing Codecov policy. No application branch has been pushed for this issue yet.

Environment: Windows, Python 3.14.6, NumPy 2.3.5; CPU Torch was also exercised by the regression suite. Existing issues were searched before filing.

BrunoChevalier:
Zemax and CODE V export can silently remove absorption from an `IdealMaterial`. The shared air classifier ignores extinction and also rounds small nonzero differences from unit index away. An absorbing medium can therefore be exported as air or as glass without its extinction, while an existing destination is replaced.

Fixes #802.

Require exactly unit index and zero extinction for ideal air classification. Both writer converters now call a shared absorption check before glass encoding and before opening the destination. Unsupported ideal absorption raises `NotImplementedError` with a native-JSON alternative. Existing reflective-surface precedence and supported lossless exports are preserved. The file-I/O API documents the limitation.

### Validation

Fourteen regression cases failed before correction, covering NumPy/Torch, near-unit indices, absorbing ideal media at `n=1` and `n=1.5`, both writer formats and preservation of an existing destination. The final standalone writer/classification selection passed **129 tests**. Local patch coverage is **35/35 changed executable lines (100%)**, using unchanged Codecov policy. Ruff, formatting and whitespace checks pass.

This is one commit based directly on upstream `master` at `7df37590c2d561d95403dc264e27b8b9a77ee879`. It changes the current encoders' validation; it does not claim that the external formats can never represent absorption.

### DIFF
--- a/docs/api/api_fileio.rst
+++ b/docs/api/api_fileio.rst
@@ -7,6 +7,12 @@ The ``optiland.fileio`` package handles saving and loading of optical systems.
 It supports the native Optiland JSON format, as well as Zemax (.zmx), CODE V (.seq), 
 and OSLO (.len) files.
 
+Zemax and CODE V export reject an ``IdealMaterial`` with nonzero extinction
+before opening the destination file, because their current glass encodings do
+not preserve that absorption. Save the native JSON format to retain it.
+The shared writer air classifier requires exactly unit index and zero extinction;
+a small refractive-index difference is still an optical property.
+
 .. autosummary::
    :toctree: generated
    :nosignatures:
--- a/optiland/fileio/codev/writer/formatter.py
+++ b/optiland/fileio/codev/writer/formatter.py
@@ -15,7 +15,12 @@ from typing import TYPE_CHECKING, Any
 import optiland.backend as be
 from optiland.fileio.codev.model import CodeVDataModel
 from optiland.fileio.codev.surfaces import get_handler_for_optiland_type
-from optiland.fileio.common import compute_abbe_number, field_type_string, is_air
+from optiland.fileio.common import (
+    compute_abbe_number,
+    field_type_string,
+    is_air,
+    reject_unsupported_ideal_absorption,
+)
 from optiland.materials.material import Material
 
 if TYPE_CHECKING:
@@ -267,6 +272,7 @@ class OpticToCodeVConverter:
         if is_reflective:
             return {"name": "REFL"}
 
+        reject_unsupported_ideal_absorption(mat)
         if is_air(mat):
             return None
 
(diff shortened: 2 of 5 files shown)


HarrisonKramer:
Hi @BrunoChevalier,

Fix is clear. Thanks for this one. Merging.

Kramer

PR #809, opened 2026-09-10
Fix Windows notification clicks reaching the underlying console

### LINKED ISSUE #808: GUI error notifications cannot be dismissed on Windows: clicks reach the console underneath
### Bug report

On Windows, an error notification can remain visible over the docked Python console while clicks on its message or red X are handled by the console underneath. The notification cannot be dismissed normally. This is particularly disruptive for error notifications because they intentionally have no automatic timeout.

The original red X is the error-severity icon, rather than a separate close button, but the whole card is intended to support click-to-dismiss.

### Reproduction

1. Start the Optiland GUI on Windows with the Scripts Terminal/Console docked at the bottom-right, underneath the notification area.
2. In the embedded console, run:

   ```python
   iface.get_main_window().toast_manager.notify("Dismiss this error notification", "error")
   ```

3. Wait for the notification's entry animation to finish.
4. Click its red X or message text.

The console receives the click and the error notification stays visible. This was first observed after an OPD analysis reported that polarization must be set for polarization-dependent coatings, but the notification problem does not require that optical system or error.

**Expected:** clicking the notification dismisses it and consumes the click, without interacting with the underlying console. Error notifications should remain persistent until dismissal; changing the timeout policy is unnecessary.

### Environment

- Windows, using Qt's native Windows platform plugin
- Python 3.14.6
- PySide6/Qt 6.11.2
- Reproduced against upstream master `3dd53cbe35ae785d72cb10aea450799d9cede26b`

### Diagnosis and validation

`ToastWidget` sets `WA_TranslucentBackground`. Alongside native console/VTK widgets, the toast becomes a layered native child window on Windows. Actual Windows hit testing (`WindowFromPoint`) selects the underlying `QTextEdit`, even though the toast is rendered above it. Direct Qt test clicks on the toast's labels, or hit testing through `QApplication.widgetAt`, bypass this native behavior and miss the bug.

Removing that flag makes Windows select the notification as the mouse recipient. A fix is prepared, including an explicit accessible close button, consumed card clicks, native Windows regression coverage, and documentation. The native regression fails on the original code and passes with the fix; the reporter also confirmed that the fix resolves the problem in the full GUI.

I searched open and closed issues/PRs for toast, notification, dismissal, click-through and unclickable controls, reviewed GUI reports, and checked all current open PRs. I found no matching report or competing fix. The related historical PR is #541, which introduced toast notifications.

BrunoChevalier:
Fixes #808.

On Windows, clicking an error notification over the docked Python console can send the click to the console and leave the notification visible. Remove `WA_TranslucentBackground` from the notification child widget so native Windows hit testing selects the card. Add an accessible close button and consume card clicks after dismissal. Error persistence, seven-second non-error timeouts, and slide/fade animations are preserved.

The branch starts directly from current upstream `master` (`3dd53cbe`) and contains only the notification fix, related tests, and GUI quickstart/developer documentation.

Validation:

- GUI suite: **59 passed, 3 skipped** with Qt's offscreen platform; the three skips require native Windows hit testing.
- Native Windows regression: **3 passed** on this branch, covering icon, message, and close-button clicks. The test uses Windows `WindowFromPoint` before delivering input, so it catches click-through behavior that direct Qt label clicks miss. The icon/message regressions fail against the original upstream implementation.
- The full GUI check over the docked Jupyter console passes, and the reporter confirmed the fix resolves the issue.
- Card/detail/background clicks, keyboard close, timeouts, repeated dismissal, and stack cleanup are covered.
- Ruff lint/formatting and `git diff --check` pass; both edited reStructuredText documents parse without warnings.

Tested on Windows with Python 3.14.6 and PySide6/Qt 6.11.2. Native Windows tests are skipped on other platforms and with the offscreen Qt plugin.

### DIFF
--- a/docs/developers_guide/gui.rst
+++ b/docs/developers_guide/gui.rst
@@ -114,6 +114,18 @@ Key Widgets
     *   Similar to the command palette found in VS Code, it provides a centralized interface for discovering and running various tools, analyses, and layout operations.
     *   Triggered from the `MainWindow` via the `Ctrl+K` shortcut.
 
+*   **Toast Notifications (``widgets/toast.py``)**:
+
+    *   ``MainWindow`` owns a ``ToastManager``, also exposed as
+        ``connector.toast_manager``. Call ``notify(message, severity)`` for
+        information, success, warning, or error feedback.
+    *   Cards can be dismissed by clicking them or their accessible close button.
+        Non-error cards expire after seven seconds; errors persist until dismissed
+        or evicted from the three-card stack.
+    *   Keep these child widgets free of ``WA_TranslucentBackground``. Alongside
+        native console/VTK widgets on Windows, this flag creates a layered native
+        child window whose mouse hit testing can select the underlying console.
+
 Styling and Resources
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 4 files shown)


HarrisonKramer:
Hi @BrunoChevalier,

Thanks for the fix! merging now.

Kramer

PR #811, opened 2026-09-10
fix(wavefront): correct launch phase for infinite conjugates

### LINKED ISSUE #810: Spurious OPD for infinite-conjugate fields with vignetting or non-unit object index
# Spurious OPD for infinite-conjugate fields with vignetting or non-unit object index

## Summary

For an object at infinity, `OPD` can report a nonzero wavefront error for an
oblique plane wave passing through a homogeneous, index-matched system. The
minimal cases below use either field vignetting in air or an object-space
refractive index other than 1.

Reproduced on upstream commit
`7df37590c2d561d95403dc264e27b8b9a77ee879`.

With a 5-degree Y field, a 4 mm entrance pupil, and a wavelength of 0.55 um,
25% Y-pupil compression in air produces 158.464987 waves peak-to-valley.
An unvignetted system with index 1.5 produces 316.929974 waves. Both should
produce zero OPD relative to the chief-ray-normal reference plane.

## Minimal Reproducer

The model uses only `IdealMaterial`; no material catalog is required. The
script prints all four cases before asserting the expected flat wavefront.

```python
import numpy as np

import optiland.backend as be
from optiland.wavefront import OPD


def collimated_planes(index, vignette):
    medium = IdealMaterial(index)
    optic = Optic()
    optic.surfaces.add(index=0, thickness=be.inf, material=medium)
    optic.surfaces.add(index=1, thickness=10.0, material=medium, is_stop=True)
    optic.surfaces.add(index=2, thickness=10.0, material=medium)
    optic.surfaces.add(index=3, material=medium)
    optic.set_aperture("EPD", 4.0)
    optic.fields.set_type("angle")
    optic.fields.add(y=0.0)
    optic.fields.add(y=5.0, vx=0.0, vy=vignette)
    optic.wavelengths.add(0.55, is_primary=True)
    return optic


be.set_backend("numpy")
be.set_precision("float64")
failures = []

for index, vignette in ((1.0, 0.0), (1.0, 0.25), (1.5, 0.0), (1.5, 0.25)):
    optic = collimated_planes(index, vignette)
    analysis = OPD(
        optic,
        (0.0, 1.0),
        "primary",
        num_rays=6,
        strategy="chief_ray",
        afocal=True,
        remove_tilt=False,
    )
    data = analysis.get_data((0.0, 1.0), 0.55)
    opd = be.to_numpy(data.opd)
    intensity = be.to_numpy(data.intensity)
    max_error = float(np.max(np.abs(opd)))
    print(
        f"n={index}, vignetting={vignette}: "
        f"OPD peak-to-valley={np.ptp(opd):.6f} waves"
    )
    if (
        not np.all(intensity > 0)
        or not np.all(np.isfinite(opd))
        or max_error >= 1e-10
    ):
        failures.append((index, vignette, max_error))

assert not failures, failures
```

## Observed And Expected Behavior

Observed on the pinned commit:

| Object index | Y vignetting | Observed OPD peak-to-valley (waves) | Expected |
| ---: | ---: | ---: | ---: |
| 1.0 | 0.00 | 0.000000 | 0 |
| 1.0 | 0.25 | 158.464987 | 0 |
| 1.5 | 0.00 | 316.929974 | 0 |
| 1.5 | 0.25 | 79.232493 | 0 |

All surfaces are planar, all regions have the same index, and there is no
physical clipping. The input bundle therefore remains collimated. With
`afocal=True` and the chief-ray reference, OPD should be zero to numerical
precision without removing a fitted tilt.

As a separate diagnostic, there was no spread in the post-launch accumulated
optical path for these plane-stack cases. This is consistent with the error
arising in launch/reference-phase bookkeeping rather than propagation between
the parallel planes.

## Suspected Cause

`ReferenceStrategy._correct_tilt` reconstructs the incident phase term using
nominal pupil-distribution coordinates:

```python
X_m = xs * epd / 2
Y_m = ys * epd / 2
tilt = ux * X_m + uy * Y_m
return opd + tilt
```

For a plane wave, the required relative incident optical path is instead

```text
dOPL_launch = n_object(wavelength) * u . (r_launch - r_reference_launch)
```

The existing expression omits the object-space refractive index and uses
nominal offsets rather than the offsets actually produced by ray generation.
For paraxial aiming, vignetting scales those offsets by `1 - vx` and `1 - vy`.

For the one-axis example above, the resulting predicted peak-to-valley error is

```text
abs(n_object * (1 - v) - 1) * sin(field_angle) * EPD / wavelength_mm
```

This agrees with all four reported values. The index and vignetting errors
partially cancel in the last case; they do not necessarily add.

Aiming strategies that generate launch positions different from the nominal
paraxial-pupil reconstruction may expose the same inconsistency. This is
supporting context rather than the correctness oracle for the reproducer.

## Suggested Direction

Evaluate the incident phase from the actual generated launch geometry, using
the object-space index at the analysis wavelength. This avoids duplicating
pupil-scaling and vignetting logic in wavefront analysis.

The correction should remain separate from optical path added during
propagation and surface interactions. Removing a fitted output tilt is not a
substitute for consistent incident phase.

## Environment

Optiland source checkout: `7df37590c2d561d95403dc264e27b8b9a77ee879`.
Python 3.14.2 on Windows; NumPy 2.5.3 and Torch 2.13.0 CPU, float64.
Both backends agree to the displayed precision.

mabl:
## Summary

- restore relative incident optical path from the generated launch positions and directions for infinite-conjugate angular fields
- use the launch medium's refractive index, including sequences that begin after an omitted transition
- retain copied launch state only when requested and release it after wavefront phase restoration
- add regressions for signed and diagonal fields, reverse propagation, vignetting, ray aiming, reference strategies, sequence entry media, pupil ordering, physical surface phase, and Torch gradients

Closes #810.

## Validation

- `331 passed, 1 skipped` across focused wavefront, ray, sequence, and golden-regression tests on NumPy and Torch
- `ruff check optiland/`
- `ruff format --check optiland/`
- `git diff --check`


HarrisonKramer:
The physical wavefront launch phase correction looks great, but CI failed on `tests/test_folded_paraxial_finalization.py::TestTangentSingularityRejection::test_wavefront_tilt_correction_is_guarded` due to `ImportError: cannot import name 'require_nonsingular_tangent_angles' from 'optiland.wavefront.strategy'`.

Could you either re-export `require_nonsingular_tangent_angles` in `optiland/wavefront/strategy.py` or update the test import so CI can pass?

mabl:
> Could you either re-export `require_nonsingular_tangent_angles` in `optiland/wavefront/strategy.py` or update the test import so CI can pass?

Done. But likely needs a rebase onto master after #837 is merged.

mabl:
@HarrisonKramer  Rebased and tests are passing.

Issue #337, opened 2025-09-27
Feature Request: Add support for Gradient Index (GRIN) surfaces


goldengrape:
### Checklist

- [x] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or feature request.
- [x] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [x] I am using the **latest version** of Optiland.
- [x] I have included **all necessary context**.

*Thanks for taking the time to go through this — it really helps us help you!*

### Feature Request

Of course. Here is the final draft of the issue, written in English and incorporating all the details you've provided. You can copy and paste this directly into the GitHub issue submission form.

-----

### Feature Request

**Is your feature request related to a problem? Please describe.**
Currently, Optiland lacks support for surfaces with a gradient refractive index (GRIN), such as the `gradient3` surface type available in Zemax. This limits the software's capability to design and simulate optical systems that utilize GRIN lenses. These lenses are crucial for correcting aberrations and reducing the number of optical elements in a system.

Furthermore, GRIN lenses are of significant importance in the fields of ophthalmology and optometry for simulating realistic eye models. Many established models, such as the Atchison and Liou-Brennan model eyes, rely on gradient refractive indices to accurately represent the human eye's crystalline lens. The absence of this feature in Optiland is a considerable limitation for researchers and engineers working on ophthalmic and vision science applications.

**Describe the solution you'd like**
I would like to request the addition of a new surface type that supports a gradient refractive index, similar to Zemax's `gradient3`. This would involve implementing a ray tracing algorithm capable of handling propagation through a medium with a continuously varying refractive index.

**Describe alternatives you've considered**
I have considered using a series of shells with uniform refractive indices to approximate a gradient index. However, this method is computationally intensive and would significantly slow down simulations, making it an impractical alternative for complex or iterative design processes.

**Additional context**
The importance of GRIN lenses is particularly evident in the simulation of human eye models, where they are used to accurately model the crystalline lens. For example, the Atchison (doi:10.1016/j.visres.2006.01.004) and Liou-Brennan model eyes (doi: 10.1364/josaa.14.001684) both utilize gradient-index lenses, which is a key reason for their anatomical and optical accuracy. Adding this functionality would make Optiland a much more powerful tool for ophthalmic and vision science research.

[...]

PR #813, opened 2026-09-11
fix(raytrace): preserve signed optical path for virtual propagation

### LINKED ISSUE #812: Virtual propagation incorrectly adds positive optical path
# Virtual propagation incorrectly adds positive optical path

## Summary

For a negative propagation parameter `t`, `Surface._trace_real` moves the ray
backward along its direction but **increases** its accumulated optical path,
because the contribution is calculated as `abs(n * t)`.

For a normalized ray in a homogeneous positive-index medium, the signed
propagation contribution should be `n * t`. Virtual propagation should reduce
the accumulated optical path. Real propagation toward negative global Z should
still increase it.

Reproduced on upstream `master` commit
`b9127f76b7805c6bf572de7eca3d2b063eaa551a`, with both NumPy and Torch.

## Minimal reproducer

This uses an index-matched plane and three unit rays: ordinary forward travel,
virtual travel, and real travel toward negative Z. It exercises the public
`Surface.trace` API and requires no material catalog or wavefront initialization.
Here `rays.opd` is the tracing-time optical-path accumulator, in millimeters.

```python
import numpy as np

import optiland.backend as be
from optiland.coordinate_system import CoordinateSystem
from optiland.geometries import Plane
from optiland.rays import RealRays
from optiland.surfaces import Surface

failures = []
    be.set_backend(backend)
    be.set_precision("float64")
    medium = IdealMaterial(1.5)
    plane = Plane(CoordinateSystem())  # z = 0
    surface = Surface(None, medium, plane)  # Index-matched interface.

    rays = RealRays(
        x=be.zeros(3),
        y=be.zeros(3),
        z=be.array([-5.0, 5.0, 5.0]),
        L=be.zeros(3),
        M=be.zeros(3),
        N=be.array([1.0, 1.0, -1.0]),
        intensity=be.ones(3),
        wavelength=be.full(3, 0.55),
    )
    t = be.to_numpy(plane.distance(rays)).copy()
    surface.trace(rays)
    actual = be.to_numpy(rays.opd)
    expected = 1.5 * t
    print(f"{backend}: t={t}, OPL={actual}, expected={expected}")
    np.testing.assert_allclose(be.to_numpy(rays.z), 0.0, rtol=0, atol=1e-12)
    if not np.allclose(actual, expected, rtol=0, atol=1e-12):
        failures.append(backend)

assert not failures, f"Incorrect signed OPL on: {failures}"
```

Output:

```text
numpy: t=[ 5. -5.  5.], OPL=[7.5 7.5 7.5], expected=[ 7.5 -7.5  7.5]
torch: t=[ 5. -5.  5.], OPL=[7.5 7.5 7.5], expected=[ 7.5 -7.5  7.5]
AssertionError: Incorrect signed OPL on: ['numpy', 'torch']
```

All rays reach the correct plane. Only the virtual ray has the wrong OPL sign.

## Expected behavior and sign convention

For a normalized direction vector, write the ray as

$$
\mathbf r(t)=\mathbf r_0+t\mathbf d,\qquad \|\mathbf d\|=1.
$$

Here `t` is an **oriented ray parameter**, measured in millimeters along the
current ray direction. It can be negative for virtual propagation; it is not
the nonnegative physical arclength of a traveled segment. In a homogeneous
medium, the oriented optical-path increment is

$$
\Delta S=\int_0^t n\,d\tau=nt.
$$

[`Plane.distance`](https://github.com/optiland/optiland/blob/b9127f76b7805c6bf572de7eca3d2b063eaa551a/optiland/geometries/plane.py#L72-L88)
returns `t = -rays.z / rays.N`. Thus `z=5`, `N=1` gives `t=-5 mm`,
and `n=1.5` requires an increment of `-7.5 mm`, rather than `+7.5 mm`.
There is no quadratic-root selection involved.

The sign is relative to the **current ray direction**. After a physical mirror
reflection, a returning ray can have `N < 0` while still advancing with `t > 0`.
Its propagation contribution remains positive. In a physical mirror round
trip, the **propagation optical paths of both real legs add**; any interface
or mirror phase is a separate contribution. A virtual extrapolation followed
by its inverse, with no intervening interaction, must cancel to numerical
precision.

## Impact and additional checks

The error need not be a common piston. For a virtual plane separation `D`,
`t_i = -D / N_i`, so rays with different angles receive different errors.

An end-to-end `OPD` check used a finite on-axis object, an entirely homogeneous
`n=1.5` medium, a stop 100 mm after the object, a 4 mm EPD, and an image plane
20 mm after the stop. The comparison inserted two index-matched dummy planes
10 mm and 5 mm after the stop, giving this sequential traversal relative to
the stop:

```text
Baseline:  0 -> 20 mm
Detour:    0 -> 10 -> 5 -> 20 mm
```

With wavelength 0.55 µm, `num_rays=6`, `strategy="chief_ray"`, `afocal=True`
(planar reference), and `remove_tilt=False`, **final ray positions and directions
are unchanged to numerical precision**, but the **difference between the two
OPD arrays has 5.454000109 waves peak-to-valley**. This is optical-path/phase
corruption with preserved ray geometry. The two wavefronts should agree; the
baseline wavefront itself is not required to be flat.

The following targeted diagnostics agree on both backends. The last column
uses an isolated in-memory replacement of the accumulation expression with
`n * t`:

| Check | Upstream | Signed-expression probe / expected |
| --- | ---: | ---: |
| Single virtual segment, `n=1.5`, `t=-5 mm` | `+7.5 mm` | `-7.5 mm` |
| Virtual inverse excursion, rays with `N=1` and `N=0.8` | `[15, 18.75] mm` net | `[0, 0] mm` net |
| Propagation OPL of a physical mirror round trip, two real 5 mm legs | `15 mm` | `15 mm` |
| OPD change caused by the dummy-plane detour | `5.454000109 waves PV` | approximately `1e-10 waves PV` numerical residual |
| Torch `d(OPL)/dt` at `t=-5 mm`, fixed `n=1.5` | `-1.5` | `+1.5` |

A thin-lens interaction control also retained the same additive surface-phase
contribution with the signed-expression probe.

## Reference behavior

The signed-parameter expectation also agrees with the
[OpticStudio 2026 R1.03 numeric-function documentation](https://ansyshelp.ansys.com/public/Views/Secured/Zemax/v26103/en/OpticStudio_User_Guide/OpticStudio_Help/topics/Numeric_Functions.html):

> RAYO: The optical path length is the path length times the index of refraction,
> either or both of which may be negative.

`RAYT` explicitly permits negative path lengths. The same definitions appear
in the earlier manual (*ZEMAX User's Guide v9.0*, Chapter 24, pp. 366–367:
`RAYO`, `RAYT`).

The [polarization documentation](https://ansyshelp.ansys.com/public/Views/Secured/Zemax/v26103/en/OpticStudio_User_Guide/OpticStudio_Help/topics/Definition_of_Polarization_Terms.html)
also specifies negative path length for virtual propagation and uses path
length in propagation phase factors. These are documentation references;
the numerical reproduction above is from Optiland.

## Cause and suggested correction

[`Surface._trace_real`](https://github.com/optiland/optiland/blob/b9127f76b7805c6bf572de7eca3d2b063eaa551a/optiland/surfaces/standard_surface.py#L300-L306)
currently contains:

```python
t = _aperture_aware_distance(self, rays)
self.material_pre.propagation_model.propagate(rays, t)
rays.opd = rays.opd + be.abs(t * self.material_pre.n(rays.w))
```

If the existing homogeneous propagation contract is retained, the local
correction is:

```python
rays.opd = rays.opd + t * self.material_pre.n(rays.w)
```

Regression coverage should include the negative-`t` plane case, virtual inverse
cancellation, a real reflected path, pupil-dependent OPD, Torch gradients, and
preservation of additive surface phase.

## Related work

This is separate from the infinite-conjugate launch-phase issue #810 and its
proposed fix #811: the minimal reproducer directly initializes `RealRays`, and
the end-to-end OPD check uses a finite object. Neither needs an
infinite-conjugate incident-phase correction.

Moving optical-path integration into the propagation abstraction is related
to GRIN support (#337), but a `propagate(rays, t) -> delta_opl` API would not
by itself address the general GRIN case. For homogeneous media, geometry can
determine a scalar signed `t` before propagation. For GRIN media, the trajectory
is curved, so determining when it reaches the exit geometry and evaluating
$\int n(\mathbf r(s),\lambda)\,ds$ are coupled problems that naturally belong
to the same propagation process. The maintainer's
[proposal in #337 to pass exit geometry to the propagator](https://github.com/optiland/optiland/issues/337#issuecomment-3478080904)
is therefore the more appropriate architectural direction. This does not need
to block the local homogeneous `n*t` correction.

## Environment

- Upstream source: `b9127f76b7805c6bf572de7eca3d2b063eaa551a`.
- Python 3.14.2 on Windows.
- NumPy 2.5.3; Torch 2.13.0+cpu.
- Both backends tested with float64.
- Validation: targeted analytical, surface-interaction, autograd, and OPD checks.

mabl:
## Summary

- Preserve the signed homogeneous propagation contribution, `t * material_pre.n(w)`, in `Surface._trace_real`.
- Document that virtual propagation subtracts optical path, while physical reflected return legs have positive travel along the current direction and add optical path.
- Add NumPy/Torch regressions for all four travel/direction sign combinations, incident-medium selection, inverse cancellation, actual mirror reflection, gradients through zero travel, additive surface phase, real/virtual thin-lens focus, and wavefront invariance under an index-matched virtual detour.

Closes #812.

## Correctness

The intersection parameter is oriented along the current normalized ray. Taking its absolute optical path changes a virtual segment from `n*t` to `-n*t` and reverses its local gradient. A real return path after reflection still has `t > 0`, even when global Z decreases.

The finite-object wavefront regression checks pointwise OPD equality and preserved final geometry after inserting index-matched dummy planes. It also requires angular diversity so the original error cannot disappear as common piston. The virtual-focus regression couples pupil-dependent negative travel to the actual thin-lens phase contribution.

This homogeneous correction is independent of the launch-phase work in #810/#811 and the propagation API design for GRIN in #337.

## Validation

Local validation on pinned upstream base `b9127f76b7805c6bf572de7eca3d2b063eaa551a`, using Python 3.14.2 on Windows, NumPy 2.5.3 and Torch 2.13.0+cpu in float64:

| Run | Result |
| --- | --- |
| All eight new regression methods against pristine original production | 14 failed, 7 passed, 3 skipped |
| Same methods against the fix | 21 passed, 3 skipped |
| Surrounding test modules and all of `tests/regression/` | 460 passed, 4 skipped |

The targeted runs checked that production imports originated from the selected pristine/fixed checkout. All Torch parameterizations executed; the three targeted skips are NumPy autograd cases. All eight golden snapshot cases passed.

Additional in-memory fault-injection checks verified that the regressions reject transmitted-index accumulation, a collimated wavefront fixture, incorrect negative-Z-direction sign handling, and omitted intermediate OPL accumulation.

Checks passed:

- `ruff check optiland/`
- `ruff format --check optiland/surfaces/standard_surface.py`
- `git diff --check`

<details>
<summary>Surrounding test command</summary>

```bash
.venv/Scripts/python.exe -m pytest -q --tb=short \
  tests/test_standard_surface.py \
  tests/test_object_surface.py \
  tests/test_image_surface.py \
  tests/test_surface_group.py \
  tests/test_surface_factory.py \
  tests/test_rays.py \
  tests/propagation/test_homogeneous.py \
  tests/test_wavefront.py \
  tests/test_wavefront_strategy.py \
  tests/test_thin_lens_interaction_model.py \
  tests/test_phase_interaction_model.py \
  tests/sequences/test_surface_view.py \
  tests/sequences/test_sequenced_optic.py \
  tests/sequences/test_sequenced_surface_group.py \
  tests/regression/
```

</details>

### DIFF
--- a/optiland/surfaces/standard_surface.py
+++ b/optiland/surfaces/standard_surface.py
@@ -299,7 +299,9 @@ class Surface(ObserverMixin):
         """
         t = _aperture_aware_distance(self, rays)
         self.material_pre.propagation_model.propagate(rays, t)
-        rays.opd = rays.opd + be.abs(t * self.material_pre.n(rays.w))
+        # t is oriented along the ray: virtual propagation subtracts OPL.
+        # Real return paths after reflection still have t > 0 and add OPL.
+        rays.opd = rays.opd + t * self.material_pre.n(rays.w)
         if self.aperture:
             self.aperture.clip(rays)
         rays = self.interaction_model.interact_real_rays(rays)
--- a/tests/test_standard_surface.py
+++ b/tests/test_standard_surface.py
@@ -10,6 +10,7 @@ from optiland.interactions.refractive_reflective_model import RefractiveReflecti
 from optiland.materials import IdealMaterial
 from optiland.rays import ParaxialRays, RealRays
 from optiland.surfaces.standard_surface import Surface
+from tests.utils import assert_allclose
 
 
 class TestSurface:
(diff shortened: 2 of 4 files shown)


HarrisonKramer:
Hi @mabl,

Fix makes sense. Thanks for submitting. Merging.

Kramer

PR #824, opened 2026-09-11
Keep image-plane markers stable and exclude extinct ray extents

### LINKED ISSUE #823: Extinct rays enlarge image-plane markers and distort layout auto-fit
Traced rays can retain finite coordinates after an upstream aperture extinguishes them. Using all stored coordinates to size downstream drawing surfaces allows these nonphysical points to enlarge an image-plane marker as the ray count changes. Raw first-blocking hits can also dominate auto-fit despite a small explicit aperture, and tilted image markers can be cropped when horizontal limits use only surface vertices.

Expected: sampled extents include only valid incident rays, explicit physical apertures remain authoritative, and an image plane without a specified physical size has a stable, clearly identified schematic marker. Display sampling must not invent or modify a detector size. Auto-fit should include the resolved geometry, including tilted image markers.

Reproduce with a generated sequential system containing a finite clipping aperture and an unbounded image endpoint; vary the displayed ray count and compare marker size, layout limits and unchanged prescription.

BrunoChevalier:
Downstream coordinates of already extinguished rays currently enlarge image-plane drawings and layout limits. Compute sampled extents from finite incident hits with cumulative ray liveness, retaining the first blocking endpoint while excluding later coordinates. Reset sampling between plots and copy coordinates before display masking.

Keep finite physical apertures authoritative. Give an unbounded image endpoint a ray-count-independent schematic marker based on the nearest preceding finite aperture, or a 1 mm fallback, and identify its schematic meaning in labels and surface information. Use resolved component extents for auto-fit and include transformed outlines so tilted markers are visible. The prescription and detector/aperture settings are unchanged.

Validation: 307 affected tests passed on NumPy/Torch, including 3D actor bounds; 100% of changed executable production lines covered; Ruff and formatting passed. Full CI/Codecov results will be checked on this PR.

Closes #823

### DIFF
--- a/optiland/visualization/info/providers.py
+++ b/optiland/visualization/info/providers.py
@@ -47,6 +47,8 @@ class SurfaceInfoProvider(BaseInfoProvider):
 
         if hasattr(surface, "comment") and surface.comment:
             info.append(f"Comment: {surface.comment}")
+        if obj.extent_source == "schematic":
+            info.append("Image plane marker: schematic; physical size unspecified")
         if hasattr(surface, "geometry") and hasattr(surface.geometry, "radius"):
             info.append(f"Radius: {surface.geometry.radius:.3f}")
         if hasattr(surface, "geometry") and hasattr(surface.geometry, "conic"):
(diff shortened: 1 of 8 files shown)


HarrisonKramer:
@BrunoChevalier - can you rebase this one as well? there's a conflict because of another PR hitting the rays.py file.

Thanks,
Kramer

BrunoChevalier:
@HarrisonKramer done

PR #837, opened 2026-09-12
Fix IdealMaterial precision loss in mixed-precision object-NA calculations

### LINKED ISSUE #832: Material index downcast breaks mixed-precision object-NA calculations on NumPy and Torch
## Bug

Creating an `IdealMaterial` with a float64 index, then switching the backend default to float32, causes `material.n()` to return float32. This breaks the existing mixed-precision object-NA contract: a Python scalar NA loses the precision of its float64 material partner. In the near-grazing case, a valid `NA < n` is rounded to `NA == n` and rejected.

This produces **six existing test failures on both NumPy and Torch CPU**. They were noticed in #804 CI but also reproduce on unmodified upstream master.

## Reproduce

With the project's test dependencies and Torch installed, run from the repository root:

```sh
python -m pytest tests/test_telecentric_object_na.py -k float64_operand_preserves_python_partner -q
```

The existing [regression test](https://github.com/optiland/optiland/blob/66d5caa2805232fb57cdf2ca826a6922ffcaf04c/tests/test_telecentric_object_na.py#L333-L383) creates the material/index at float64, switches the default to float32, and evaluates `object_space_sine()` and `compute_epd()`. It checks output dtype, numerical results, and Torch gradients.

Observed on each backend:

| Float64 operand | Python scalar partner | Result |
| --- | --- | --- |
| Index = 1.5 | NA = 0.6 | Fails: launch sine has float32 dtype, expected float64 |
| Index = 1.5 | NA = 1.4999 | Fails: launch sine has float32 dtype, expected float64 |
| Index = 1.5 | NA = 1.49999999 | Fails: `ValueError: Object NA must be finite and satisfy 0 <= NA < n.` |
| NA = 1.5 | Index = 1.50000001, supplied by the test's mocked `material.n()` | Passes |

The focused selection reports **6 failed, 2 passed, 80 deselected** on both:

- Unmodified upstream master baseline `66d5caa2805232fb57cdf2ca826a6922ffcaf04c`.
- PR #804 head `163dcbe82e0048e069a2edd8a1ac94fd08ed80ac`.

The material/aperture modules and this test file also have no differences between that PR head and the subsequently checked master `cbd5af715e21ab519ff162677b19051dbb417c00`.

## Expected behavior

The existing regression expects the float64 operand to preserve the Python partner's precision after the default changes. The result should be a finite float64 scalar with `sin(theta) = NA / n`, and the corresponding entrance pupil diameter should agree with the analytical expression in the test. `1.49999999 < 1.5` is a valid forward cone and should not become an invalid grazing cone through an unintended downcast. Torch differentiation should remain connected to the original parameter.

## Diagnosis

[`IdealMaterial._calculate_n()`](https://github.com/optiland/optiland/blob/66d5caa2805232fb57cdf2ca826a6922ffcaf04c/optiland/materials/ideal.py#L46-L63) calls [`BaseMaterial._as_backend_array()`](https://github.com/optiland/optiland/blob/66d5caa2805232fb57cdf2ca826a6922ffcaf04c/optiland/materials/base.py#L251-L256), which uses `be.asarray(value)` with the active default precision. The stored parameter remains float64, but the returned index is float32. By the time `ObjectNAAperture.object_space_sine()` adopts the index dtype for its Python NA partner, the higher precision has already been lost.

The fix needs to reconcile material conversion and aperture precision semantics while preserving backend/device conversion and Torch gradients. Merely relaxing the dtype assertion would leave the valid-cone rejection unresolved.

## CI evidence and environment

PR #804 run `34712689479` reports the same six failures on [Python 3.11](https://github.com/optiland/optiland/actions/runs/34712689479/job/103604237269) and [Python 3.14](https://github.com/optiland/optiland/actions/runs/34712689479/job/103604237282). Each job otherwise reports **10,436 passed, 180 skipped, 2 xfailed**.

Local A/B reproduction: Windows, Python 3.14.6, NumPy 2.5.2, Torch 2.13.0+cu130, with NumPy and Torch CPU execution. CUDA execution was not tested for this issue.

## Related tracking

Searched open issues and pull requests for telecentric, float32/float64, precision, IdealMaterial, and the failing test name. No open report or fix covering this failure was found. Closed #797 concerned the missing object-medium factor in telecentric launch; closed #794 concerned stale material caches. This report tracks the mixed-precision interaction now exposed by the existing tests.

BrunoChevalier:
Fixes #832.

Creating an `IdealMaterial` at float64 and then selecting float32 previously downcast its returned index. This rounded valid near-grazing object-NA cones into invalid grazing cones and broke six existing NumPy/Torch regression cases.

Preserve the stored index and extinction-coefficient dtype during IdealMaterial backend/device conversion and wavelength broadcasting. Keep the original parameter storage and Torch gradient graph, retain other material models' conversion policy, and document the precision contract. Add regressions for scalar and array wavelengths, cache hits, backend round trips, and repeated differentiation.

Validation:

- 755 tests passed, 21 skipped across focused materials, apertures, and backends; all six existing failures now pass.
- 86 tests passed, 24 skipped in a fresh process with Torch imports blocked.
- Local patch coverage: **100% (9/9 changed executable lines)** against `cbd5af71`, from 332 passing tests and 20 skips. Coverage configuration and exclusions are unchanged.
- Ruff check and formatting checks pass. CUDA regression cases are included but skipped in the local CPU-only environment.

This follows up on merged #805 and applies independently of #804. The open-PR overlap audit found no competing implementation.

### DIFF
--- a/docs/api/api_materials.rst
+++ b/docs/api/api_materials.rst
@@ -107,6 +107,13 @@ track their live parameter arrays; changing a parameter invalidates old results.
 Their calculations convert parameters for the active backend without replacing
 the original arrays, preserving existing Torch parameter identity and gradients.
 
+``IdealMaterial.n`` and ``IdealMaterial.k`` retain the dtype of their stored index
+and extinction coefficient, including for array-valued wavelength queries.
+Changing the default backend precision after creating the material does not
+recast these parameters or their returned values. Conversion still follows the
+active backend and device. This preserves mixed-precision calculations such as
+object-space NA with a float64 material and a Python scalar aperture value.
+
 Each concrete custom material opts into caching by implementing ``_cache_state``.
 An immutable model may return ``()``. A mutable model can use
 ``self._state_key((self.parameter, ...))`` for all numerical state used by its
--- a/optiland/materials/base.py
+++ b/optiland/materials/base.py
@@ -249,10 +249,16 @@ class BaseMaterial(ABC):
         return backend, be.get_precision(), device, gradients, inference
 
     @staticmethod
-    def _as_backend_array(value):
-        """Use live parameters in the active backend without replacing their storage."""
+    def _as_backend_array(value, *, preserve_dtype: bool = False):
+        """Use live parameters in the active backend without replacing their storage.
+
+        Typed parameters can opt out of conversion to the backend's default
+        precision. Untyped values still use that default.
+        """
         if be.get_backend() == "numpy" and hasattr(value, "detach"):
             value = value.detach().cpu().numpy()
+        if preserve_dtype and hasattr(value, "dtype"):
+            return be.asarray(value, dtype=None)
         return be.asarray(value)
 
     @staticmethod
--- a/optiland/materials/ideal.py
+++ b/optiland/materials/ideal.py
@@ -23,6 +23,9 @@ class IdealMaterial(BaseMaterial):
     """Represents an ideal material with a fixed refractive index and extinction
     coefficient for all wavelengths.
 
+    Evaluation preserves each stored parameter's dtype when the backend's default
+    precision changes, while converting to the active backend and device.
+
     Attributes:
         index (float): The refractive index of the material.
         absorp (float): The extinction coefficient of the material.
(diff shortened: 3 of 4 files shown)


mabl:
Thanks! I also just looked into it, and came to the same conclusion.

PR #838, opened 2026-09-13
test: restore tests hidden by duplicate definitions

Anai-Guo:
## Summary

A few test names in `tests/` are defined twice in the same scope. Python keeps only the last definition, so pytest never collects the earlier one. `pyflakes` reports these as `redefinition of unused ...`. They're hidden from lint here because `tests` is in ruff's `exclude`.

| File | Name | What was lost | Change |
|---|---|---|---|
| `test_operand.py` | `TestParaxialOperand.test_magnification` | The second copy asserts `total_track`, so the real magnification test never ran | Rename the second one to `test_total_track` |
| `test_analysis.py` | `class TestThroughFocusSpotDiagram` (x2) | The first class (older API) was never collected | Remove it, but move `test_image_surface_z_restoration` into the live class |
| `test_analysis.py` | `test_view_encircled_energy_larger_fig` (x2), `test_view_distortion_larger_fig` (x3) | Nothing (the copies are byte-identical) | Remove the extra copies |
| `test_geometries.py` | `TestForbesValidation.test_qnm_values_against_analytical_formula` (x2) | Nothing (same assertions, one copy has no comments) | Remove the comment-less copy |

### About the old `TestThroughFocusSpotDiagram`

I ran the shadowed class under a temporary name against current `master`:

- `test_init`, `test_analysis_results_content` and `test_view_method` **fail**. They expect the old API: dict-of-RMS results, a printed `view()`, and an even `num_steps=2`, which is now rejected with `'num_steps' must be an odd integer`. The live class already covers the current behaviour, so these three are dropped.
- `test_image_surface_z_restoration` **passes** and checks something the live class doesn't: the image surface `z` is restored after the through-focus sweep. I kept it and moved it verbatim into the live class.

## Verification

Run locally (numpy backend, `-k "not torch"`, since torch isn't installed in my env):

- `pytest --collect-only`: `test_analysis.py` 83 → 84, `test_operand.py` 86 → 87, `test_geometries.py` 127 → 127
- `pytest tests/test_analysis.py tests/test_operand.py tests/test_geometries.py -k "not torch"`: **286 passed** before, **288 passed** after. The two newly collected tests are `test_total_track` and `test_image_surface_z_restoration`, and both pass.
- `pyflakes` no longer reports any redefinitions in these three files.

The diff only touches tests.

### DIFF
--- a/tests/test_analysis.py
+++ b/tests/test_analysis.py
@@ -198,15 +198,6 @@ class TestCookeTripletEncircledEnergy:
         assert isinstance(ax, Axes)
         plt.close(fig)
 
-    def test_view_encircled_energy_larger_fig(self, set_test_backend, cooke_triplet):
-        encircled_energy = analysis.EncircledEnergy(cooke_triplet)
-        fig, ax = encircled_energy.view(figsize=(20, 10))
-        assert fig is not None
-        assert ax is not None
-        assert isinstance(fig, Figure)
-        assert isinstance(ax, Axes)
-        plt.close(fig)
-
 
 class TestCookeTripletRayFan:
     def test_ray_fan(self, set_test_backend, cooke_triplet):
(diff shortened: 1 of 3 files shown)


HarrisonKramer:
Hi @Anai-Guo,

Thanks for opening this. I see CI failing. Can you rebase to master and check the test status / push updates? Let me know if you need any help here.

thanks,
Kramer

PR #840, opened 2026-09-13
refactor(gui): remove dead duplicated methods in AnalysisPanel

Anai-Guo:
### Problem

`AnalysisPanel` (`optiland_gui/analysis_panel.py`) defines two methods
twice in the same class body, so Python keeps only the **last** binding and
the earlier copies are dead:

- **`_set_combobox_value`** — the first copy is byte-for-byte equivalent to
  the second (only the docstring wording differs).
- **`_load_analysis_settings_slot`** — the first copy is superseded by a
  later, more complete implementation (adds `analysis_name` handling and a
  `toast_manager` fallback). The "Load Settings" button is wired to the
  *later* one (`btnLoadSettings.clicked.connect(self._load_analysis_settings_slot)`),
  so the first never runs.
- **`_apply_loaded_settings_to_ui`** — only referenced from the shadowed
  load slot above, hence unreachable. It also calls
  `self._set_widget_value(widget, value)` with two arguments against the
  three-argument signature `_set_widget_value(self, widget, value, param_name)`,
  so it would raise `TypeError` if it were ever executed.

### Fix

Remove the three dead definitions. Runtime behavior is unchanged — the live
`_set_combobox_value`, `_load_analysis_settings_slot`, and `_set_widget_value`
(still called at its correct 3-arg call site) all remain. `ruff check` passes.


HarrisonKramer:
and one more. please rebase and check test failures. Let me know if you need any help.

Thanks again!

PR #800, opened 2026-09-10
Validate OSLO sequential import and loss-aware export

BrunoChevalier:
OSLO imports could silently discard optical commands, ignore lens units, misalign asphere powers and turn every PY solve into final-image autofocus. Add source-located diagnostics and opt-in strict import, map validated sequential prescriptions onto existing Optiland capabilities, and reject unrepresentable exports before opening their destination.

This is the **first scoped replacement for [draft PR #786](https://github.com/optiland/optiland/pull/786)**, following the [maintainer discussion](https://github.com/optiland/optiland/pull/786#issuecomment-5607238556). The draft retains the original discussion and broader implementation history. This PR contains one feature commit and is limited to file I/O, its tests, documentation and the demo audit script.

Included:

- Quote-aware parsing, command/source diagnostics, units, indexed spectra and explicit field tables.
- Independently selected declarative configurations, with isolated prescription state and explicit rejection of unverified pickup/configuration precedence.
- Supported conic/aspheric/toric geometry, signed coordinates and local mirror bends using existing native geometry.
- Static pickups, targeted paraxial and edge-contact solves, nominal image focus and post-solve validation.
- Export validation and preservation of supported spectral, field and geometric setup.
- A [support guide](https://github.com/BrunoChevalier/optiland/blob/feature/oslo-io/docs/oslo_import.rst) and reproducible audit of the separately obtained official Lens Demos archive.

Subsequent PRs will add sampled materials, physical/drawing apertures, polarization, explicit glass catalogs and GUI import. Each native feature will arrive **together with the OSLO import/export mappings that depend on it**. This first part requires none of those additions. The existing public reader's d/F/C Abbe approximation remains explicitly warned and is rejected in strict mode; exact sampled dispersion belongs to the material PR. Unsupported clipping behavior is diagnosed until the aperture PR can represent it faithfully.

Validation:

- The scoped file-I/O run passed 746 tests on NumPy/Torch; a subsequent five-case boundary run also passed after adding one regression. Tests include independent geometry, ray-launch, solve and configuration-isolation assertions.
- Local patch coverage is **100% (2526/2526 changed executable lines)** against upstream `7df37590`, using the current Codecov exclusions. No coverage configuration or threshold changes are included.
- Ruff lint/format checks and documentation syntax validation pass. A full Sphinx build was not run locally because Sphinx is unavailable.
- Review regressions cover mixed-case pickup conflicts, converter reuse after success/failure, and strict conversion of permissively parsed models.

[...]

PR #573, opened 2026-04-30
feat(materials): Add Buchdahl polynomial model material

Xiong-Ziyi:
## Summary
- Adds `ModelMaterial` using a 6th-order Buchdahl polynomial model.
- Supports visible and IR routing through regression matrices.
- Loads regression matrices from `optiland.database`.
- Adds docs notebooks for the model material workflow.

## Validation
- Ran Ruff on `optiland/materials/model.py`.

## Tests
- Not Run

PR #436, opened 2026-01-08
feat: Implement GRIN (Gradient-Index) material and propagation

jiecaix:
## Summary
This PR implements Gradient-Index (GRIN) material support in Optiland, enabling accurate modeling of optical components with spatially varying refractive indices. Open #337

### 1. GRIN Material Implementation
- New `GradientMaterial` class in optiland/materials/gradient_material.py
- Supports arbitrary refractive index profiles n(x, y, z)
- Built-in profiles: axial, radial, and gradient distributions
- Extensible API for custom gradient functions

### 2. GRIN Ray Propagation
- Enhanced GRIN propagation algorithm in optiland/propagation/grin.py
- Numerical ray tracing through gradient-index media using 4th-order Runge-Kutta method
- Bidirectional propagation support (forward and backward)
- Proper handling of reflections at GRIN material boundaries
- Automatic ray tracing step adaptation based on gradient strength

### 3. Comprehensive Testing
- 269 lines of new tests in tests/propagation/test_grin.py
- Tests for axial, radial, and spherical gradient profiles
- Validation of bidirectional propagation with reflections
- Serialization test coverage

### 4. Examples and Documentation
- Detailed implementation guide: GRIN_IMPLEMENTATION_SUMMARY.md
- Interactive examples:
  - examples/simple_grin_test.py - Basic GRIN lens behavior
  - examples/grin_reflection_detailed.py - Reflection effects in GRIN media

## Changes
- **8 files changed**, 1,322 insertions(+), 25 deletions(-)
- Three commits implementing core functionality, testing, and reflection support

## Testing
All existing tests pass, plus new comprehensive GRIN material tests covering:
- Axial gradient propagation
- Radial gradient focusing effects
- Spherical gradient profiles
- Bidirectional ray tracing with reflections
- Edge cases and boundary conditions

This implementation enables Optiland to model advanced optical components like GRIN lenses, gradient-index fibers, and other waveguide devices critical in modern optical systems.

PR #842, opened 2026-09-13
feat(materials): preserve owned spectral data and OSLO index samples

### LINKED ISSUE #841: Add self-contained spectral materials and shared material evaluation contracts
OSLO prescriptions can contain refractive-index samples at several wavelengths without naming a catalog glass. The current reader reduces this data to an Abbe approximation, or rejects it in strict mode. That changes the imported dispersion and cannot preserve arbitrary measured samples. Users with their own sampled or analytic optical data also need a material that can be saved and restored without creating a YAML file or registering a catalog entry.

This calls for one data-owning material and shared numerical helpers, rather than separate “embedded” and “tabulated” material implementations. Embedding describes how data is persisted; tabulation describes one way to evaluate it. Those are independent concerns.

## Proposed architecture

| Component | Responsibility | Principal callers |
| --- | --- | --- |
| `BaseMaterial` | Common `n(wavelength, **kwargs)` / `k(...)` interface, subclass dispatch, optional state-aware caching, and central propagation restoration. | Tracing, interactions, plotting, native persistence, custom material families. |
| `DataMaterial` | Own a copied, immutable optical definition: sampled or analytic refractive index and optional independently sampled extinction. Native JSON embeds the complete definition and copied descriptive metadata. | User-created data, OSLO sampled-index import, explicit material overrides, native reload. |
| `MaterialFile` | Read existing refractiveindex.info YAML data, retain its live runtime state and public filename-based persistence, and delegate evaluation. | Existing YAML users and the catalog material wrapper. |
| `Material` | Catalog resolution and catalog identity on top of `MaterialFile`. | Named-glass lookup, material selection and readers with catalog references. |
| `IdealMaterial` and Abbe families | Own their existing constant or fitted live parameters. Derive fitted coefficients from current parameters. | Tracing, model-glass optimization and existing readers. |
| Shared spectral / dispersion helpers | Implement sample validation, interpolation with an explicit bounds policy, the nine supported refractiveindex.info equations, and Buchdahl coordinate/polynomial algebra. | Data-backed materials, existing file materials and fitted models. |
| Private definition codec | Validate and encode owned optical records without file access, caches or backend-specific storage. | `DataMaterial` construction and native persistence. |
| File adapters | Own parsing, physical restrictions and the subset that each target format can preserve. | OSLO, CODE V and Zemax import/export. |

```mermaid
flowchart TD
    Base[BaseMaterial runtime interface] --> Data[DataMaterial: owned optical data]
    Base --> File[MaterialFile: YAML adapter]
    File --> Catalog[Material: catalog identity]
    Base --> Models[Ideal and fitted materials]
    Data -. evaluation .-> Spectral[Shared spectral and formula helpers]
    File -. evaluation .-> Spectral
    Models -. Buchdahl evaluation .-> Buchdahl[Shared coordinate and polynomial kernels]
    Data -. persistence .-> Codec[Immutable definition codec]
    OSLO[OSLO sampled-index reader] -. constructs .-> Data
    Views[Plots and information views] -. identity and range queries .-> Base
```

Solid arrows show specialization; dotted arrows show use. This is not a requirement for every future material to use the owned-data codec.

## Optical and persistence contracts

- Wavelengths are in micrometers; `k` is the dimensionless extinction coefficient, not internal transmission. Keep refractive-index and extinction sample grids independent.
- Preserve the actual supplied sample pairs. Sort them together; require finite data, distinct positive wavelengths and nonnegative extinction. Owned tables reject extrapolation. Existing file materials retain their supported endpoint-clamping behavior, including one-sample tables.
- Preserve analytic equations as equations; do not replace them with sampled or fitted approximations. Optional declared wavelength limits are checked for owned analytic data.
- Scalar queries return a one-element result; array queries retain shape and wavelength gradients. Keep numerical evaluation backend-agnostic and Torch optional.
- Imported data is immutable and replaced as a complete definition. Live optimization parameters belong to parameterized material models, not that immutable definition. Share evaluation math without sharing unrelated regression fits or changing reference-line conventions.
- Native persistence must restore the optical definition, descriptive metadata and registered propagation model. A display name must never trigger catalog lookup. Preserve released public file/catalog formats; no compatibility aliases are needed for earlier unmerged prototypes.
- Consumers use `display_name` and `spectral_range('n'/'k')` rather than duplicating type-specific naming and range inference. Unknown bounds require an explicit plotting interval. Explicit custom display formatters remain supported.
- OSLO should read distinct direct-index samples as `DataMaterial` and write supported positive index tables without extinction. CODE V and Zemax should reject owned definitions they cannot preserve before replacing the destination. All three writers should reject unsupported custom propagation before air, catalog, fictitious-glass or mirror shortcuts discard it.

## Relationship to existing work

The open issues were reviewed before filing this request. #572 / PR #573 covers a sixth-order Buchdahl glass model with visible and IR routing; it does not cover owned sampled data, shared file-material evaluation or the persistence/display architecture above. Its model should remain a live parameterized material and can reuse the same Buchdahl algebra as the existing three-term Abbe models. Regression assets, model fitting and the proposed `IndexVariable` changes remain in that work.

#337 / PR #436 covers GRIN. A future spatial material can be a sibling of `DataMaterial`, optionally composing a homogeneous background. It needs position-aware callers and a consistent field/propagator contract; the present work must not imply that wavelength-only tracing or file writers support GRIN. No spatial fields or integrator settings need to be added to the owned spectral-data schema in advance.

#223 covers an interactive glass map, #458 / PR #687 fabrication drawings, and #733 wavelength-dependent phase maps. They are consumers or separate optical capabilities. They can use the common identity/property interface without requiring another data material or copying dispersion equations. PR #705's scalar-wave propagation takes a homogeneous numerical index and does not require a new material hierarchy.

## Delivery scope and acceptance

Implement the shared material foundation and the OSLO sampled-data mapping together. Physical aperture support, polarization changes, OSLO catalog/thermal/transmission decoding and GUI import workflows remain separate feature deliveries; their format support should ship with the native capability it requires.

- [ ] Add self-contained sampled/analytic `DataMaterial` with independent extinction data and native round-trip tests.
- [ ] Reuse spectral, formula and Buchdahl calculations across the applicable existing material families.
- [ ] Preserve existing file/catalog persistence, live model parameters and optional-backend behavior.
- [ ] Preserve OSLO index samples and reject unsupported lossy exports before touching an existing destination.
- [ ] Use the common material identity/range contract in plotting and information displays, including new custom families.
- [ ] Document each material's responsibility, state ownership, callers and future extension obligations.
- [ ] Validate NumPy, Torch, repeated gradients, precision/broadcasting boundaries, native persistence and meaningful line/branch/patch coverage with unchanged CI thresholds and exclusions.

BrunoChevalier:
OSLO direct-index prescriptions currently lose their sampled dispersion through an Abbe approximation, or fail strict import. This change preserves the supplied wavelength/index pairs in a self-contained `DataMaterial`, with native JSON containing its optical definition and descriptive metadata. It also supports analytic refractiveindex.info coefficients and independent extinction samples without requiring a temporary file or catalog registration.

The material architecture keeps owned data separate from live fitted models and file/catalog adapters. Shared helpers implement spectral validation/interpolation, the nine refractiveindex.info formulas and Buchdahl coordinate/polynomial evaluation. Existing file materials retain their public persistence; existing Abbe fits retain their constants and reference conventions. Plotting and information displays use common identity/range contracts, including unregistered material families.

## Explicit table bounds

The bounds policy and older-file fixes are separate commits from the architecture change:

```python
DataMaterial.from_samples(wavelengths, indices, bounds="clamp")
MaterialFile("measurements.yml", bounds="raise")
Material("N-BK7", catalog="schott", bounds="raise")
```

`DataMaterial` continues to default to `"raise"`; `MaterialFile` and `Material` continue to default to `"clamp"`. Each tabulated n/k property uses its own sample interval. Clamping explicitly assumes the nearest endpoint value beyond supplied data and does not expand the reported measured support. The interpolation's wavelength derivative is zero outside its interval, while gradients to live file-table endpoint samples are preserved. Analytic formula limits and thermal corrections retain their existing separate contracts.

The policy is read-only after construction, included in cache state, and serialized in native JSON. Older `DataMaterial` records that omit `bounds` restore with `"raise"`, preserving their original strict interpolation instead of failing with `KeyError`. Older file/catalog records retain clamping. Explicit policies are preserved, invalid explicit values still fail validation, and new saves record the chosen policy. Loading an older design does not rewrite the source file. This also preserves saved choices if a later breaking release changes the file-material default.

OSLO export preserves the supported positive sampled-index subset with the default strict policy; it rejects explicit DataMaterial clamping because LEN cannot encode that choice. CODE V and Zemax reject owned definitions they cannot preserve, and all three writers reject custom propagation before air/catalog/model/mirror shortcuts can discard it. Rejections happen before replacing the destination. Catalog-name exports continue to use the destination program's conventions. The developer guide documents each family's responsibilities, state owner, callers, bounds policy and future extension obligations.

Related integration work: PR #573 can reuse the Buchdahl kernels while retaining its own six-term fitting/optimization work; PR #436 can introduce a spatial sibling/composed background with its own tracing contract. This PR follows merged #800. Catalog/thermal/transmission support, physical apertures, polarization and GUI import remain separate deliveries.

## Validation

Current revision: `1df4b0d260232e75c8f8f3f08602b9b90c9aaea9` (September 21).

- Fresh affected-scope run with the unchanged CI line-coverage configuration: **1,917 passed, 31 skipped**. The cumulative PR patch is **429/429 executable lines covered** against its merge base `8bd86cf4c318`.
- Separate affected-scope branch run: **1,917 passed, 31 skipped**, with **146/146 changed branch paths covered**. No changed line or branch gaps remain in the NumPy/Torch measurement.
- Torch genuinely absent: **1,223 passed, 53 skipped**, with no deselections. The two cumulative-patch lines requiring Torch are covered by the NumPy/Torch run. The corrected DataMaterial loader is fully covered in both environments.
- Thirteen new regression cases cover fixed historical records, strict n/k intervals, analytic validity, invalid explicit policies, and full native-design import/save/reload on NumPy and Torch. Existing tests retain explicit `raise`/`clamp` and old file/catalog defaults.
- `data.py`: **73/73 statements and 14/14 branch outcomes**. The one changed keyword expression maps to its enclosing return statement, which is covered; it introduces no new Python branch.
- Ruff lint/format, configured mypy (**532 source files**), and a focused materials guide/API Sphinx build with warnings treated as errors pass.

These are fresh targeted Windows CPU/Python 3.14.6 measurements using the retained frozen-lock environments with Numba JIT enabled. The scope includes material families, native serialization, file I/O/export, propagation, visualization consumers and implicit differentiation. The separate branch/Torch-free measurements are not combined with ordinary CI line coverage. CPU execution skips CUDA cases; GitHub runs the Linux/Python-version matrix. No coverage threshold, exclusion, CI configuration or warning suppression was changed.

Historical full-suite evidence from the preceding revision `028af0f9`: **11,373 passed, 198 skipped, 2 xfailed**, with 92.93% eligible tracked line coverage. That complete-suite result was recorded on September 16 and was not rerun or relabeled as a full-suite result for this focused deserialization correction.

Closes #841.


BrunoChevalier:
@HarrisonKramer sorry for all the pull requests these last days 😄 
I'm about halfway through them.

This one might need some discussion and architectural feedback as it touches on the materials.
I tried to line up as much as possible, taking any open PRs related to materials into account as well.

If anything needs rework, let me know.

HarrisonKramer:
Hi @BrunoChevalier,

I went through the PR and ran the test suite locally. Everything looks very solid.

A few specific things that look good:
* Using frozen dataclasses for `MaterialDefinition` keeps the immutable definitions well separated from mutable state and caching.
* Centralizing the 9 refractiveindex.info formulas in `optiland.materials.dispersion` between `MaterialFile` and `DataMaterial` cleans up a lot of duplication.
* Breaking out `buchdahl_coordinate` and `evaluate_buchdahl` sets up PR #573 nicely.
* The autograd handling (bypassing caches when tensors have gradients) works as expected for PyTorch optimization.
* Failing fast in the CODE V and Zemax writers instead of guessing an Abbe fit is definitely the right call.

One quick question before merging:
* **Wavelength out-of-bounds policy:** In `DataMaterial.from_samples`, out-of-bounds queries clamp to boundary values (matching `MaterialFile`). Is clamping preferred here over raising an error or warning when tracing outside the sampled window?

Looks good so far. Thanks for opening! Once we sort it all out, should be good to merge.

BrunoChevalier:
@HarrisonKramer 

Thanks for checking this! The intended policy for `DataMaterial` is to raise outside the sampled window, and that is already implemented and tested. It uses the shared interpolation helper’s default `bounds="raise"`.
`MaterialFile` explicitly selects `"clamp"` to preserve its existing public behavior.
The reasoning is that holding the endpoint value would introduce an unsupported assumption about dispersion or absorption beyond the supplied data. For newly supplied samples, I prefer an explicit error over silently continuing with that assumption. We share the interpolation implementation while keeping this deliberate distinction between the new material’s contract and the existing file adapter.

I have pushed an update to allow an opt-in to use the clamping behavior as well for `DataMaterial` as I understand some people would want that behavior. They should of course first understand the implications before doing this.

Issue #746, opened 2026-08-19
5 `test_ray_aiming.py tests fail on NixOS


doronbehar:
### Checklist

- [x] I have searched the [existing issues](https://github.com/HarrisonKramer/optiland/issues) and [discussions](https://github.com/HarrisonKramer/optiland/discussions) for a similar question or problem.
- [x] I have read the [documentation](https://optiland.readthedocs.io/en/latest/) and tried to find an answer there.
- [x] I am using the **latest version** of Optiland (if not, please update and retry).
- [x] I have tried to reproduce or debug the issue myself before opening this.
- [ ] I have included **all necessary context**, such as version info, error messages, or minimal reproducible examples.

### Bug Report

**Describe the bug**

These 5 tests all fail the same way:

```
=================================== FAILURES ===================================
______ test_epd_invariant_under_translation[backend=numpy-25.0-issue613] _______
[gw3] linux -- Python 3.14.7 /nix/store/3n4qphl9s728sz8frmpqqrv9b1m87g68-python3-3.14.7/bin/python3.14

set_test_backend = None
build = <function _issue_613_reproducer at 0x7fff6ecb2610>, dz = 25.0

    @pytest.mark.parametrize(
        "build",
        [CookeTriplet, ReverseTelephoto, SingletStopSurf2, _issue_613_reproducer],
        ids=["CookeTriplet", "ReverseTelephoto", "SingletStopSurf2", "issue613"],
    )
    @pytest.mark.parametrize("dz", [25.0, -15.0])
    def test_epd_invariant_under_translation(set_test_backend, build, dz):
        """``EPD()`` is a length, unchanged by rigid translation."""
>       epd_ref = float(build().paraxial.EPD())
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       TypeError: only 0-dimensional arrays can be converted to Python scalars

tests/test_ray_aiming.py:335: TypeError
______ test_epd_invariant_under_translation[backend=numpy--15.0-issue613] ______
[gw3] linux -- Python 3.14.7 /nix/store/3n4qphl9s728sz8frmpqqrv9b1m87g68-python3-3.14.7/bin/python3.14

set_test_backend = None
build = <function _issue_613_reproducer at 0x7fff6ecb2610>, dz = -15.0

    @pytest.mark.parametrize(
        "build",
        [CookeTriplet, ReverseTelephoto, SingletStopSurf2, _issue_613_reproducer],
[...]

PR #850, opened 2026-09-14
Add Friday dependency compatibility matrix

### LINKED ISSUE #849: Expand dependency-version coverage with a weekly compatibility matrix
The current CI varies Python 3.11–3.14 but installs the same dependency versions
from `uv.lock` in every test job. This gives useful Python-version coverage, but
leaves other supported combinations of NumPy, Numba/llvmlite, SciPy and optional
PyTorch largely unchecked.

Issue #746 is a concrete example: the locked NumPy 2.3.5 permits an array-to-scalar
conversion with a deprecation warning, while NumPy 2.4+ raises an error. A green
locked CI run therefore does not establish compatibility with the broader
dependency versions users install. The immediate scalar-test fix is handled
separately in #848.

### Proposed approach

Add a separate **weekly compatibility run**, on Fridays so contributors have
the weekend to investigate failures. Run the existing full suite against a
small, explicit matrix of supported Python and scientific-library combinations,
including CPU PyTorch and an environment where PyTorch is genuinely absent.

The weekly run will detect failures in the selected combinations without making
every ordinary CI run pay for that expanded matrix. Keep normal push/PR CI on
its locked environments to maintain reasonable runtime and quick review feedback.
Changes to the matrix or its tooling should receive inexpensive validation during
normal PR review, with manual compatibility runs available for qualification.

### Initial scope

- Eight Ubuntu CPU combinations: a locked control, reviewed current stacks for
  Python 3.11–3.14, a NumPy 2.4 boundary, a preceding Torch version, and Torch absence.
- One editable configuration file for adding/removing combinations or updating
  versions without changing workflow logic.
- Review numerical-library pins monthly and after relevant releases; a pinned
  matrix only detects regressions in versions it actually tests.
- Run the full suite, preserve failing results, and retain exact dependency
  inventories, resolution data, test logs and reports for reproduction.
- Limit concurrency and per-job runtime, measure the initial jobs, and expand
  coverage deliberately rather than creating an unbounded Cartesian product.

### Acceptance

- [ ] The weekly workflow and manual row selection are implemented and documented.
- [ ] All intended combinations are qualified on GitHub Ubuntu, with recorded
      versions, results and runtime; discovered application failures are tracked.
- [ ] Configuration updates are straightforward and validated.
- [ ] Failures and missing diagnostics remain visible; there are no blanket
      failure allowances or skip filters.
- [ ] Ordinary PR CI retains a reasonable runtime without running the eight
      additional full suites on every change.

This tracks **dependency-version test coverage**, rather than a code-line
coverage percentage. Weekly sampling complements normal CI; it does not promise
exhaustive compatibility or immediate detection of every new dependency release.

Related: #746, #848.

BrunoChevalier:
Closes #849. Related: #746 and the merged fix in #848.

Issue #746 exposed a gap in the current CI: it varies Python 3.11–3.14 while using the same locked scientific-library versions. A scalar conversion accepted with a warning by NumPy 2.3.5 fails on newer NumPy. More supported dependency combinations need regular testing before users encounter those incompatibilities.

This adds a separate workflow on **Fridays at 03:23 UTC**, so contributors have the weekend to investigate failures. Manual runs accept all rows or selected comma-separated row IDs. Ordinary application PRs keep their existing CI workload; changes to the compatibility configuration/helper receive a lightweight validation job.

### Why these eight combinations

- A control follows the committed lock on the oldest supported Python.
- Four reviewed current stacks span Python 3.11–3.14 with compatible NumPy, Numba, llvmlite and SciPy versions.
- A NumPy 2.4 boundary row retains coverage of the scalar-conversion change.
- Previous-Torch and genuinely Torch-absent rows exercise backend-version and optional-dependency behavior.

These are selected representative combinations, rather than every possible cross-product. They do not claim to reproduce the exact original issue environment. Every row runs the full existing test suite, including GUI tests under Qt offscreen and Xvfb. CPU Torch wheels reduce downloads; three jobs run concurrently with a 75-minute limit each.

### Keeping the combinations maintainable

`.github/compatibility-matrix.toml` is the only maintained version/combination list. Editing pins, constraining another package, or adding/removing a row does not require changing workflow logic. The stdlib helper validates the manifest, resolves dependencies through uv, verifies installed versions/backends/JIT, and reports failed stages or missing artifacts. Primary pins are reviewed monthly and after relevant releases; unspecified dependencies resolve to compatible stable versions.

Each run retains resolved requirements with hashes, environment inventories, logs, JUnit and coverage reports, source SHA and runner information. Weekly coverage stays in its own artifacts and is not mixed into ordinary PR Codecov uploads. The workflow uses Node 24 action releases.

### Optional-Torch test handling

The first full fork run passed seven combinations but exposed 113 test failures/setup errors without Torch. Shared tests now parameterize installed backends, while actual Torch-only tests skip before importing or selecting Torch. NumPy-side rejection tests remain active. Serialization tests use real NumPy values and real Torch tensors instead of an invalid mock fallback, and diagnostic spelling checks use the always-available NumPy backend.

No application code, dependency metadata/lock, numerical assertions, global skip filters or warning filters are changed. The CI helper is outside the runtime package, so this adds no application execution overhead.

### Validation

- Local affected-module tests on Windows: **856 passed with Torch**, **450 passed without Torch**, plus 28 subtests in each environment.
- Helper validation: **15 tests**; CI-mode helper line/patch coverage **263/263 (100%)**. A real entrypoint test covers successful and failing exit statuses, including the line previously reported missing by Codecov.
- Separate supplementary helper branch coverage: **130/130 (100%)**. Ordinary coverage configuration and exclusions remain unchanged.
- Ruff's CI commands, Actionlint, real version/backend verification and the earlier incompatible-pin failure checks pass.
- [Full Ubuntu qualification](https://github.com/BrunoChevalier/optiland/actions/runs/34883800390): **all eight combinations and the final summary passed**. Each of the seven Torch-present rows reported **10,700 passed, 195 skipped and 2 xfailed**; the Torch-absent row reported **6,118 passed, 424 skipped and 1 xfailed**. All **165 issue #746 regression cases** passed across the matrix.
- All eight saved environment inventories match the intended source SHA and manifest hash. Every row reports helper line coverage **263/263 (100%)**. The workflow took **74 minutes 25 seconds**, using approximately **3.31 aggregate matrix runner-hours** with three concurrent jobs. No Node 20 deprecation annotations remain.

The qualified SHA is `eaa73361193d293879736584e76c5c316966d626`: it has the same tree as implementation `2ecac904`, plus a temporary push trigger for the fork qualification branch. That trigger is not included here. A subsequent one-line workflow correction at `758bcd67` uses the unique matrix row ID instead of the Python version for the uv cache suffix, addressing the run's two harmless cache-reservation warnings. Actionlint and the eight unique cache namespaces were checked locally; the completed run predates that cache-only edit and does not validate its remote save behavior. The cache-only edit changed no Python code, tests or dependency selections.

The scalar-conversion fix remains in the separate, now-merged #848. Harrison's Vercel archive change in #851 belongs to documentation deployment; this workflow only stores GitHub test artifacts and needs no Vercel deployment.

### Upstream merge validation (2026-09-15)

Merged upstream `17610805` into this branch at `ad77b443`, resolving the overlaps with #854 while retaining its regression tests and this PR's optional-Torch handling. The new Adam/SGD total-track tests explicitly request the Torch fixture. Application code and #851's Vercel changes match upstream unchanged.

On the merged revision, all 17 affected/local regression modules pass with **975 tests and 28 subtests with Torch** and **504 tests and 28 subtests without Torch**. CI-mode helper line/patch coverage remains **263/263 (100%)** in both environments. Ruff and Actionlint pass. The eight-row Ubuntu result above belongs to the earlier qualification SHA; ordinary GitHub PR checks will validate this new merge revision separately.


HarrisonKramer:
Thanks for putting this together, @BrunoChevalier, and for rebasing after #848.

The setup is clean and well thought out -- especially isolating the heavy matrix runs to the Friday schedule and cleaning up the test suite to handle environments without Torch. All checks look solid.

Merging now.

Kramer

